PR #27, opened 2023-05-16
Add test with successful call to definitional engine.

andrewmwells-amazon:
Add test with successful call to definitional engine.


john-h-kastner-aws (approved):
Looks useful

khieta (approved):

PR #30, opened 2023-05-18
cedar-spec changes for arithmetic overflow validation

mattmccutchen-amazon:
*Issue #, if available:* cedar-policy/cedar#63

*Description of changes:*
Changes to cedar-spec that are part of our plan to build confidence in the correctness of the arithmetic overflow validation feature (see cedar-policy/cedar#56). (Draft; more details will be provided here when the work is further along.)

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #38, opened 2023-05-31
Update DRT build instructions

khieta:
*Issue #, if available:*

---------

*Description of changes:*

* Added `cedar` as a submodule instead of requiring users to create their own local clone.
* Minor updates to the build instructions.
* Marked as "draft" until we decide which version of `cedar-*` we want to pin.

**Note:** We can use the `cedar-*` packages in crates.io instead of the submodule, if desired. But we still need to include `cedar` as a submodule because the integration tests in `cedar-spec/cedar-drt` depend on the source in `cedar/cedar-integration-tests`.

---------

### DIFF
--- a/cedar-drt/Cargo.toml
+++ b/cedar-drt/Cargo.toml
@@ -10,10 +10,10 @@ highway = "0.8.1"
 env_logger = "0.10"
 log = "0.4"
 libfuzzer-sys = "0.4"
-cedar-policy = { path = "../cedar/cedar-policy", version = "2.0.0", features = ["integration_testing"] }
-cedar-policy-core = { path = "../cedar/cedar-policy-core", version = "2.0.0" }
-cedar-policy-validator = { path = "../cedar/cedar-policy-validator", version = "2.0.0" }
-cedar-policy-formatter = { path = "../cedar/cedar-policy-formatter", version = "2.0.0" }
+cedar-policy = { path = "../cedar/cedar-policy", version = "2.*", features = ["integration_testing"] }
+cedar-policy-core = { path = "../cedar/cedar-policy-core", version = "2.*" }
+cedar-policy-validator = { path = "../cedar/cedar-policy-validator", version = "2.*" }
+cedar-policy-formatter = { path = "../cedar/cedar-policy-formatter", version = "2.*" }
 jni = { version = "0.19.0", features = ["invocation"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ cedar-dafny/.dotnet
 cedar-dafny/.local
 cedar-dafny/.nuget
 cedar-dafny/build
+cedar-dafny/TestResults
 
 # cedar-dafny-java-wrapper build artifacts
 cedar-dafny-java-wrapper/build
@@ -29,8 +30,5 @@ cedar-drt/fuzz/target
 cedar-drt/fuzz/Cargo.lock
 cedar-drt/fuzz/corpus
 cedar-drt/fuzz/artifacts
-
-# cedar if checkout out in root dir
-cedar/
 
 .vscode
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "cedar"]
+	path = cedar
+	url = https://github.com/cedar-policy/cedar.git
--- a/README.md
+++ b/README.md
@@ -7,19 +7,19 @@ This repository contains the Dafny formalization of Cedar and infrastructure for
 * `cedar-spec` contains the Dafny formalization of, and proofs about, Cedar.
 * `cedar-dafny-java-wrapper` contains the Java interface for DRT.
 * `cedar-drt` contains code for input generation, fuzzing, property-based testing, and differential testing of Cedar.
+* `cedar` is a git submodule, pinned to the `main` branch of [cedar](https://github.com/cedar-policy/cedar).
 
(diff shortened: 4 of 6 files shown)


cedar-drt/Cargo.toml
```diff
@@ -10,10 +10,10 @@ highway = "0.8.1"
...
 log = "0.4"
 libfuzzer-sys = "0.4"
-cedar-policy = { path = "../cedar/cedar-policy", version = "2.0.0", features = ["integration_testing"] }
-cedar-policy-core = { path = "../cedar/cedar-policy-core", version = "2.0.0" }
-cedar-policy-validator = { path = "../cedar/cedar-policy-validator", version = "2.0.0" }
-cedar-policy-formatter = { path = "../cedar/cedar-policy-formatter", version = "2.0.0" }
+cedar-policy = { path = "../cedar/cedar-policy", version = "2.*", features = ["integration_testing"] }
+cedar-policy-core = { path = "../cedar/cedar-policy-core", version = "2.*" }
+cedar-policy-validator = { path = "../cedar/cedar-policy-validator", version = "2.*" }
+cedar-policy-formatter = { path = "../cedar/cedar-policy-formatter", version = "2.*" }
```
khieta, line 16: Which version do we want to use here? I opted for `2.*` because it won't require us to update this repo with new minor versions.

andrewmwells-amazon: Seems reasonable to me.

andrewmwells-amazon (approved):
LGTM. It might be a good idea to also update ci.yml to match these build instructions.

andrewmwells-amazon (approved):
LGTM. It might be a good idea to also update ci.yml to match these build instructions.

khieta:
Updated `ci.yml` to use the `cedar` submodule rather than a separate clone.

I did not update CI to use the `./build.sh` script because the CI uses slightly different settings (e.g., warning flags). I also thought it might be useful to have separate steps in the CI (rather than a single "build" step) to diagnose failures. Open to second opinions.

aaronjeline (approved):

andrewmwells-amazon (approved):

cedar-drt/set_env_vars.sh
```diff
@@ -1,6 +1,22 @@
...
+# Set JAVA_HOME
 if [ -z "${JAVA_HOME-}" ]; then
     # Idea from https://www.baeldung.com/find-java-home.
     export JAVA_HOME="$(java -XshowSettings:properties -version 2>&1 | sed -ne 's,^ *java\.home = ,,p')"
 fi
-export LD_LIBRARY_PATH=${LD_LIBRARY_PATH+$LD_LIBRARY_PATH:}$JAVA_HOME/lib/amd64/server
-export CLASSPATH="$(< ../cedar-dafny-java-wrapper/build/runtimeClasspath.txt):$(pwd)/../cedar-dafny-java-wrapper/build/libs/cedar-dafny-java-wrapper.jar"
+
+# Set LD_LIBRARY_PATH
+libjvm_dirs=($(find "$JAVA_HOME" -name 'libjvm.*'))
```
mattmccutchen-amazon, line 10: This is not working for me when I follow the instructions to run DRT.  I get the usual error: `fuzz/target/x86_64-unknown-linux-gnu/release/rbac: error while loading shared libraries: libjvm.so: cannot open shared object file: No such file or directory`.  It looks like the problem is that compared to my version [here](https://github.com/cedar-policy/cedar-spec/pull/30/files/2187bd634789001de55cbfc4b62299f57f70d9db#diff-fe65b77103b1950854ab4df96bca59a7c3431a4106c43366f2e526d64c461606R12), you removed the `-printf '%h\n'`, so the command prints the path to `libjvm.so` instead of the parent directory, and `LD_LIBRARY_PATH` gets set (on my system) to `/usr/lib/jvm/java-11-amazon-corretto.x86_64/lib/server/libjvm.so`, which doesn't work because `LD_LIBRARY_PATH` is supposed to contain directories.

What was your motivation in making the change?  To make the script work on macOS?  If we want to make DRT actually work on macOS, I suspect we have a lot more work to do than this.  For one thing, I think we would need to set `DYLD_LIBRARY_PATH` instead of `LD_LIBRARY_PATH`.

Unless I'm missing something, I'd like to fix this code to output the parent directory again so it works on my system.  If you want, I can do that in a way that still runs successfully on macOS, for what that's worth.  If it were my decision, I'd be more inclined to remove this illusion of generality and revert the whole thing to my version.

(Was there discussion elsewhere that I missed?)

Aside: It looks like the reason the CI didn't catch the problem before the PR was merged is that `ci.yml` duplicates parts of the logic from `set_env_vars.sh` rather than calling it.  I don't know if that's worth changing.

khieta: *sigh* Sorry, I should have tested my solution better. I'll file an issue shortly.

Yes, removing the `-printf` was to make the script work on macOS. I (incorrectly) assumed this wouldn't break anything else. The current script works in my local development environment (macOS).

khieta: Also: Yes, `ci.yml` should call the `./set_env_vars.sh` and `./build.sh` scripts directly to avoid errors like this in the future. I (again incorrectly) assumed that my changes wouldn't break anything.

khieta: Issue here: https://github.com/cedar-policy/cedar-spec/issues/41 @mattmccutchen-amazon can I assign this to you?

mattmccutchen-amazon: I assigned #41 to myself.

Issue #35, opened 2023-05-26
Reduce proof resource usage


khieta:
### Category

Other

### Describe the feature you'd like to request

Several of our proofs (in particular in `cedar-dafny/validation/thm/soundness.dfy` and `cedar-dafny/validation/thm/model.dfy`) are computationally expensive, which makes them likely to timeout with future versions of Dafny and/or Z3. We can avoid this by refactoring the proofs to reduce their required "resource units" (as determined by Z3).

*Note:* you can compute the resource units required for each proof by setting the `GEN_STATS` flag to 1 before compiling the cedar-dafny proofs. The results will be recorded in a `TestResults` directory.

### Describe the solution you'd like

N/A

### Describe alternatives you've considered

N/A

### Additional context

Useful reference for improving Dafny proofs: https://dafny.org/dafny/VerificationOptimization/VerificationOptimization

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

shaobo-he-aws:
I have a general question about this issue. Suppose we can break a complicated lemma costs 40M RUs into 4 lemmas, each of which costs 10M RUs. Does it count as an improvement?

khieta:
I believe so, since by default Dafny reports time outs per definition. So while the one 40M RU lemma may time out in the VSCode environment, the four 10M RU lemmas might verify.

khieta:
We've made good progress in this direction (see #40, #56, #57, #63) and added CI to help check for proofs with high resource usage (#61). Marking this issue as closed, but we may open more specific issues in the future.

PR #40, opened 2023-06-02
Some optimizations to the validator proofs.

mattmccutchen-amazon:
*Issue #, if available:* None, but related to #30

*Description of changes:* Some optimizations to the validator proofs. In combination with the overflow validation branch, these were needed to keep the proofs under the limit both at the command line and in VS Code on my system. But they are valid optimizations in their own right.


khieta (approved):

shaobo-he-aws (approved):

khieta:
Note: partially addresses #35

PR #50, opened 2023-06-22
Enable generating unknowns for PE fuzzing

aaronjeline:
I think we want this setting on.


cdisselkoen:
We're enabling it for `pp`?  do unknowns have a syntax representation that we expect to roundtrip correctly (prettyprint --> parse)?

PR #56, opened 2023-06-29
Optimize proof RUs

john-h-kastner-aws:
Reduces proofs over 10,000,000 RU limit from 12 to 6.

The first commit tweaks some proofs in `model.dfy`. The second commit hides some common preconditions in `model.dfy` behind `opaque`. 

Before:
```
validation.thm.soundness.SemanticSoundnessProof.SoundIn (correctness)(Passed) - Duration = 00:00:19.0112924, Resource Count = 23905971
validation.thm.soundness.SemanticSoundnessProof.SoundGetAttr (correctness)(Passed) - Duration = 00:00:20.4276508, Resource Count = 21252770
validation.thm.soundness.SemanticSoundnessProof.SoundCall (correctness)(Passed) - Duration = 00:00:16.1544087, Resource Count = 20447879
validation.thm.model.CallSafe (correctness)(Passed) - Duration = 00:00:16.4184827, Resource Count = 19604861
validation.thm.model.InSetFalseTypes (correctness)(Passed) - Duration = 00:00:11.1384372, Resource Count = 18854443
validation.thm.model.InterpretRecordLemmaOk (correctness)(Passed) - Duration = 00:00:17.8690448, Resource Count = 18617886
validation.thm.soundness.SemanticSoundnessProof.SoundInSetMemberFalse (correctness)(Passed) - Duration = 00:00:14.7439875, Resource Count = 17814454
validation.thm.soundness.SemanticSoundnessProof.SoundEq (correctness)(Passed) - Duration = 00:00:23.4532100, Resource Count = 16979533
validation.thm.soundness.SemanticSoundnessProof.SoundHasAttr (correctness)(Passed) - Duration = 00:00:14.7334905, Resource Count = 15451607
validation.thm.soundness.SemanticSoundnessProof.SoundContains (correctness)(Passed) - Duration = 00:00:13.0147554, Resource Count = 13884065
validation.thm.model.DecimalComparisonSafe (correctness)(Passed) - Duration = 00:00:10.1198876, Resource Count = 12154290
validation.thm.soundness.SemanticSoundnessProof.SoundLike (correctness)(Passed) - Duration = 00:00:14.4489338, Resource Count = 10872901
Proofs Over threshold: 12
```

After:
```
validation.thm.soundness.SemanticSoundnessProof.SoundHasAttr (correctness)(Passed) - Duration = 00:00:14.1801322, Resource Count = 14727577
[...]

PR #62, opened 2023-07-05
proof sanitization

andrewmwells-amazon:
Improve RUs for `DecimalComparisonSafe` and `CallSafe`

### DIFF
--- a/cedar-dafny/validation/thm/model.dfy
+++ b/cedar-dafny/validation/thm/model.dfy
@@ -840,7 +840,7 @@ module validation.thm.model {
         assert isSafe;
       }
     } else {
-      ListSemantics(args, E);
+      ListSemanticsErr(args, E);
     }
   }
 
--- a/cedar-dafny/validation/thm/model.dfy
+++ b/cedar-dafny/validation/thm/model.dfy
@@ -758,7 +758,16 @@ module validation.thm.model {
     requires IsDecimalComparisonName(name)
     requires ExtensionFunSafeRequires(name, args)
     ensures ExtensionFunSafeEnsures(name, args)
-  {}
+  {
+    var res := extFuns[name].fun(args);
+    match res
+    case Err(e) => {
+      assert e == base.ExtensionError;
+    }
+    case Ok(value) => {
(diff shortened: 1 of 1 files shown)


shaobo-he-aws:
It seems it duplicates functionalities in #56

cedar-dafny/validation/thm/model.dfy
```diff
@@ -758,7 +758,16 @@ module validation.thm.model {
     requires IsDecimalComparisonName(name)
     requires ExtensionFunSafeRequires(name, args)
     ensures ExtensionFunSafeEnsures(name, args)
-  {}
+  {
```
andrewmwells-amazon, line 761: This is greater than 10M according to my VSCode and the internal dashboard but doesn't show up in https://github.com/cedar-policy/cedar-spec/pull/56 for some reason?

andrewmwells-amazon:
> It seems it duplicates functionalities in #56
Agree for `CallSafe`, but I don't understand why `DecimalComparisonSafe` isn't showing up in that PR. I can rebase once John's PR is merged.

shaobo-he-aws:
> > It seems it duplicates functionalities in #56
> 
> Agree for `CallSafe`, but I don't understand why `DecimalComparisonSafe` isn't showing up in that PR. I can rebase once John's PR is merged.

Probably because of the discrepancy of RU costs with different Dafny flags? The set of flags in the Makefile is different from VSCode's.

cedar-dafny/validation/thm/model.dfy
khieta: The internal dashboard is (currently) based on the internal code, so there may be differences there. As far as the difference between VSCode and the metric used for #56 (which is outlined under "Checking Proof Stability" in the README) -- likely just due to different default settings between the command line build and VSCode. 

@emina may know more.

khieta (approved):

cedar-dafny/validation/thm/model.dfy
emina: That and there are some differences between RU counts between VSCode and the command line even when the same flags are used.  We should use the command line counts as the ground truth.

PR #67, opened 2023-07-11
Move most of the remaining generators code into `cedar-policy-generators`

### LINKED ISSUE #46: Separate generators for Cedar types into their own crate
### Category

User level API changes

The logic for generating pseudorandom Cedar policies, hierarchies, and requests, which currently lives in `cedar-drt` alongside a lot of other code (for fuzz targets, differential testing, etc), could be pulled out into its own crate.  This would enable easier reuse of these generators for other purposes. `cedar-drt` would depend on the new crate, which I propose to call `cedar-policy-generators`.

As described above.

`cedar-drt` could export the generators, and would-be consumers could depend on `cedar-drt` instead.  This arrangement would be made a little more complicated by the fact that currently, the generators actually live in the inner crate of `cedar-drt` (its `fuzz` folder).

### Additional context

_No response_

cdisselkoen:
*Issue #, if available:*

Resolves #46.  (We could still consider moving a few other things after this PR, like `dump.rs`, but I think #46 can be considered done after this PR.)

*Description of changes:*

Move most of the remaining generators code into `cedar-policy-generators`.  See discussion on #46.


cdisselkoen:
The build failure here is confusing because it's in a file this PR doesn't touch

shaobo-he-aws:
> The build failure here is confusing because I don't see how it relates to this PR

It seems to be related to the generator macros I wrote. I will take a look at it.

shaobo-he-aws (approved):

andrewmwells-amazon (approved):

PR #73, opened 2023-07-17
[cedar-policy-generators] Add method for starting from existing schema

### LINKED ISSUE #68: Decouple schema-based hierarchy generation from schema generation
### Category

Other

Schema-based hierarchy generation is implemented as a method of the `Schema` struct, which implicitly takes input as the generated schema. We want it to make the parameter explicit and let users provide their schemas. This implies that the current hierarchy generation needs to take cycles in the schemas into account (schema generation do not produce cycles).

See above.

Currently no alternatives.

### Additional context

_No response_

cdisselkoen:
*Issue #, if available:*

Fixes #68 

*Description of changes:*

New methods `Schema::from_nsdef()` and `Schema::from_schemafrag()`. Associated refactors and helper functions.


shaobo-he-aws (approved):
LGTM.

shaobo-he-aws:
Btw, we need to eliminate the chances to generate cycles in the hierarchy since now we allow cycles in entity types.

shaobo-he-aws:
I took the liberty to implement code preventing the generator from generating loops in the entity store that result from self loops in the entity types of a schema. @cdisselkoen could you review it when you have time?

cedar-policy-generators/src/hierarchy.rs
```diff
@@ -5,6 +5,41 @@ use arbitrary::{Arbitrary, Unstructured};
...
+            uids: value,
+        }
+    }
+}
+
+impl<'a> From<&'a EntityUIDs> for &'a [EntityUID] {
+    fn from(value: &'a EntityUIDs) -> Self {
+        &value.uids
+    }
+}
```
cdisselkoen, line 32: I think this should be an `AsRef` rather than a `From`?

cedar-policy-generators/src/hierarchy.rs
```diff
@@ -5,6 +5,41 @@ use arbitrary::{Arbitrary, Unstructured};
...
+
+impl EntityUIDs {
+    // Get an `EntityUID` whose index is greater than `uid`'s
```
cdisselkoen, line 35: This comment implies we're getting a single `EntityUID`, but the function actually returns a slice of `EntityUID`s

andrewmwells-amazon (approved):

Issue #86, opened 2023-07-27
Hierarchy generator generates only a few entities when the exact entity number argument is large


shaobo-he-aws:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/cedar-policy/cedar/issues?q=is%3Aissue+).
- [X] I have read the guide for [submitting bug reports](https://github.com/cedar-policy/cedar/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### Bug Category

Other

### Describe the bug

Only around 5 entities are generated for the TinyTodo example when `--num-entities` is set to 500.

It seems the probably of generating `Eid`s with empty strings is very high. The program below illustrates the bias.

```Rust
use rand::{thread_rng, RngCore};
use arbitrary::Unstructured;

fn main() {
    let mut bytes = [0 as u8; 4096];
    let mut rng = thread_rng();
    rng.fill_bytes(&mut bytes);
    println!("{bytes:?}");
    let mut u = Unstructured::new(&bytes);
    let mut strs: Vec<String> = Vec::new();
    for _ in 0..512 {
      strs.push(u.arbitrary().unwrap());
    }
    for s in strs {
      println!("len: {},", s.len());
    }
}
```
When I run it on my laptop, only one of 512 generated strings is not empty.

### Expected behavior

The number of generate entities should respect the `--num-entities` flag.

### Reproduction steps

N/A

### Code Snippet

```rust
// Put your code below this line.

```

### Log output

```
// Put your output below this line

```

shaobo-he-aws:
Looking at the relevant code in the Arbitrary crate, it seems the observation is justified. Its generation strategy is to get some random bytes and try to produce a utf-8 string out of them; if it fails, it tries to get a slice of the bytes that could produce a valid string, which could easily be an empty slice.

shaobo-he-aws:
Fixed via #87

PR #87, opened 2023-07-27
Use nanoids as Eids for schema-based hierarchy generation

shaobo-he-aws:
*Issue #, if available:* #86

*Description of changes:*


cdisselkoen (changes requested):
This looks great, I'm just requesting one change -- can we get a separate setting to control whether ids are generated via `Arbitrary` or `nanoid`, rather than reusing the setting for schema-based generation vs not?

aaronjeline (approved):
Just making sure I'm correct here: We don't actually care about the security properties of these IDs, we're just using them as an easy way to generate IDs right?

shaobo-he-aws:
> 

Correct.

khieta (approved):

khieta:
Approved, but I imagine you'll have to fix the CI break before merging.

cdisselkoen (approved):

andrewmwells-amazon (approved):

PR #116, opened 2023-09-07
Add two options to the generator

shaobo-he-aws:
`enable_unspecified_apply_spec` enables/disables unspecified entities where actions can apply. Enabling it causes `ActionType::applies_to` to match `Some(ApplySpec {resource_types: Some(_), principal_types: Some(_), context: _})`. `enable_action_in_constraints` enables/disables action constraints in forms of `in` operations.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cedar-policy-generators/src/schema.rs
+++ b/cedar-policy-generators/src/schema.rs
@@ -640,7 +640,7 @@ impl Schema {
         };
         let mut principal_types = HashSet::new();
         let mut resource_types = HashSet::new();
-        // optionally return a list of entity types and add them to `tys` in the same time
+        // optionally return a list of entity types and add them to `tys` at the same time
         let pick_entity_types = |tys: &mut HashSet<Name>, u: &mut Unstructured<'_>| {
             Result::Ok(if u.ratio::<u8>(1, 4)? {
                 // The action applies to an unspecified
--- a/cedar-policy-generators/src/schema.rs
+++ b/cedar-policy-generators/src/schema.rs
@@ -29,7 +29,7 @@ use crate::settings::ABACSettings;
 use crate::size_hint_utils::{size_hint_for_choose, size_hint_for_range, size_hint_for_ratio};
 use crate::{accum, gen, gen_inner, uniform};
 use arbitrary::{self, Arbitrary, Unstructured};
-use cedar_policy_core::ast::{self, Effect, PolicyID};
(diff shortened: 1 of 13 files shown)


andrewmwells-amazon (approved):

cedar-policy-generators/src/schema.rs
```diff
@@ -640,74 +640,84 @@ impl Schema {
         };
         let mut principal_types = HashSet::new();
         let mut resource_types = HashSet::new();
+        // optionally return a list of entity types and add them to `tys` in the same time
```
andrewmwells-amazon, line 643: [nit] "in the same" -> "at the same"

cedar-policy-generators/src/schema.rs
```diff
@@ -640,74 +640,84 @@ impl Schema {
...
+                // resource and no other entity types.
+                None
+            } else {
+                // for each entity type, flip a coin to see
+                // whether to include it as a possible
+                // principal type for this action
+                Some(
+                    entity_types
+                        .iter()
+                        .filter_map(|(name, _)| match u.ratio::<u8>(1, 2) {
```
andrewmwells-amazon, line 656: [nit] Maybe worth using named constants for these

cedar-policy-generators/src/schema.rs
```diff
@@ -640,74 +640,84 @@ impl Schema {
...
-                                                Ok(false) => None,
-                                                Err(_) => None,
-                                            })
-                                            .collect(),
-                                    )
-                                },
-                                context: arbitrary_attrspec(&settings, &entity_type_names, u)?,
-                            })
+                        applies_to: {
+                            let apply_spec = if u.ratio::<u8>(1, 8)? {
```
andrewmwells-amazon, line 680: [nit] same here

andrewmwells-amazon (approved):

cedar-policy-generators/src/schema.rs
shaobo-he-aws: Good catch.

cedar-policy-generators/src/schema.rs
shaobo-he-aws: I think I'll keep this way because we've got ratios like this in several places.

cdisselkoen (approved):

PR #124, opened 2023-09-26
Make it easier to test new implementations of Cedar

khieta:
*Issue #, if available:*

N/A

*Description of changes:*

This PR does some refactoring to make it easier to plug in new implementations of Cedar to DRT. The core of the change is moving the content of `cedar-drt/src/lib.rs` into two new files: `cedar-drt/src/cedar_test_impl.rs` and `cedar-drt/src/dafny_java_impl.rs`. The former defines a trait `CedarTestImplementation` that provides `is_authorized`, `interpret`, and `validate` methods. The latter provides an implementation of this trait for our Cedar Java engine extracted from Dafny. 

The differential testing functions in `cedar-drt/fuzz/src/lib.rs` are all updated to take in a `&impl CedarTestImplementation` as input. So, in theory, all a user needs to do to set up DRT for a custom version of Cedar is to provide an implementation for the `CedarTestImplementation` trait and change 2-3 LOC in the fuzz target to use this new implementation.

The diff for this PR is large, but it is primarily just the result of moving code around via copy & paste.

Other changes:
* Changed a few more instances of "query" (our old terminology) to "request".
* Various renaming for consistency.
* Updated the `cedar` submodule. This was not necessary for the PR to build -- I just figured we'd want to be up to date.

I ran each of the DRT targets for ~10 minutes locally to confirm that I didn't break anything obvious.


cedar-drt/fuzz/src/lib.rs
```diff
@@ -33,180 +31,158 @@ use cedar_policy_core::extensions::Extensions;
...
+        Ok(v) => Some(v),
+        Err(e) if matches!(e.error_kind(), EvaluationErrorKind::IntegerOverflow(_)) => {
+            // ignore the input if it results in an integer overflow error
+            return;
+        }
+        Err(_) => None,
+    };
+    // custom_impl.eval() returns true when the result of evaluating expr
+    // matches the expected value v
+    assert!(custom_impl.interpret(request, entities, expr, v))
```
cdisselkoen, line 63: the comment is about `custom_impl.eval()`, but the code uses `custom_impl.interpret()`

cedar-drt/fuzz/src/lib.rs
```diff
@@ -33,180 +31,158 @@ use cedar_policy_core::extensions::Extensions;
...
-    authorizer: Authorizer,
-    /// Definitional engine instance
-    def_engine: DefinitionalEngine<'e>,
-    /// Definitional validator instance
-    def_validator: DefinitionalValidator<'e>,
+/// Compare the behavior of the evaluator in cedar-policy against a custom Cedar
+/// implementation. Panics if the two do not agree. `expr` is the expression to
+/// evaluate and `request` and `entities` are used to populate the evaluator.
+pub fn run_eval_test(
+    custom_impl: &dyn CedarTestImplementation,
```
cdisselkoen, line 38: does this need to be `&dyn`?  here and in the similar methods.  would `&impl` just work as a drop-in replacement?

cedar-drt/fuzz/src/lib.rs
```diff
@@ -33,180 +31,158 @@ use cedar_policy_core::extensions::Extensions;
...
+    schema: ValidatorSchema,
+    policies: &PolicySet,
+    mode: ValidationMode,
+) {
+    let validator = Validator::new(schema.clone());
+    let (rust_res, rust_validation_dur) = time_function(|| validator.validate(policies, mode));
+    info!("{}{}", RUST_VALIDATION_MSG, rust_validation_dur.as_nanos());
+
+    let definitional_res = custom_impl.validate(schema.clone(), policies, mode);
```
cdisselkoen, line 139: does `custom_impl` really need to take an owned `schema` here?

cedar-drt/fuzz/src/lib.rs
```diff
@@ -33,180 +31,158 @@ use cedar_policy_core::extensions::Extensions;
...
+
+    assert!(
+        definitional_res.parsing_succeeded(),
+        "Dafny json parsing failed for:\nPolicies:\n{}\nSchema:\n{:?}",
```
cdisselkoen, line 143: Error message probably shouldn't refer to Dafny, since this function is generic over all `CedarTestImplementation`s

cedar-drt/fuzz/src/lib.rs
```diff
@@ -33,180 +31,158 @@ use cedar_policy_core::extensions::Extensions;
...
-        .into();
-        assert_eq!(
-            InterfaceResponse::from(rust_res_for_comparison),
-            definitional_res,
-            "Mismatch for {q}\nPolicies:\n{}\nEntities:\n{}",
-            &policies,
-            &entities
-        );
-        ret
+    // Temporary fix to ignore a known mismatch between the Dafny and Rust.
```
cdisselkoen, line 148: I had forgotten (or never knew?) that we had this temporary fix in place.  Do we have an issue to fix this?

cedar-drt/src/cedar_test_impl.rs
```diff
@@ -0,0 +1,77 @@
...
+    ) -> InterfaceResponse;
+
+    /// Custom evaluator entry point. The bool return value indicates the whether
+    /// evaluating the provided expression produces the expected value.
+    fn interpret(
+        &self,
+        request: &Request,
+        entities: &Entities,
+        expr: &Expr,
+        expected: Option<Value>,
```
cdisselkoen, line 45: What does `None` mean here? might be worth a comment

cedar-drt/src/cedar_test_impl.rs
```diff
@@ -0,0 +1,77 @@
...
+    ) -> bool;
+
+    /// Custom validator entry point.
+    fn validate(
+        &self,
+        schema: ValidatorSchema,
```
cdisselkoen, line 51: does this have to be owned?

cedar-drt/src/cedar_test_impl.rs
```diff
@@ -0,0 +1,77 @@
...
+impl ValidationInterfaceResponse {
+    pub fn validation_passed(&self) -> bool {
+        self.validation_errors.is_empty()
+    }
+
+    pub fn parsing_succeeded(&self) -> bool {
+        self.parse_errors.is_empty()
+    }
+
+    pub fn contains_error(&self, s: String) -> bool {
```
cdisselkoen, line 74: could `s` be `&str` here? or at the very minimum, `&String` to avoid copying the string?

cedar-drt/src/dafny_java_impl.rs
```diff
@@ -0,0 +1,446 @@
...
+    auth_nanos: i64,
+    response: InterfaceResponse,
+}
+
+#[derive(Debug, Serialize)]
+struct EvalRequestForDefEngine<'a> {
+    request: &'a ast::Request,
+    entities: &'a Entities,
+    expr: &'a ast::Expr,
+    expected: Option<ast::Expr>,
```
cdisselkoen, line 74: can this be `Option<&'a ast::Expr>`?

cedar-drt/src/dafny_java_impl.rs
```diff
@@ -0,0 +1,446 @@
...
+    matches: bool,
+}
+
+fn make_expr(v: Value) -> Expr {
+    PartialValue::Value(v).into()
+}
+
+#[derive(Debug, Serialize)]
+struct RequestForDefValidator<'a> {
+    schema: ValidatorSchema,
```
cdisselkoen, line 89: can this be `&'a ValidatorSchema`?

cedar-drt/src/dafny_java_impl.rs
```diff
@@ -0,0 +1,446 @@
...
+
+#[derive(Debug, Serialize, Deserialize, Clone, Copy)]
+#[repr(transparent)]
+struct DefinitionalEvalResponse {
+    matches: bool,
+}
+
+fn make_expr(v: Value) -> Expr {
+    PartialValue::Value(v).into()
+}
```
cdisselkoen, line 85: I don't think you need this, since Core defines `From<Value> for Expr`?

cedar-drt/src/dafny_java_impl.rs
```diff
@@ -0,0 +1,446 @@
...
+    /// `policies`, and `entities`
+    pub fn is_authorized(
+        &self,
+        request: &ast::Request,
+        policies: &ast::PolicySet,
+        entities: &Entities,
+    ) -> InterfaceResponse {
+        let (jstring, dur) =
+            time_function(|| self.serialize_auth_request(request, policies, entities));
+        info!("{}{}", RUST_SERIALIZATION_MSG, dur.as_nanos());
```
cdisselkoen, line 273: is `RUST_SERIALIZATION_MSG` correct here?

is it actually helpful to define the message as a constant, or should we inline it here for clarity?

cedar-drt/tests/integration_tests.rs
```diff
@@ -0,0 +1,80 @@
...
+use cedar_policy::integration_testing::{
+    perform_integration_test_from_json_custom, resolve_integration_test_path, CustomCedarImpl,
+};
+
+use cedar_drt::*;
+use std::path::Path;
+
+// Note: Some of the following code is based on cedar-policy/tests/corpus_tests.rs, but
+// there probably isn't enough commonality left to be worth trying to share the
+// code.
```
cdisselkoen, line 29: in fact, I think we can just remove this comment at this point -- almost all the sharable code is shared

cdisselkoen (approved):
Overall looks good, just a couple comments

cedar-drt/fuzz/src/lib.rs
khieta: `&impl` does indeed work as a drop-in replacement. Thanks for pointing that out.

cedar-drt/fuzz/src/lib.rs
khieta: I had the same reaction, but then I figured that the original author (possibly me) probably had a good reason 😆 I'll look into this more later today/tomorrow.

cedar-drt/src/cedar_test_impl.rs
khieta: I went with `&String` because that's the type the `contains()` function expects.

khieta:
Thanks for the feedback @cdisselkoen! I addressed some of your comments in my last commit; I'll take care of the rest later today or tomorrow.

cedar-drt/src/cedar_test_impl.rs
cdisselkoen: Sounds good

aaronjeline (approved):
Seems great, also includes a bunch of nice code cleanup that needed to happen.

cedar-drt/src/dafny_java_impl.rs
khieta: Yes, `RUST_SERIALIZATION_MSG` is correct. It refers to the time taken to produce a JSON version of the input for the Java engine to consume. Not too important to have it as a constant, although it does help ensure consistency between its use in `is_authorized` and `validate`.

My latest commit moves these constants from `logger.rs` to `dafny_java_impl.rs` since that's the only place they're actually used. I also added some comments to explain their purpose.

cedar-drt/fuzz/src/lib.rs
khieta: Confirmed this is still a difference between the Rust & Dafny. Will file an issue to track.

khieta:
Filed three new issues as a result of this PR: #125, #126, and #127. I think I've addressed all the comments now. Let me know if I missed anything!

cedar-drt/fuzz/src/lib.rs
```diff
@@ -33,180 +31,158 @@ use cedar_policy_core::extensions::Extensions;
...
+    enable_extensions: bool,
+) {
+    let exts = if enable_extensions {
+        Extensions::all_available()
+    } else {
+        Extensions::none()
+    };
+    let eval = match Evaluator::new(request, entities, &exts) {
+        Ok(e) => e,
+        Err(_) => return, // FOR NOW just ignore errors in the restricted exprs
```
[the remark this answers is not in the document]
khieta: Filed #127. Feel free to edit my description or add additional comments on the issue.

cdisselkoen (approved):

aaronjeline (approved):

PR #128, opened 2023-10-02
Make the generator to produce special `i64` values easier

shaobo-he-aws:
This commit also uses library implementations of the `Arbitrary` trait for IP addresses.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cedar-policy-generators/src/abac.rs
+++ b/cedar-policy-generators/src/abac.rs
@@ -120,11 +120,38 @@ pub struct ConstantPool {
     string_constants: Vec<SmolStr>,
 }
 
+#[derive(Debug, Clone, Copy)]
+struct BiasedI64(i64);
+
+impl<'a> Arbitrary<'a> for BiasedI64 {
+    fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
+        Ok(gen!(u,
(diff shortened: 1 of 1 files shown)


cedar-policy-generators/src/abac.rs
```diff
@@ -120,11 +120,38 @@ pub struct ConstantPool {
...
 
+#[derive(Debug, Clone, Copy)]
+struct BiasedI64(i64);
+
+impl<'a> Arbitrary<'a> for BiasedI64 {
+    fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
+        Ok(gen!(u,
+            1 => std::i64::MAX,
+            1 => std::i64::MIN,
+            8 => <i64 as Arbitrary>::arbitrary(u)?
```
cdisselkoen, line 131: more "special" values we might want to bias for include 0, 1, and -1

cdisselkoen (approved):

cedar-policy-generators/src/abac.rs
shaobo-he-aws: I ended up adding `-1` and `0` because 1 should be easily obtained by mutating 0.

aaronjeline (approved):

PR #137, opened 2023-10-25
Aaronjeline/logmore

aaronjeline:
*Issue #, if available:*
N/A
*Description of changes:*
Provides the ability to log the corpus generated.


shaobo-he-aws (approved):

cedar-drt/fuzz/fuzz_targets/abac-type-directed.rs
```diff
@@ -28,22 +28,27 @@ use cedar_policy_generators::{
...
     /// Should we pre-evaluate entity attributes
+    #[serde(skip)]
     pub should_cache_entities: bool,
     /// generated policy
     pub policy: ABACPolicy,
     /// the requests to try for this hierarchy and policy. We try 8 requests per
     /// policy/hierarchy
+    #[serde(skip)]
     pub requests: [ABACRequest; 8],
 }
```
cdisselkoen, line 53: I would expect `serialize` to include most of these fields.  Can we just call `serialize` on `.policy` if this is what we want?

cedar-drt/fuzz/src/abac_type_directed.rs
```diff
@@ -0,0 +1,116 @@
+
```
cdisselkoen, line 1: committing a second copy of this file?

cedar-drt/README.md
```diff
@@ -22,3 +22,9 @@ The table below lists all available fuzz targets, including which component of t
 | [`validation-drt`](fuzz/fuzz_targets/validation-drt.rs) | Validator | DRT | Diff test permissive validation |
 | [`validation-pbt`](fuzz/fuzz_targets/validation-pbt.rs) | Validator | PBT | Test that validated policies do not result in type errors |
 | [`wildcard-matching`](fuzz/fuzz_targets/wildcard-matching.rs) | String matching algorithm used for the `like` operator | DRT | Diff test wildcard matching using a regex-based implementation |
+
+
+## Logging
+
+If the fuzz targets are compiled with the `log` features, then they will log their entire corpus to the file pointed at in the `LOGFILE` environment variable.
```
cdisselkoen, line 29: idk if "corpus" is the right word here -- does "corpus" imply some amount of minimization?

cedar-drt/fuzz/fuzz_targets/abac-type-directed.rs
aaronjeline: Nah, because I need it to be the FuzzTargetInput so I can log all of them via the macro easily.

cedar-drt/README.md
aaronjeline: I don't think so, "corpus minimization" is a second step one applies to the corpus

andrewmwells-amazon (approved):

PR #145, opened 2023-11-01
Update to generate `is` operator

john-h-kastner-aws:
Update for https://github.com/cedar-policy/cedar/pull/396. Generators may generate `is` based on an `enable_is` flag. This is disabled for differential testing because `is` is not modeled in Dafny, but it is enabled for property tests.


aaronjeline (approved):

cedar-policy-generators/src/settings.rs
```diff
@@ -55,6 +55,9 @@ pub struct ABACSettings {
     /// Flag to globally enable/disable generation of expressions containing the
     /// `like` operator.
     pub enable_like: bool,
+    /// Flag to globally enable/disable generation of expressions containing the
+    /// `is` operator.
+    pub enable_is: bool,
```
shaobo-he-aws, line 60: Do we need this flag? `is` should be enabled for all targets?

john-h-kastner-aws: for the moment, yes. We're not modeling it in dafny, so I need to turn it off for difftesting

shaobo-he-aws: Then I propose we update and merge it once the definitional model is done otherwise this PR essentially does nothing.

john-h-kastner-aws: I need to turn it off for difftesting, but not property testing on the Rust only. So, the soundness PBT and formatting targets are run with `enable_is: true`

shaobo-he-aws: I see.

cedar-drt/fuzz/fuzz_targets/validation-pbt.rs
```diff
@@ -143,6 +144,9 @@ fn log_err<T>(res: Result<T>, doing_what: &str) -> Result<T> {
             Err(Error::LikeDisabled) => {
                 checkpoint(LOG_FILENAME_ERR_LIKE_DISABLED.to_string() + "_" + doing_what)
             }
+            Err(Error::IsDisabled) => {
+                checkpoint(LOG_FILENAME_ERR_LIKE_DISABLED.to_string() + "_" + doing_what)
```
khieta, line 148: Presumably should be `LOG_FILENAME_ERR_IS_DISABLED`?

cedar-policy-generators/src/expr.rs
```diff
@@ -472,6 +482,21 @@ impl<'a> ExprGenerator<'a> {
...
+                                Ok(ast::Expr::is_entity_type(
+                                    self.generate_expr_for_type(
+                                        &Type::entity(),
+                                        max_depth - 1,
+                                        u,
+                                    )?,
+                                    u.choose(&self.schema.entity_types)?.clone(),
+                                ))
+                            } else {
+                                Err(Error::LikeDisabled)
```
khieta, line 497:
```suggestion
                                Err(Error::IsDisabled)
```

khieta (changes requested):
This PR only updates expression generation. I think you want to update the generation of head constraints as well (look for instances of `PrincipalOrResourceConstraint` in  the `cedar-policy-generators` crate)

PR #151, opened 2023-11-13
Change log to log ESTs

aaronjeline:
*Issue #, if available:*
N/A
*Description of changes:*
Before the logging option logged the debug printer form of policies. What we actually want is to log the EST format so we can read it back in.

### DIFF
--- a/cedar-policy-generators/src/policy.rs
+++ b/cedar-policy-generators/src/policy.rs
@@ -3,10 +3,10 @@ use crate::err::Result;
 use crate::hierarchy::Hierarchy;
 use crate::size_hint_utils::size_hint_for_ratio;
 use arbitrary::{Arbitrary, Unstructured};
-use cedar_policy_core::ast;
 use cedar_policy_core::ast::{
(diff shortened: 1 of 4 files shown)


cedar-policy-generators/src/policy.rs
```diff
@@ -30,6 +30,14 @@ pub struct GeneratedPolicy {
     abac_constraints: Expr,
 }
 
+impl From<GeneratedPolicy> for est::Policy {
+    fn from(gp: GeneratedPolicy) -> est::Policy {
+        let sp: StaticPolicy = gp.into();
+        let p: Policy = sp.into();
+        p.into()
```
cdisselkoen, line 37: this is simultaneously beautiful and gross

cdisselkoen (approved):

andrewmwells-amazon (approved):

cedar-policy-generators/src/policy.rs
aaronjeline: I mean 35 and 36 compile into no-ops, but I do wish the type system could figure out transitive into's

cdisselkoen (approved):

PR #152, opened 2023-11-13
Updating to breaking upstream changes

aaronjeline:
*Issue #, if available:*
N/A
*Description of changes:*
Just updating to some breaking API changes in 3.0


cdisselkoen (changes requested):
I'm pretty sure this actually reverts to the old (pre-3.0) APIs?  I think we want the version that's currently checked in if we want to be compatible with 3.0.  E.g., this partially reverts #148.

aaronjeline:
> I'm pretty sure this actually reverts to the old (pre-3.0) APIs

Maybe my local env is messed up

PR #155, opened 2023-11-16
Followup to #145

khieta:
*Issue #, if available:*

*Description of changes:*

Followup to #145, which updated generators with support for the `is` operator. This PR adds support for generating `is` in scope constraints to `schema.rs` (which is used for most DRT targets). It mirrors the corresponding change made previously in `policy.rs` (which is used for the `rbac` target).

### DIFF
--- a/cedar-policy-generators/src/policy.rs
+++ b/cedar-policy-generators/src/policy.rs
@@ -292,7 +292,7 @@ impl PrincipalOrResourceConstraint {
                     1 => Ok(Self::IsTypeInSlot(hierarchy.arbitrary_entity_type(u)?))
                 )
             } else {
-                // 32% Eq, 32% In (16% with and without Is), 32% Is (16% with and without In)
+                // 32% Eq, 16% In, 16% Is, 16% IsIn
                 let uid = hierarchy.arbitrary_uid(u)?;
                 gen!(u,
                     2 => Ok(Self::Eq(uid)),
--- a/cedar-policy-generators/src/schema.rs
+++ b/cedar-policy-generators/src/schema.rs
@@ -1039,7 +1039,9 @@ impl Schema {
             Ok(PrincipalOrResourceConstraint::NoConstraint)
         } else {
             // 32% Eq, 16% In, 16% Is, 16% IsIn
-            let uid = self.exprgenerator(Some(hierarchy)).arbitrary_principal_uid(u)?;
+            let uid = self
+                .exprgenerator(Some(hierarchy))
+                .arbitrary_principal_uid(u)?;
             let ety = u.choose(self.entity_types())?.clone();
             gen!(u,
                 2 => Ok(PrincipalOrResourceConstraint::Eq(uid)),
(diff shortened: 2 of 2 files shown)


cedar-policy-generators/src/schema.rs
```diff
@@ -1034,15 +1034,20 @@ impl Schema {
...
+            let ety = u.choose(self.entity_types())?.clone();
+            gen!(u,
+                2 => Ok(PrincipalOrResourceConstraint::Eq(uid)),
+                1 => Ok(PrincipalOrResourceConstraint::In(uid)),
+                1 => Ok(PrincipalOrResourceConstraint::IsType(ety)),
+                1 => Ok(PrincipalOrResourceConstraint::IsTypeIn(ety, uid))
+            )
+        }
     }
     fn arbitrary_principal_constraint_size_hint(depth: usize) -> (usize, Option<usize>) {
```
khieta, line 1052: I'm not sure whether/how the `*_size_hint` functions need to be updated. Suggestions welcome.

john-h-kastner-aws (approved):

andrewmwells-amazon (approved):

cedar-policy-generators/src/schema.rs
cdisselkoen: They are mostly performance optimizations; see docs on [Arbitrary::size_hint](https://docs.rs/arbitrary/latest/arbitrary/trait.Arbitrary.html#method.size_hint).  If/when we migrate to Bolero, they will go away.

cedar-policy-generators/src/policy.rs
```diff
@@ -292,7 +292,7 @@ impl PrincipalOrResourceConstraint {
                     1 => Ok(Self::IsTypeInSlot(hierarchy.arbitrary_entity_type(u)?))
                 )
             } else {
-                // 32% Eq, 32% In (16% with and without Is), 32% Is (16% with and without In)
+                // 32% Eq, 16% In, 16% Is, 16% IsIn
```
cdisselkoen, line 295: doesn't add up to 100%

cedar-policy-generators/src/schema.rs
```diff
@@ -1034,15 +1034,22 @@ impl Schema {
...
-            self.exprgenerator(Some(hierarchy)).arbitrary_principal_uid(u)?,
-        )))
+        // 20% of the time, NoConstraint
+        if u.ratio(1, 5)? {
+            Ok(PrincipalOrResourceConstraint::NoConstraint)
+        } else {
+            // 32% Eq, 16% In, 16% Is, 16% IsIn
+            let uid = self
+                .exprgenerator(Some(hierarchy))
+                .arbitrary_principal_uid(u)?;
```
cdisselkoen, line 1044: generates the `uid` even in the `IsType` case where it isn't used.  Probably ok though

cdisselkoen (approved):

cedar-policy-generators/src/policy.rs
john-h-kastner-aws: +20% no constraint

cedar-policy-generators/src/schema.rs
khieta: So it's fine to leave these functions as-is?

PR #166, opened 2023-12-07
Lean parser refactor

khieta:
*Issue #, if available:*

* Fixed a couple bugs found while doing preliminary diff testing (prior code didn't support unspecified entity types & used the wrong argument name for `MulByConst`)
* Refactored the parser to use a `Result` type & monad operations instead of `panic!` -- this makes it easier to pass the error messages back to Rust during DRT
* Added a Cli command for evaluation (used during debugging)

*Description of changes:*


cedar-lean/DiffTest/Main.lean
```diff
@@ -30,26 +30,39 @@ open Cedar.Spec
...
+      let policies ← getJsonField json "policies" >>= jsonToPolicies
+      let schema ← getJsonField json "schema" >>= jsonToSchema
+      .ok (validate policies schema)
+  toString (Lean.toJson result)
+
+@[export evaluateDRT] def evaluateDRT (_req : String) : String :=
+  panic! "TODO: implement evaluateDRT"
+
+-- variant of `evaluateDRT` that returns the resulting value; used in the Cli
+def evaluate (req : String) : String :=
```
cdisselkoen, line 57: what's the exact difference between this and `evaluateDRT`? this says that it returns the resulting value, but both `evaluate` and `evaluateDRT` return a `String`

cedar-lean/DiffTest/Main.lean
```diff
@@ -30,26 +30,39 @@ open Cedar.Spec
...
+  let result : ParseResult (Result Value) :=
+    match Lean.Json.parse req with
+    | .error e => .error s!"evaluate: failed to parse input: {e}"
+    | .ok json => do
+      let expr ← getJsonField json "expr" >>= jsonToExpr
+      let request ← getJsonField json "request" >>= jsonToRequest
+      let entities ← getJsonField json "entities" >>= jsonToEntities
+      .ok (Cedar.Spec.evaluate expr request entities)
+  s!"{repr result}" -- use repr to display the value as a string
```
cdisselkoen, line 66: would it make more sense to return `Result Value` or something, and leave it to the caller to make the `repr` call?

cedar-lean/DiffTest/Parser.lean
```diff
@@ -36,296 +36,334 @@ open Cedar.Spec
...
-  let attrsKVs := jsonObjToKVList (getJsonField json "attrs")
-  let attrs := Map.mk (List.map (λ (k,v) => (k,jsonToValue v)) attrsKVs)
-  {
-    ancestors := ancestors,
-    attrs := attrs
+def jsonToEntityData (json : Lean.Json) : ParseResult EntityData := do
+  let ancestorsArr ← getJsonField json "ancestors" >>= jsonToArray
+  let ancestors ← List.mapM jsonToEuid ancestorsArr.toList
+  let attrsKVs ← getJsonField json "attrs" >>= jsonObjToKVList
+  let attrs ← List.mapM (λ (k,v) => jsonToValue v >>= λ v => .ok (k,v)) attrsKVs
```
cdisselkoen, line 268: is there not a shorthand for `λ v => .ok (k,v)`? in Haskell this would be `return (k,v)`

cdisselkoen: this occurs at least one other place as well

cedar-lean/DiffTest/Parser.lean
```diff
@@ -36,296 +36,334 @@ open Cedar.Spec
...
   }
 
-def jsonToEntities (json : Lean.Json) : Entities :=
-  let entities := getJsonField json "entities"
-  let kvs := jsonArrayToKVList entities
-  Map.mk (List.map (λ (k,v) => (jsonToEuid k, jsonToEntityData v)) kvs)
+def jsonToEntities (json : Lean.Json) : ParseResult Entities := do
+  let entities ← getJsonField json "entities"
+  let kvs_json ← jsonArrayToKVList entities
+  let kvs ← List.mapM (λ (k,v) => jsonToEuid k >>= λ k => jsonToEntityData v >>= λ v => .ok (k, v)) kvs_json
```
cdisselkoen, line 277: instead of this string of `>>= λ`, could we use a nested `do` block, or does that not work in Lean?

cdisselkoen (approved):

andrewmwells-amazon (approved):

cedar-lean/Cli/Main.lean
```diff
@@ -43,5 +43,8 @@ def main (args : List String) : IO Unit :=
       | "validate" =>
         let response := validateDRT request
         IO.println response
-      | _ => printUsage s!"Invalid command `{command}` (expected `authorize` or `validate`)"
+      | "evaluate" =>
+        let response := evaluate request
+        IO.println response
+      | _ => printUsage s!"Invalid command `{command}` (expected `authorize`, `validate`, or `evaluate`)"
```
andrewmwells-amazon, line 49: [nit] oxford commas waste electricity

andrewmwells-amazon, line 49:
```suggestion
      | _ => printUsage s!"Invalid command `{command}` (expected `authorize`, `validate` or `evaluate`)"
```

cedar-lean/DiffTest/Parser.lean
```diff
@@ -36,296 +36,334 @@ open Cedar.Spec
...
-def jsonToEuid (json : Lean.Json) : EntityUID :=
-  let eid := jsonToString (getJsonField json "eid")
-  let ty := jsonToEntityType (getJsonField json "ty")
-  {
+def jsonToEntityType (json : Lean.Json) : ParseResult EntityType := do
+  let (tag, body) ← unpackJsonSum json
+  match tag with
+  | "Specified" => jsonToName body
+  | "Unspecified" => 
+    -- "unspecified" entities are treated as normal entities with a unique name
```
andrewmwells-amazon, line 53:
```suggestion
    -- "Unspecified" entities are treated as normal entities with a unique name
```

cedar-lean/Cli/Main.lean
khieta: 🤣

andrewmwells-amazon (approved):

cedar-lean/Cli/Main.lean
khieta: I'll humor you, but I like oxford commas

cedar-lean/DiffTest/Main.lean
khieta: Revised in the latest commit.

cedar-lean/Cli/Main.lean
cdisselkoen: I also vote for oxford commas

cedar-lean/DiffTest/Main.lean
khieta: The difference was that `evaluate` returns the raw string representation of a value, while `evaluateDRT` should return some proper json value. I've switched to returning the value directly from `evaluate`.

cedar-lean/DiffTest/Parser.lean
khieta: Added two new functions (`mapMValues` and `mapMKeysAndValues`) in the latest commit to minimize the icky notation. Also switched to using the `do` notation to avoid confusion.

shaobo-he-aws (approved):

PR #170, opened 2023-12-08
Lean fuzz targets

khieta:
*Issue #, if available:*

*Description of changes:*

This PR adds new `-lean` fuzz targets that use the Lean harness. The new targets are exact copies of the old targets, but with `JavaDefinitionalEngine` replaced with `LeanDefinitionalEngine`. I do not expect to maintain two copies for long — this is just an interim solution that will allow us to start testing the Lean code in CI without breaking our current infrastructure.

This PR copies the following targets:

* abac.rs
* abac-type-directed.rs
* eval-type-directed.rs
* rbac.lean
* rbac-authorizer.rs
* strict-validation-drt-type-directed.rs

It does not copy the following targets:

* formatter.rs
* partial-eval.rs
* pp.rs
* simple-parser.rs
* wildcard-matching.rs
* validation-pbt.rs
* validation-drt.rs
* validation-drt-type-directed.rs

The first 6 aren’t relevant because they are property-based tests, and the last two aren’t relevant because they test permissive validation, which is not encoded in the Lean spec.

All targets (except `rbac-lean`) will run for a few minutes without errors, but there are still kinks to work out so I’d expect them to fail for longer runs. The reason the `rbac-lean` target fails quickly is because the Lean spec doesn't currently support templates.


shaobo-he-aws:
I wonder instead of having dedicated targets. Can we add a command line option or a feature?

aaronjeline (approved):

andrewmwells-amazon (changes requested):
This duplicates a ton of code. Instead, I think we should have a configuration flag to run with `LeanDefinitionalEngine` or `JavaDefinitionalEngine` (or both?).

khieta:
By popular demand, I've revised the PR to remove duplicate code. I made an `abac_shared.rs` file with common definitions and revised `abac.rs` and `abac_lean.rs` to use this shared code (and similarly for other duplicated targets). I didn't rename `abac.rs` to `abac_dafny.rs` to avoid breaking our current (internal) CI.

@shaobo-he-aws and @andrewmwells-amazon let me know what you think. Open to other suggestions.

khieta:
Also, to respond to @shaobo-he-aws's question,
> I wonder instead of having dedicated targets, if we can add a command line option or a feature?

This will be problematic for our (internal) CI that expects to run a command of the form `cargo fuzz run TEST_NAME`. I appreciate that multiple targets is ugly, but I think it's worth to avoid having to muck with the CI (for now). If you feel strongly, I can reconsider.

khieta:
I should also note: once we deprecate the Dafny, I expect to switch back to just `abac.rs` instead of `abac_shared.rs` + `abac.rs` (and similarly for the other targets). Possibly at that point we can consider adding a single place to specify the definitional engine for all targets.

andrewmwells-amazon (approved):

andrewmwells-amazon:
LGTM

shaobo-he-aws (approved):

Issue #174, opened 2023-12-11
Add support for templates to the Lean formalization


khieta:
### Category

Lean formalization

### Describe the feature you'd like to request

Add the notion of templates to the Lean formalization to allow using DRT targets that use template-linked policies (for now, just `rbac`). See the Dafny formalization for reference: https://github.com/cedar-policy/cedar-spec/blob/main/cedar-dafny/def/templates.dfy

Note that the way we support templates in Dafny does not impact proofs. During DRT, we parse templates & links from the JSON input, instantiate all policies, and then call the usual `is_authorized`/`validate` functions.

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

PR #179, opened 2023-12-14
Add templates to Lean

khieta:
*Issue #, if available:*

#174 

*Description of changes:*

This PR adds support for templates to the Lean formalization, following the [model in Dafny](https://github.com/cedar-policy/cedar-spec/blob/main/cedar-dafny/def/templates.dfy). I've run the `rbac-lean` fuzz target (which generates template-linked policies) for ~30 minutes with no failures.

There may be some CI issues to smooth out, but happy to accept reviews on the Lean formalization while I work on those.


cedar-lean/DiffTest/Parser.lean
```diff
@@ -463,7 +475,7 @@ partial def jsonToEntityOrRecordType (json : Lean.Json) : ParseResult CedarType
     let lub ← Array.mapM jsonToName lubArr
     if lub.size == 1
     then .ok (.entity lub[0]!)
-    else .error s!"jsonToEntityOrRecordType: expected lub to have exactly one element¬{json.pretty}"
```
andrewmwells-amazon, line 466: 😆

andrewmwells-amazon (approved):

cedar-lean/DiffTest/Parser.lean
```diff
@@ -345,29 +345,41 @@ def jsonToActionScope (json : Lean.Json) : ParseResult ActionScope := do
...
+def jsonToTemplate (json : Lean.Json) : ParseResult Template := do
   let effect ← getJsonField json "effect" >>= jsonToEffect
-  let principalConstraint ← getJsonField json "principal_constraint" >>= (getJsonField · "constraint") >>= jsonToScope
+  let principalConstraint ← getJsonField json "principal_constraint" >>= (getJsonField · "constraint") >>= jsonToScopeTemplate "?principal"
   let actionConstraint ← getJsonField json "action_constraint" >>= jsonToActionScope
-  let resourceConstraint ← getJsonField json "resource_constraint" >>= (getJsonField · "constraint") >>= jsonToScope
+  let resourceConstraint ← getJsonField json "resource_constraint" >>= (getJsonField · "constraint") >>= jsonToScopeTemplate "?resource"
   let condition ← getJsonField json "non_head_constraints" >>= jsonToExpr
   .ok {
-    id := id
```
cdisselkoen, line 356: do templates not have ids

cdisselkoen (approved):

cedar-lean/DiffTest/Parser.lean
khieta: I love the unicode support for Lean in VSCode! But it's a little over-eager sometimes

cedar-lean/DiffTest/Parser.lean
khieta: Templates do have ids, but I decided that it was more straightforward to store them as map keys (see defn below) rather than in the data structure directly.
```
abbrev Templates := Map TemplateID Template
```
The current parsing code pulls out the ids in the `jsonToPolicies` function.

emina: In case it matters, we decided to keep IDs as part of the `Policy` datatype: https://github.com/cedar-policy/cedar-spec/blob/d17739d370bccc9836f59053a56b558fc9803a70/cedar-lean/Cedar/Spec/Policy.lean#L57

This is largely because it made the `isAuthorized` code and proofs nicer, in particular the proof that authorization decisions are independent of duplicates and evaluation order (https://github.com/cedar-policy/cedar-spec/blob/d17739d370bccc9836f59053a56b558fc9803a70/cedar-lean/Cedar/Thm/Authorization.lean#L90).

Not sure if we plan to do any proofs re templates, but if we do, I suspect that having a list of templates with embedded IDs may be easier to work with than maps.

emina (approved):

cedar-lean/DiffTest/Parser.lean
khieta: Makes sense! I'll leave it as-is for now, and we can reconsider if/when we decide to do proofs with templates.

khieta:
I think the build is now fixed for both `ubuntu-latest` (used by GitHub CI) and `al2` (used by internal CI). Please re-approve if you're ok with my edits to `set_env_vars.sh`. Open to suggestions if you know a better way to bash hack.

shaobo-he-aws (approved):

PR #180, opened 2023-12-14
Lean parser fixes

khieta:
*Issue #, if available:*

*Description of changes:*

This PR switch to using `Set.make` and `Map.make` in the Lean parser. The functions it was previously using (`Set.mk` and `Map.mk`) don't actually enforce the set/map invariants -- they're just the builtin constructors. (My bad.)

This was an issue detected quickly by the `eval-type-directed-lean` target, which checks if an expression evaluates to the expected value. The `Map.mk`/`Map.make` difference is important for record-typed expressions and the `Set.mk`/`Set.make` difference is important for set-typed expressions.

### DIFF
--- a/cedar-lean/DiffTest/Parser.lean
+++ b/cedar-lean/DiffTest/Parser.lean
@@ -240,10 +240,10 @@ partial def exprToValue : Expr → ParseResult Value
   | Expr.lit p => .ok (Value.prim p)
   | Expr.record r => do
     let kvs ← mapMValues r exprToValue
-    .ok (Value.record (Map.mk kvs))
+    .ok (Value.record (Map.make kvs))
   | Expr.set s => do
     let arr ← List.mapM exprToValue s
-    .ok (Value.set (Set.mk arr))
+    .ok (Value.set (Set.make arr))
   | Expr.call xfn args => extExprToValue xfn args
   | expr => .error ("exprToValue: invalid input expression\n" ++ toString (repr expr))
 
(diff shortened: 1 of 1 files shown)


andrewmwells-amazon (approved):

andrewmwells-amazon:
If we did [this ](https://lean-lang.org/functional_programming_in_lean/getting-to-know/structures.html?highlight=constructor#behind-the-scenes:~:text=To%20override%20a%20structure%27s%20constructor%20name%2C%20write%20it%20with%20two%20colons%20at%20the%20beginning.%20For%20instance%2C%20to%20use%20Point.point%20instead%20of%20Point.mk%2C%20write%3A) would it override the `mk` constructor? If so, maybe we could replace it with `mk_with_no_invariants` or something to avoid this in the future?

shaobo-he-aws (approved):

khieta:
That sounds like a good idea. Would prevent people like me from only looking at the function name, but not the documentation 😅 What do you think @emina?

emina:
We've had problems earlier on using structures instead of ADTs.  It was much more painful to deal with them in proofs, with some automation failing for structs when it works just fine for ADTs.  That's why we chose to use inductive datatypes instead.  

You'll also see that a lot of proofs match against the ADT directly, as in `| (Map.mk rty) => ...`.  For example: https://github.com/cedar-policy/cedar-spec/blob/1c4e92d15199a169dd5ee457407bcd69b3e924f3/cedar-lean/Cedar/Thm/Validation/Typechecker/Types.lean#L401

I suspect many things would get broken if we switched the representation to structrures now.  

If someone wants to give it a shot, my main ask is to use a different name than `mk_with_no_invariants`.  It's too long for my preference :) and snake case is against Lean conventions here, so maybe something like `mkRaw`?

Issue #173, opened 2023-12-11
Improve efficiency of Lean wildcard matching


khieta:
### Category

Dafny formalization

### Describe the feature you'd like to request

The Lean formalization of wildcard matching is currently inefficient, and will quickly cause DRT to fail with a "slow unit" once it generates a non-trivial `like` expression. We need to either (1) rewrite the relevant Lean code to be more efficient (e.g., along the lines of https://github.com/cedar-policy/cedar-spec/blob/main/cedar-dafny/def/wildcard.dfy), or (2) limit the number of `*`s that can be generated in a pattern during DRT.

Example: DRT generated the following small-ish policy, which caused the Lean code to hit the timeout of 20 mins.
```
permit(
  principal is a in a::"\0\0\0\0+\0\u{6}",
  action in [Action::"action"],
  resource is a in a::"\0\0\0\0+\0\u{6}"
) when {
  true && ("ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff" like "*****\0\0*")
};
```

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

PR #183, opened 2023-12-15
Disable `like` in ABAC DRT targets

khieta:
*Issue #, if available:*

*Description of changes:*

Disallow `like` expressions from being generated by the ABAC DRT targets, pending a solution for #173. 

This is in the "shared" code, so it will affect both the Dafny and Lean targets. Obviously we will need to revert this PR in the future, but it will be convenient for now to ignore this (known) failure so it doesn't mask other (unknown) failures. Note that `like` expressions can still be generated in PBT targets.


cdisselkoen (approved):
Let's make another issue to remind us to re-enable it

andrewmwells-amazon (approved):

aaronjeline (approved):

shaobo-he-aws (approved):

khieta:
Issue is here: #185

PR #186, opened 2023-12-28
Cleanup in integration tests

khieta:
*Issue #, if available:*

*Description of changes:*

Followup to [cedar#550](https://github.com/cedar-policy/cedar/pull/550). Rather than explicitly listing which integration tests to run (which is difficult to maintain), we can now just just run every test in the `tests` folder.

### DIFF
--- a/cedar-drt/Cargo.toml
+++ b/cedar-drt/Cargo.toml
@@ -18,6 +18,7 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 lazy_static = "1.4"
 smol_str = { version = "0.2", features = ["serde"] }
+walkdir = "2.4"
 
 [dependencies.uuid]
 version = "1.3.1"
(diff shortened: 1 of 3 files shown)


cedar-drt/Cargo.toml
```diff
@@ -18,6 +18,7 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 lazy_static = "1.4"
 smol_str = { version = "0.2", features = ["serde"] }
+walkdir = "2.4"
```
john-h-kastner-aws, line 21: This should work
```suggestion

[dev-dependencies]
walkdir = "2.4"
```

john-h-kastner-aws (approved):

cedar-drt/Cargo.toml
khieta: Nice, thanks!

shaobo-he-aws (approved):

PR #190, opened 2024-01-07
Return timing info in Lean response

khieta:
*Issue #, if available:*

This PR adds timing info to the Lean responses for authorization, validation, and evaluation. Times are obtained using [`IO.monoNanosNow`](https://leanprover-community.github.io/mathlib4_docs/Init/System/IO.html#IO.monoNanosNow). The returned times are printed to the DRT log using `info!`, which matches how we report timing info for the Rust & Dafny.

*Description of changes:*

### DIFF
--- a/cedar-drt/src/lean_impl.rs
+++ b/cedar-drt/src/lean_impl.rs
@@ -41,6 +41,7 @@ use lean_sys::{
 use serde::Deserialize;
 use std::ffi::CStr;
 use std::str::FromStr;
+use log::info;
 
 #[link(name = "Cedar", kind = "static")]
 #[link(name = "Std", kind = "static")]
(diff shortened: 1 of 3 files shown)


cedar-drt/src/lean_impl.rs
```diff
@@ -52,40 +53,40 @@ extern "C" {
     fn initialize_DiffTest_Main(builtin: i8, ob: *mut lean_object) -> *mut lean_object;
 }
 
-#[derive(Debug, Deserialize)]
-struct ListDef<String> {
-    l: Vec<String>,
-}
+pub const LEAN_AUTH_MSG: &str = "Lean authorization time (ns) : ";
```
andrewmwells-amazon, line 56: [nit] personally I'd prefer milliseconds.

cedar-lean/Cli/Main.lean
```diff
@@ -30,7 +30,7 @@ def readFile (filename : String) : IO String :=
 def printUsage (err : String) : IO Unit :=
   IO.println s!"{err}\nUsage: Cli <command> <file>"
 
-def main (args : List String) : IO Unit :=
+unsafe def main (args : List String) : IO Unit :=
```
andrewmwells-amazon, line 33: Just out of curiousity, why is this needed now?

andrewmwells-amazon (approved):

cedar-drt/src/lean_impl.rs
khieta: I don't feel strongly either way, but I like consistency, and the rest of the code uses ns. Do you think it's worth changing the Rust & Dafny timings to use ms too?

cedar-lean/Cli/Main.lean
khieta: It's the result of calling `unsafeBaseIO` in `Cli/Main.lean` in order to pull out the timing info from an IO monad. A little hacky -- open to better suggestions 🙂

emina (approved):

PR #192, opened 2024-01-07
Add scripts to generate corpus

### LINKED ISSUE #191: Reuse `cedar-policy` definitions for integration tests
### Category

DRT infrastructure

In `cedar-spec`, we define [`IntegrationTestCase`](https://github.com/cedar-policy/cedar-spec/blob/b172454af9101e0325fb1e35c223243234922ea3/cedar-drt/fuzz/src/dump.rs#L152) and [`IntegrationRequest`](https://github.com/cedar-policy/cedar-spec/blob/b172454af9101e0325fb1e35c223243234922ea3/cedar-drt/fuzz/src/dump.rs#L168), which are similar to (but not exactly the same as) [`JsonTest`](https://github.com/cedar-policy/cedar/blob/02646fe717f2fa4ef31e4aa9c487788d17396757/cedar-policy/src/integration_testing.rs#L50) and [`JsonRequest`](https://github.com/cedar-policy/cedar/blob/02646fe717f2fa4ef31e4aa9c487788d17396757/cedar-policy/src/integration_testing.rs#L70) in `cedar-policy`. It would be nice to reuse the `cedar-policy` definitions to avoid the code duplication here.

_No response_

### Additional context

_No response_

khieta:
*Issue #, if available:*

*Description of changes:*

Added scripts to generate corpus files, towards solving [cedar#552](https://github.com/cedar-policy/cedar/issues/552). Also included some minor refactoring + added support for logging corpus tests from the `abac-type-directed` targets and resolves #191.

### DIFF
--- a/cedar-drt/create_corpus.sh
+++ b/cedar-drt/create_corpus.sh
@@ -1,11 +1,14 @@
 #!/bin/bash
 
-FUZZ_TARGET=abac-type-directed
-TIMEOUT="15m" # 15m = 900s
+FUZZ_TARGET="${FUZZ_TARGET:-abac}"
+TIMEOUT="${TIMEOUT:-15m}" # 15m = 900s
+JOBS="${JOBS:-4}"
+DUMP_DIR="${DUMP_DIR:-corpus_tests}"
 
 # Assumes cedar-drt is already correctly built and configured
-rm -rf corpus_tests
+rm -rf $DUMP_DIR
 source set_env_vars.sh
 export FUZZ_TARGET
-timeout $TIMEOUT cargo fuzz run -s none $FUZZ_TARGET -j4 -- -rss_limit_mb=8196
+export DUMP_DIR
+timeout $TIMEOUT cargo fuzz run -s none $FUZZ_TARGET -j $JOBS -- -rss_limit_mb=8196
 ./synthesize_tests.sh
--- a/cedar-drt/create_corpus.sh
+++ b/cedar-drt/create_corpus.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+FUZZ_TARGET=abac-type-directed
+TIMEOUT="15m" # 15m = 900s
+
+# Assumes cedar-drt is already correctly built and configured
+rm -rf corpus_tests
+source set_env_vars.sh
+export FUZZ_TARGET
+timeout $TIMEOUT cargo fuzz run -s none $FUZZ_TARGET -j4 -- -rss_limit_mb=8196
+./synthesize_tests.sh
--- a/cedar-drt/fuzz/fuzz_targets/abac_shared.rs
+++ b/cedar-drt/fuzz/fuzz_targets/abac_shared.rs
@@ -17,6 +17,7 @@
 use cedar_drt::*;
 use cedar_drt_inner::*;
 use cedar_policy_core::ast;
+use cedar_policy_core::authorizer::Authorizer;
 use cedar_policy_core::entities::Entities;
 use cedar_policy_generators::{
     abac::{ABACPolicy, ABACRequest},
@@ -111,7 +112,8 @@ pub fn fuzz(input: FuzzTargetInput, def_impl: &impl CedarTestImplementation) {
     initialize_log();
     if let Ok(entities) = Entities::try_from(input.hierarchy) {
         let mut policyset = ast::PolicySet::new();
-        policyset.add_static(input.policy.into()).unwrap();
+        let policy: ast::StaticPolicy = input.policy.into();
+        policyset.add_static(policy.clone()).unwrap();
         debug!("Policies: {policyset}");
         debug!("Entities: {entities}");
         let requests = input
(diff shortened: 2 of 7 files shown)


andrewmwells-amazon (approved):

cedar-drt/create_corpus.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+FUZZ_TARGET=abac-type-directed
+TIMEOUT="15m" # 15m = 900s
```
cdisselkoen, line 4: small suggestion to default to 1hr instead. we've had real bugs only discovered after multiple hours of fuzzing. most interesting bugs discovered by fuzzing were not discovered in the first 15 min IIRC

cedar-drt/fuzz/src/dump.rs
```diff
@@ -135,6 +131,18 @@ pub fn dump<'a>(
     Ok(())
 }
 
+/// Check whether a policyset passes validation
+fn passes_validation(schema: &SchemaFragment, policies: &PolicySet) -> bool {
+    if let Ok(schema) = ValidatorSchema::try_from(schema.clone()) {
```
cdisselkoen, line 136: nit: since we unconditionally clone `schema`, it might make sense to take an owned `SchemaFragment` in this function, allowing caller to give us ownership if they have it.  (This won't make a difference in this particular case, so I guess it's a style judgment call.)

cedar-drt/create_corpus.sh
```diff
@@ -0,0 +1,11 @@
...
+
+# Assumes cedar-drt is already correctly built and configured
+rm -rf corpus_tests
+source set_env_vars.sh
+export FUZZ_TARGET
+timeout $TIMEOUT cargo fuzz run -s none $FUZZ_TARGET -j4 -- -rss_limit_mb=8196
```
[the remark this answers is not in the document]
john-h-kastner-aws: something like `jobs="${jobs:-4}"` to make it easily configurable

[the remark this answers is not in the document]
cdisselkoen: is that like make's `?=` to set only if not already set?  if so we could do this for TIMEOUT and FUZZ_TARGET as well, then you could run this script like `TIMEOUT=3h ./create_corpus.sh`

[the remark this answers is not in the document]
john-h-kastner-aws: Yeah, that's the idea

cedar-drt/create_corpus.sh
andrewmwells-amazon: I think we decided on 15mins because we wanted to make regenerating the corpus easier. People were running scripts to modify the corpus rather than regenerate, which meant the code to generate the corpus could get out of sync with the actual corpus.

IMO it makes sense for to leave this at 15mins for now and to continue to rely on the nightly runs to find deeper bugs. I'll try to get some mutation testing experiments done "soon," and there I'll check what percentage of mutations are caught by a 15min vs 1hr corpus. If the difference is significant, then I'd obviously support increasing this number.

khieta:
Apologies for the delay in updating this PR! My latest commits resolve #191 and make the suggested changes to the scripts. I decided to leave the default timeout to 15 minutes for now -- I expect we'll change it in the future after we have more evidence about what timeout leads to the "best" corpus tests.

This PR will need to be merged _after_ [cedar#589](https://github.com/cedar-policy/cedar/pull/589).

aaronjeline (approved):

andrewmwells-amazon (approved):

andrewmwells-amazon (approved):

cedar-drt/fuzz/fuzz_targets/abac_shared.rs
```diff
@@ -106,50 +106,45 @@ impl<'a> Arbitrary<'a> for FuzzTargetInput {
...
+            // Recreate the policy set and compute responses here to account for this.
+            let mut policyset = ast::PolicySet::new();
+            let policy = policy.new_id(ast::PolicyID::from_string("policy0"));
+            policyset.add_static(policy).unwrap();
+            let mut responses = Vec::with_capacity(requests.len());
+            for request in requests.iter() {
+                let authorizer = Authorizer::new();
+                let response = authorizer.is_authorized(request.clone(), &policyset, &entities);
+                responses.push(response);
+            }
```
cdisselkoen, line 142:
```suggestion
            let mut responses = requests.iter().map(|request| {
                let authorizer = Authorizer::new();
                authorizer.is_authorized(request.clone(), &policyset, &entities)
            }).collect::<Vec<_>>();
```

cedar-drt/fuzz/fuzz_targets/abac_type_directed_shared.rs
```diff
@@ -135,4 +144,28 @@ pub fn fuzz(input: FuzzTargetInput, def_impl: &impl CedarTestImplementation) {
...
+        // Recreate the policy set and compute responses here to account for this.
+        let mut policyset = ast::PolicySet::new();
+        let policy = policy.new_id(ast::PolicyID::from_string("policy0"));
+        policyset.add_static(policy).unwrap();
+        let mut responses = Vec::with_capacity(requests.len());
+        for request in requests.iter() {
+            let authorizer = Authorizer::new();
+            let response = authorizer.is_authorized(request.clone(), &policyset, &input.entities);
+            responses.push(response);
+        }
```
cdisselkoen, line 159: same here

cdisselkoen: In fact, would it be possible to share some of this code, which is currently duplicated.  (doesn't have to be in this PR)

cedar-drt/fuzz/src/dump.rs
```diff
@@ -87,45 +94,35 @@ pub fn dump<'a>(
...
+                    reason: cedar_policy::Response::from(a.clone())
+                        .diagnostics()
+                        .reason()
+                        .cloned()
+                        .collect(),
+                    errors: cedar_policy::Response::from(a.clone())
+                        .diagnostics()
+                        .errors()
+                        .map(AuthorizationError::id)
+                        .cloned()
```
cdisselkoen, line 125: do we need to do `cedar_policy::Response::from(a.clone())` twice?

cedar-drt/fuzz/src/dump.rs
```diff
@@ -135,53 +132,62 @@ pub fn dump<'a>(
...
-/// Dump the entity uid to a string if it is specified, otherwise return `None`,
-/// which will appear as `null` in the JSON dump.
-fn dump_request_var(var: &EntityUIDEntry) -> Option<String> {
+/// Check whether a policy set can be successfully parsed
+fn well_formed(policies: &PolicySet) -> bool {
+    policies
+        .static_policies()
+        .map(ToString::to_string)
+        .all(|p| Policy::from_str(&p).is_ok())
+}
```
cdisselkoen, line 141: is it sufficent to check the `.static_policies()`? should we also check templates?  is there an easy way to iterate over both static policies and templates?

cdisselkoen (approved):

cedar-drt/fuzz/src/dump.rs
khieta: The current code only writes out the static policies because the relevant `abac*` targets only generate static policies. The general function is `.policies()`.

cedar-drt/fuzz/fuzz_targets/abac_type_directed_shared.rs
khieta: Reformatted in the latest commit. I didn't deal with the code reuse, but you're right that would be cleaner

cedar-drt/fuzz/src/dump.rs
cdisselkoen: Sounds good.

cdisselkoen (approved):

PR #198, opened 2024-01-10
simplify some proofs in Slicing.lean

andrewmwells-amazon:
*Issue #, if available:*

*Description of changes:*
simplify some proofs in Slicing.lean

### DIFF
--- a/cedar-lean/Cedar/Thm/Slicing.lean
+++ b/cedar-lean/Cedar/Thm/Slicing.lean
@@ -64,8 +64,7 @@ theorem isAuthorized_eq_for_sound_policy_slice (req : Request) (entities : Entit
     satisfiedPolicies_eq_for_sound_policy_slice .forbid req entities slice policies h₀.left h₀.right
   have hp : satisfiedPolicies .permit slice req entities = satisfiedPolicies .permit policies req entities :=
     satisfiedPolicies_eq_for_sound_policy_slice .permit req entities slice policies h₀.left h₀.right
-  unfold isAuthorized
-  simp [hf, hp]
+  simp [isAuthorized, hf, hp]
 
 /--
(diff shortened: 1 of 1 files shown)


khieta (approved):

cedar-lean/Cedar/Thm/Slicing.lean
```diff
@@ -166,21 +163,20 @@ theorem scope_bound_is_sound (policy : Policy) :
...
 
 /--
 Scope-based bound analysis is sound.
 -/
 theorem scope_analysis_is_sound :
   IsSoundBoundAnalysis scopeAnalysis
 := by
-  unfold IsSoundBoundAnalysis
-  apply scope_bound_is_sound
+  simp [IsSoundBoundAnalysis, scope_bound_is_sound]
```
cdisselkoen, line 179: is the new version easier to read?  It seems to me the old version is more explicit and perhaps easier to understand what it's doing / why it works

andrewmwells-amazon: Agree this case is arguable. My preferred proof method is spamming the `simp` tactic and under that metric it's better, but I'm fine with removing this change.

khieta: +1 for spamming `simp`

andrewmwells-amazon: I guess let's see what @emina says.

emina: I also find the more verbose proof perhaps slightly easier to understand, step through, and fix.  But short proofs look nicer :). How about we adopt the shorter proofs, and if we find they are harder to maintain, we go back to the more verbose ones?

emina (approved):

PR #199, opened 2024-01-12
chore: use omega for linear arithmetic

semorrison:
*Description of changes:*

This PR updates the dependency on the Lean Standard Library to the latest version, and then uses the new tactic `omega` to handle all the proofs about inequalities in `Nat` and `Int`.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Please feel free to accept or reject the various changes separately, as they should all be independent.

PR #200, opened 2024-01-18
replace trivial `rcases` with `have`

cdisselkoen:
This PR is pure refactoring, no functional change.

Replaces (almost) all of our use of `rcases` with a trivial pattern (no destructuring, or just destructuring a conjunction) with an equivalent `have`.

Some side effects:
* Affects case names subsequent to the replacement.  This is a beneficial side effect; the new names are simpler but equally informative.
* Some behavior is different when we use `rcases`/`have` to shadow a previous binding, as in `have ⟨_, h₁⟩ := h₁`.  I believe the shadowed (but still present) binding is put in a different order by `rcases` vs `have`, which may affect any subsequent renaming operations.  I've made the necessary adjustments in this PR; they are minor.  In many of these cases I just removed the shadowing for clarity.

### DIFF
--- a/cedar-lean/Cedar/Data/LT.lean
+++ b/cedar-lean/Cedar/Data/LT.lean
@@ -32,7 +32,7 @@ theorem StrictLT.irreflexive [LT α] [StrictLT α] (x : α) :
   ¬ x < x
 := by
   by_contra h₁
-  rcases (StrictLT.asymmetric x x h₁) with h₂
+  have h₂ := StrictLT.asymmetric x x h₁
   contradiction
 
 theorem StrictLT.if_not_lt_gt_then_eq [LT α] [StrictLT α] (x y : α) :
@@ -40,7 +40,7 @@ theorem StrictLT.if_not_lt_gt_then_eq [LT α] [StrictLT α] (x y : α) :
 := by
   intro h₁ h₂
   by_contra h₃
-  rcases (StrictLT.connected x y h₃) with h₄
+  have h₄ := StrictLT.connected x y h₃
   simp [h₁, h₂] at h₄
 
 theorem StrictLT.not_eq [LT α] [StrictLT α] (x y : α) :
@@ -49,7 +49,7 @@ theorem StrictLT.not_eq [LT α] [StrictLT α] (x y : α) :
   intro h₁
   by_contra h₂
   subst h₂
-  rcases (StrictLT.irreflexive x) with h₃
+  have h₃ := StrictLT.irreflexive x
   contradiction
 
 abbrev DecidableLT (α) [LT α] := DecidableRel (α := α) (· < ·)
@@ -82,14 +82,14 @@ theorem List.lt_irrefl [LT α] [StrictLT α] (xs : List α) :
   case nil => by_contra; contradiction
   case cons _ _ hd tl ih =>
     by_contra h₁
-    rcases (StrictLT.irreflexive hd) with h₂
+    have h₂ := StrictLT.irreflexive hd
     cases tl
     case nil =>
-      rcases (List.lt_cons_cases h₁) with h₃
+      have h₃ := List.lt_cons_cases h₁
       simp [h₂] at h₃
       contradiction
     case cons _ _ hd' tl' =>
-      rcases (List.lt_cons_cases h₁) with h₃
+      have h₃ := List.lt_cons_cases h₁
       simp [h₂] at h₃
       contradiction
 
@@ -105,19 +105,19 @@ theorem List.lt_trans [LT α] [StrictLT α] {xs ys zs : List α} :
       apply List.lt.head
       apply StrictLT.transitive _ _ _ h₃ h₄
     case tail _ _ zhd ztl h₄ h₅ h₆ =>
-      rcases (StrictLT.if_not_lt_gt_then_eq yhd zhd h₄ h₅) with h₇
+      have h₇ := StrictLT.if_not_lt_gt_then_eq yhd zhd h₄ h₅
       subst h₇
       apply List.lt.head
       exact h₃
   case tail _ _ xhd xtl yhd ytl h₃ h₄ h₅ =>
     cases h₂
     case head _ _ zhd ztl h₆ =>
-      rcases (StrictLT.if_not_lt_gt_then_eq xhd yhd h₃ h₄) with h₇
+      have h₇ := StrictLT.if_not_lt_gt_then_eq xhd yhd h₃ h₄
       subst h₇
       apply List.lt.head
       exact h₆
     case tail _ _ zhd ztl h₆ h₇ h₈ =>
-      rcases (StrictLT.if_not_lt_gt_then_eq xhd yhd h₃ h₄) with h₉
+      have h₉ := StrictLT.if_not_lt_gt_then_eq xhd yhd h₃ h₄
       subst h₉
       apply List.lt.tail h₆ h₇
       apply List.lt_trans h₅ h₈
@@ -135,8 +135,8 @@ theorem List.lt_asymm [LT α] [StrictLT α] {xs ys : List α} :
     case nil => contradiction
     case cons _ _ hd' tl' =>
       by_contra h₂
-      rcases (List.lt_trans h₁ h₂) with h₃
-      rcases (List.lt_irrefl (hd :: tl)) with h₄
+      have h₃ := List.lt_trans h₁ h₂
+      have h₄ := List.lt_irrefl (hd :: tl)
       contradiction
 
 theorem List.lt_conn [LT α] [StrictLT α] {xs ys : List α} :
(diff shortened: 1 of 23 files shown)


cedar-lean/Cedar/Thm/Validation/Typechecker/BinaryApp.lean
```diff
@@ -807,16 +815,15 @@ theorem type_of_mem_is_sound {x₁ x₂ : Expr} {c₁ c₂ : Capabilities} {env
...
-  rcases (type_of_mem_inversion h₃) with ⟨hc, ety₁, ety₂, ⟨c₁', h₄⟩ , c₂', h₅⟩
+  have ⟨hc, ety₁, ety₂, ⟨c₁', h₄⟩ , c₂', h₅⟩ := type_of_mem_inversion h₃
   subst hc
   constructor
   case left => exact empty_guarded_capabilities_invariant
   case right =>
-    rcases h₅ with h₅ | h₅ <;>
-    rcases h₅ with ⟨h₅, h₆⟩ <;> subst h₆
-    case inl.intro =>
+    rcases h₅ with ⟨h₅, h₆⟩ | ⟨h₅, h₆⟩ <;> subst h₆
```
khieta, line 823: Not sure what the `|` syntax is here. Is this a case where `rcases` can't be replaced due to recursion?

cedar-lean/Cedar/Data/LT.lean
```diff
@@ -145,36 +145,36 @@ theorem List.lt_conn [LT α] [StrictLT α] {xs ys : List α} :
...
   simp [not_or] at h₂
-  rcases h₂ with ⟨h₂, h₃⟩
+  have ⟨h₂, h₃⟩ := h₂
   cases xs <;> cases ys
-  case intro.nil.nil => contradiction
-  case intro.nil.cons _ _ xhd xtl =>
-    rcases (List.lt.nil xhd xtl) with h₄
+  case nil.nil => contradiction
+  case nil.cons _ _ xhd xtl _ =>
+    have h₄ := List.lt.nil xhd xtl
```
khieta, line 152: What's the rationale for the case name changes here and in the rest of the PR? (Just want to check if there's an established style rule to follow.)

cedar-lean/Cedar/Data/LT.lean
```diff
@@ -145,36 +145,36 @@ theorem List.lt_conn [LT α] [StrictLT α] {xs ys : List α} :
...
     contradiction
-  case intro.cons.nil _ _ yhd ytl =>
-    rcases (List.lt.nil yhd ytl) with h₄
+  case cons.nil _ _ yhd ytl _ =>
```
khieta, line 154: Adding the extra underscore to pad out to the total # of arguments?

khieta (approved):

cedar-lean/Cedar/Thm/Core/LT.lean
```diff
@@ -433,16 +435,18 @@ theorem Values.lt_trans {vs₁ vs₂ vs₃: List Value}
...
+    subst h₂ ; simp [h₁]
   case inr.inl =>
-    rcases h₁ with ⟨h₁, _⟩ ; subst h₁ ; simp [h₂]
+    have ⟨h₁, _⟩ := h₁
+    subst h₁ ; simp [h₂]
   case inr.inr =>
-    rcases h₁ with ⟨hl₁, h₁⟩ ; subst hl₁
-    rcases h₂ with ⟨hl₂, h₂⟩ ; subst hl₂
-    rcases (Values.lt_trans h₁ h₂) with h₃
+    have ⟨hl₁, h₃⟩ := h₁ ; subst hl₁
```
emina, line 447: I think we should use `rcases` in all cases that requires deconstructing the output, like this one.

cedar-lean/Cedar/Data/LT.lean
cdisselkoen: We don't choose the case names.  The adjustments to case names are all in response to errors about case-name-not-found.  As you can see, `rcases` was polluting our case names with uninformative garbage like `intro.`, which we no longer `have` (lol) to put up with.  I also touched in this in the PR description

cedar-lean/Cedar/Data/LT.lean
cdisselkoen: This is because of the shadowing behavior change discussed in the PR description.  I think `have` results in the shadowed hypothesis being in a different position in the list, which affects what gets bound here.  Adding this underscore ensures that `yhd` and `ytl` are bound to the same things that they were previously, before the change.

cedar-lean/Cedar/Thm/Core/LT.lean
cdisselkoen: what's your rationale?

emina: I think `rcases` looks better in this case, and it makes the code style consistent.  

We clearly can't replace `rcases` with `have` everywhere, because it's a more powerful tactic.  So, I'd prefer a simple rule for when to use one versus the other:  use `rcases` when you need to deconstruct, and use `have` otherwise.

Using `rcases` also makes it possible / easier to apply a more advanced deconstruction pattern if that becomes necessary in the future as proofs change.

cdisselkoen: I've thought about this more, and I propose a different simple rule for one versus the other: `rcases` for patterns involving `|`, and `have` in all other cases.

The reason I think this is sensible is that it reserves `rcases` for the times when we're actually splitting goals into separate _cases_, hence the name.  When we destructure an `and` with `⟨_, _⟩`, we're not casing on anything.  Lean splits our goal into 1 subcase called "intro", which is not helpful or elegant.

I'm skeptical that we will often be refactoring in a way that changes a term from a non-`|` pattern to a `|` pattern or vice versa.  And realistically, if we do need to refactor in that way, that will involve more invasive changes to the remainder of the proof (as we now have multiple cases to consider, or no longer have multiple cases to consider, respectively), so it's not a huge burden to change between `rcases` and `have` during that refactor.

khieta: Personally I think `have` looks nicer than `rcases` for `⟨_, _⟩` expressions. The syntax looks sort of like a `let`, which makes it easier for me to remember.

cdisselkoen: Thanks for pointing this out -- I think this is also the reason it feels more intuitive for me.  In most other languages, destructuring with `let` or an equivalent puts the bound variables on the left, not the right.

emina: I think `rcases` looks nicer because it looks like a `match` expression :)

Is it possible to use `have` when we're deconstructing an existential quantifier too?  And is it possible to use `have` for compound deconstruction patterns like `⟨a, ⟨b, c⟩, d⟩` ?  If not, then using `rcases` just for `|` won't work as a rule.

cdisselkoen: Yes, `have` can deconstruct existentials, using `⟨a, h⟩` where `a` is the item that exists, and `h` is the new hypothesis that it satisfies.

Yes, `have` can do compound deconstructions like `⟨a, ⟨b, c⟩, d⟩`, and you can see an example of that in this PR.

cedar-lean/Cedar/Thm/Core/LT.lean
```diff
@@ -458,31 +462,31 @@ theorem ValueAttrs.lt_trans {vs₁ vs₂ vs₃: List (Attr × Value)}
...
-      rcases (Value.lt_trans h₁ h₂) with h₃
+      have ⟨hl₁, h₃⟩ := h₁ ; subst hl₁
+      have ⟨hl₂, h₄⟩ := h₂ ; subst hl₂
+      have h₃ := Value.lt_trans h₃ h₄
       simp [LT.lt] at h₃ ; simp [h₃]
   case inl.inr =>
-    rcases h₂ with ⟨h₂, _⟩
-    rcases h₂ with ⟨hl₂, hr₂⟩ ; subst hl₂ hr₂
+    have ⟨⟨hl₂, hr₂⟩, _⟩ := h₂
+    subst hl₂ hr₂
```
cdisselkoen, line 478: here's an example of compound deconstruction with `have`.  (Replacing `rcases` code that wasn't even doing the compound deconstruction 😂 )

cedar-lean/Cedar/Thm/Core/LT.lean
cdisselkoen: Example [here](https://github.com/cedar-policy/cedar-spec/pull/200/files#r1457850016)

cdisselkoen: > To me, it's more intuitive to deconstruct with `match`es than with `let`s

For sum types, yes, but for product types, no.  I find this to be a weird way to destructure a value `tuple` of product type:
```
match tuple with
| (a, b) => ...
```
which is effectively what `rcases` with `⟨_, _⟩` patterns is doing.

emina: Ok, the rule works. 

But I still find having the matching pattern on the LHS really counterintuitive.  At least for all the languages I'm used to, the pattern-matching happens on the right.

I see why `have` is more appealing for binding a single name (a regular `let`).  But for deconstruction,`rcases` still looks better to me because it reads as `match`.

Curious:  is there any observable effect on proof performance (e.g., faster with `have`)?  

Or we just have to settle this based on the aesthetics?  :smile:

cdisselkoen: CI took 2min 38sec to build the Lean for this PR, vs 2min 41sec to build the Lean in #199 (the previous change).  Could be noise, or could be that `have` is a small improvement vs `rcases`.  (I have to imagine that `have` is if anything faster, because since it is less powerful, it is doing less.  It's also not creating new subgoals like `rcases` is, even for trivial matches.)

emina: Yeah, I was hoping that `have` would be  much faster (because it's doing less), so that would be a good justification to go with `have` even for those who find it less intuitive :slightly_smiling_face:

I don't love it but I'll take the 3 seconds!

emina (approved):

cedar-lean/Cedar/Thm/Core/LT.lean
khieta: Sounds like just a matter of preference then 🤷‍♀️ I usually reach for `let` over `match` because it avoids the extra indentation. (Which is not applicable in this case -- that's just why I like `let`)

emina: Agree, just a matter of preference.  And we have 2 vs 1 votes for pattern-matching on the left, so I'm disagreeing and committing on this one :)

PR #206, opened 2024-01-29
Lean: Specify authz response includes ids of erroring policies

cdisselkoen:
*Description of changes:*

The main change is at `Spec/Response.lean` and `Spec/Authorizer.lean`.  `isAuthorized` now reports the set of ids of policies which encountered an error.

I created `Thm/SlicingDefs.lean` to hold the definition of "sound policy slice" (which was previously inlined everywhere).  Importantly, I adjusted this definition to include that all sliced-out policies evaluate without errors.

This required substantial changes to proofs.  I hesitate to call this proof repair -- really, it imposed an entire new proof obligation for the slicing results (in Thm/Slicing.lean): we have to prove that with scope-based policy slicing, all sliced-out policies evaluate without errors.

When possible, I added general-purpose lemmas to `Data/List.lean`, `Data/Set.lean`, etc rather than just inlining them where needed.

This is my first nontrivial Lean PR, suggestions and style corrections are very welcome.  :) 


cedar-lean/Cedar/Thm/Slicing.lean
```diff
@@ -39,18 +40,6 @@ namespace Cedar.Thm
 
 open Cedar.Spec
 
-/--
```
emina, line 42: Code style:  whenever possible, we prefer to have all the definitions relevant to a main theorem (like slicing here) inlined into the same module.  This makes things easier to read.  Can we put this definition back into the `Slicing` module and get rid of `SlicingDefs`?

Another reason is that Lean code style prefers short module names that don't require camel case.  In general, Lean projects put related files into folders (like how `Thm/Validation/Typechecker` is organized) rather than group them by name.  Usually, the file with definitions that are common to multiple modules will be called `Basic.lean`.  But in this case, the simplest solution is to just have everything in the single `Slicing` module.

cedar-lean/Cedar/Thm/SlicingDefs.lean
```diff
@@ -0,0 +1,20 @@
+import Cedar.Spec
+
+namespace Cedar.Thm
+
+open Cedar.Spec
+
+/--
```
emina, line 7: See above comment:  let's move back into `Slicing`.

emina (commented):
Very nice first Lean PR :))

Left some comments on code style and one suggestion for making the model shorter.  (If it's too hard to use the shorter model because it breaks too many proofs, please ignore, and we can address it in a separate PR.)

The main piece of style feedback is to get rid of the separate `SlicingDefs` file and merging its sole definition back into `Slicing`.  The merged structure fits better with how Lean projects organize code (more in the comments).

cedar-lean/Cedar/Data/Set.lean
```diff
@@ -174,6 +173,10 @@ theorem make_eq_if_eqv [LT α] [DecidableLT α] [StrictLT α] (xs ys : List α)
   intro h; unfold make; simp
   apply List.if_equiv_strictLT_then_canonical _ _ h
 
+/--
```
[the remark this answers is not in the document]
cdisselkoen: Dropped the comment and removed `make_eq_if_eqv`, inlining its proof into `make_make_eqv`.  It's effectively still accessible as `make_make_eqv.mpr`

cedar-lean/Cedar/Spec/Authorizer.lean
```diff
@@ -35,11 +42,28 @@ def satisfiedWithEffect (effect : Effect) (policy : Policy) (req : Request) (ent
 def satisfiedPolicies (effect : Effect) (policies : Policies) (req : Request) (entities : Entities) : Set PolicyID :=
   Set.make (policies.filterMap (satisfiedWithEffect effect · req entities))
 
+def hasError (policy : Policy) (req : Request) (entities : Entities) : Bool :=
```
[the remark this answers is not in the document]
cdisselkoen: We do use `hasError` by itself in several places (and not just `idIfHasError`), similar to how we use `satisfied` in several places (and not just `satisfiedWithEffect`).  Sometimes we just need a `Bool` and not an `Option PolicyID`.

I will rename `idIfHasError` to `errored` though.

cedar-lean/Cedar/Thm/Authorization/Evaluator.lean
```diff
@@ -69,4 +69,100 @@ theorem and_true_implies_right_true {e₁ e₂ : Expr} {request : Request} {enti
             CoeTC.coe, Coe.coe, Except.ok.injEq, Value.prim.injEq, Prim.bool.injEq] at h₁
         case true => rfl
 
+/- some shorthand to make things easier to read and write -/
+/--
+  `producesBool` means the expression evaluates to a bool (and not an error)
+-/
+def producesBool (e : Expr) (request : Request) (entities : Entities) : Bool :=
```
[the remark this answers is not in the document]
cdisselkoen: Nice

cedar-lean/Cedar/Thm/Slicing.lean
cdisselkoen: The reason I had to make a separate file is because putting this definition in `Slicing` makes an import cycle: `Thm/Authorization/Authorizer.lean` needs this definition, but `Slicing` needs `Thm/Authorization/Authorizer.lean`.  I could rename `SlicingDefs` to `Basic` if that would be better?

cdisselkoen: Or I suppose we could put the definition in `Thm/Authorization/Authorizer.lean`?

emina: Yeah, this is a tough one ... we have an unsat set of constraints more or less :)

For the structure to look most Lean-like, we'd have everything in the same file / module, or we'd have a folder called `Slicing` that includes `Slicing/Basic.lean` with the common defs.  Everything in this folder would be imported by `Thm/Slicing.lean`.  This is what we do for `Validation`.  But it seemed like an overkill for `Slicing`.  Also, we ideally wouldn't have any cross references between `Authorization` and `Slicing` folders (i.e., we aim to have a tree).

We can't satisfy all of these constraints.

Maybe one thing to try for now is to stick `Basic.lean` in the `Authorizer` folder and put the slicing definition in there?  Maybe some other useful definition can be consolidated in that file too.  It's not ideal because I agree that "slicing" is the odd thing out here, but it may be the closest we can get to satisfying the constraints.

cdisselkoen: I moved a bunch of the other `def`s from Slicing as well.  Let me know what you think

emina: Ah, looks like they're all slicing-related.  Is that right?  If so, maybe we just call it `Slicing.lean` instead of `Basic.lean` to make that clear?  I think it would be ok in terms of keeping names idiomatic.

emina (approved):

cedar-lean/Cedar/Data/Set.lean
```diff
@@ -193,6 +202,27 @@ theorem make_mem [LT α] [DecidableLT α] [StrictLT α] (x : α) (xs : List α)
   case mp => apply h₁ h₃
   case mpr => apply h₂ h₃
 
+theorem elts_mk_equiv [LT α] [DecidableLT α] [StrictLT α] {xs : List α} :
```
khieta, line 205: Nit: should be `elts_make_equiv`?

cedar-lean/Cedar/Spec/Response.lean
```diff
@@ -32,7 +32,8 @@ inductive Decision where
 
 structure Response where
   decision : Decision
-  policies : Set PolicyID
+  determiningPolicies : Set PolicyID
+  erroredPolicies : Set PolicyID
```
khieta, line 36: Nit: I have a slight preference for `erroringPolicies` or `errorPolicies`.

cedar-lean/Cedar/Thm/Slicing.lean
khieta: I also like the name `Slicing.lean`

khieta (approved):
Looks great! Left some small comments, but approving as-is.

cedar-lean/Cedar/Thm/Slicing.lean
cdisselkoen: Renamed

cedar-lean/Cedar/Data/Set.lean
cdisselkoen: 👍🏻

cedar-lean/Cedar/Spec/Response.lean
cdisselkoen: `errorPolicies` conflicts with the function name in `Authorizer.lean`, but I like `erroringPolicies` for symmetry-ish with `determiningPolicies`

emina (approved):

khieta (approved):

PR #210, opened 2024-02-06
Updates to Lean validator spec

khieta:
*Issue #, if available:*

*Description of changes:*

The Lean formalization of strict validation has been out of sync with the Rust definition since release. This PR aims to resolve this by 
1. Fixing two small issues in the Lean spec: adding a special "Unspecified" entity to the type store that has no attributes or ancestors, and encoding that actions cannot have ~~ancestors~~ attributes.
2. Changing the property tested by the `strict-validation-drt-*` targets. Now instead of testing whether the spec returns a validation error iff the Rust does, it will check that if the Rust validates, then the spec also validates (ignoring `impossiblePolicy` errors). This allows the Lean spec to be more precise than the Rust, which means that it may return the error `impossiblePolicy` where the Rust successfully validates, or may successfully validate when the Rust returns an error (e.g., because short-circuiting determined that some invalid code is unreachable).

The second change is the controversial one: we're weakening the property tested. See my comments in the code for why I think this is reasonable.

It may be possible to further edit the Lean to make it match the Rust, though some work is required. I've found at lease two other issues (listed below) through DRT. There are likely others.
* Lean typechecking of `==` is more precise than Rust (fixed in [this branch](https://github.com/cedar-policy/cedar-spec/tree/khieta/lean-validator-edits)).
* Rust returns a `HierarchyNotRespected` for some `in` expressions that the Lean successfully validates. Appears to be related to unspecified entities.
----

### DIFF
--- a/cedar-drt/fuzz/src/lib.rs
+++ b/cedar-drt/fuzz/src/lib.rs
@@ -203,7 +203,10 @@ pub fn run_val_test(
                 // return "impossiblePolicy" due to greater precision. In this case, the
                 // input policy is well-typed, although it is guaranteed to always evaluate
                 // to false.
-                if definitional_res.validation_errors.contains(&"impossiblePolicy".to_string()) {
+                if definitional_res
+                    .validation_errors
+                    .contains(&"impossiblePolicy".to_string())
+                {
                     return;
                 }
 
--- a/cedar-drt/fuzz/src/lib.rs
+++ b/cedar-drt/fuzz/src/lib.rs
@@ -199,9 +199,10 @@ pub fn run_val_test(
                 }
             }
             Ok(definitional_res) => {
-                // The definitional validator may return "impossiblePolicy" due to greater precision.
-                // If the definitional validator returns "impossiblePolicy" then the input expression is
-                // well-typed, although it is guaranteed to always evaluate to false.
+                // Even if the Rust validator succeeds, the definitional validator may
+                // return "impossiblePolicy" due to greater precision. In this case, the
+                // input policy is well-typed, although it is guaranteed to always evaluate
+                // to false.
                 if definitional_res.validation_errors.contains(&"impossiblePolicy".to_string()) {
                     return;
                 }
--- a/cedar-drt/fuzz/src/lib.rs
+++ b/cedar-drt/fuzz/src/lib.rs
@@ -203,10 +203,7 @@ pub fn run_val_test(
                 // return "impossiblePolicy" due to greater precision. In this case, the
                 // input policy is well-typed, although it is guaranteed to always evaluate
                 // to false.
-                if definitional_res
-                    .validation_errors
-                    .contains(&"impossiblePolicy".to_string())
-                {
+                if definitional_res.validation_errors == vec!["impossiblePolicy".to_string()] {
                     return;
                 }
 
--- a/cedar-drt/fuzz/src/lib.rs
+++ b/cedar-drt/fuzz/src/lib.rs
@@ -157,47 +157,49 @@ pub fn run_val_test(
 
     let definitional_res = custom_impl.validate(&schema, policies, mode);
 
-    match definitional_res {
-        Err(err) => {
-            // TODO(#175): For now, ignore cases where the Lean code returned an error due to
-            // an unknown extension function.
-            if !err.contains("jsonToExtFun: unknown extension function") {
-                panic!(
-                    "Unexpected parse error\nPolicies:\n{}\nSchema:\n{:?}\nError: {err}",
-                    &policies, schema
+    // If `cedar-policy` does not return an error, then the spec should not return an error.
+    // This implies type soundness of the `cedar-policy` validator since type soundness of the
+    // spec is formally proven.
+    //
+    // In particular, we have proven that if the spec validator does not return an error (B),
+    // then there are no authorization-time errors modulo some restrictions (C). So (B) ==> (C).
+    // DRT checks that if the `cedar-policy` validator does not return an error (A), then neither
+    // does the spec validator (B). So (A) ==> (B). By transitivity then, (A) ==> (C).
+
+    if rust_res.validation_passed() {
+        match definitional_res {
+            Err(err) => {
+                // TODO(#175): For now, ignore cases where the Lean code returned an error due to
+                // an unknown extension function.
+                if !err.contains("jsonToExtFun: unknown extension function") {
+                    panic!(
+                        "Unexpected parse error\nPolicies:\n{}\nSchema:\n{:?}\nError: {err}",
+                        &policies, schema
+                    );
+                }
+            }
+            Ok(definitional_res) => {
+                // The definitional validator may return "impossiblePolicy" due to greater precision.
+                // If the definitional validator returns "impossiblePolicy" then the input expression is
+                // well-typed, although it is guaranteed to always evaluate to false.
+                if definitional_res.validation_errors.contains(&"impossiblePolicy".to_string()) {
+                    return;
+                }
+
+                // But the definitional validator should not return any other error.
+                assert!(
+                    definitional_res.validation_passed(),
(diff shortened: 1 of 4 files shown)


cdisselkoen:
Before I review in detail, one comment about the description: You say that we're weakening the strict-validation-drt property from 

> spec returns validation error iff Rust does

to 

> if Rust validates then spec does, ignoring ImpossiblePolicy

Why not a middle ground,

> Rust validates iff spec does, ignoring ImpossiblePolicy

cdisselkoen:
In particular, the middle property would be satisfied by a spec validator that always passes

emina:
Some quick clarifying questions:

> 
> 1. Fixing two small issues in the Lean spec: adding a special "Unspecified" entity to the type store that has no attributes or ancestors, and encoding that actions cannot have ancestors.

The code is not adding the Unspecified entity type to the spec but rather to the diff testing harness (`DiffTest/Parser`).  I think this is right (we should indeed not add it to the spec), and want to confirm that we'll keep it this way :)

The code is preventing actions from having attributes.  They can still have ancestors.  I'm guessing this is a typo in the description.

> 2. Changing the property tested by the `strict-validation-drt-*` targets. Now instead of testing whether the spec returns a validation error iff the Rust does, it will check that if the Rust validates, then the spec also validates (ignoring `impossiblePolicy` errors). This allows the Lean spec to be more precise than the Rust, which means that it may return the error `impossiblePolicy` where the Rust successfully validates, or may successfully validate when the Rust returns an error (e.g., because short-circuiting determined that some invalid code is unreachable).
> 
> The second change is the controversial one: we're weakening the property tested. See my comments in the code for why I think this is reasonable.

I think this is quite reasonable.  It gives us the freedom to eventually make the Rust validator as precise as the Lean validator, which seems more desirable than making Lean _less_ precise to match Rust.  If we instead make the Lean validator less precise to force equivalence with Rust today, then we'd have to change _both_ Rust and Lean to recover that precision in the future.

emina (approved):

khieta:
> Why not a middle ground,
> Rust validates iff spec does, ignoring ImpossiblePolicy

Unfortunately precision differences affect more than just `ImpossiblePolicy` errors. Consider an expression like `if ?? then 1 else (2 + true)` If the Rust types `??` as `Bool` then it will flag the type error in the `else` branch. If the (more precise) Lean spec types `??` as `True`, then it will not flag the type error since it can guarantee that `then` branch will be taken.

So in general, the additional precision in Lean will result in fewer non-`ImpossiblePolicy` errors being reported.

khieta:
> The code is not adding the Unspecified entity type to the spec but rather to the diff testing harness (DiffTest/Parser). I think this is right (we should indeed not add it to the spec), and want to confirm that we'll keep it this way :)

Yes, I should have been more clear. This PR adds a special unspecified entity when parsing in the schema to match how the parser currently handles an unspecified principal/resource in requests. It does not add unspecified entities to the core formalization.

cedar-drt/fuzz/src/lib.rs
```diff
@@ -177,47 +177,53 @@ pub fn run_val_test(
...
+                }
             }
+            Ok(definitional_res) => {
+                // Even if the Rust validator succeeds, the definitional validator may
+                // return "impossiblePolicy" due to greater precision. In this case, the
+                // input policy is well-typed, although it is guaranteed to always evaluate
+                // to false.
+                if definitional_res
+                    .validation_errors
+                    .contains(&"impossiblePolicy".to_string())
```
cdisselkoen, line 208: what happens here if the definitional validator returns multiple errors, one of which is impossiblePolicy?  Or does the definitional validator never return multiple errors?

cedar-lean/Cedar/Validation/Typechecker.lean
```diff
@@ -185,13 +185,19 @@ def hasAttrInRecord (rty : RecordType) (x : Expr) (a : Attr) (c : Capabilities)
     else ok (.bool .anyBool) (Capabilities.singleton x a)
   | .none     => ok (.bool .ff)
 
+def actionType? (ety : EntityType) (acts: ActionSchema) : Bool :=
+  acts.keys.any (EntityUID.ty · == ety)
```
cdisselkoen, line 189: this isn't the definition of `actionType?` I was intuitively expecting, which would be something like "see if the last component of the entity type is the string `Action`".  Is there a reason to prefer one definition or the other?

cedar-lean/Cedar/Validation/Typechecker.lean
```diff
@@ -185,13 +185,19 @@ def hasAttrInRecord (rty : RecordType) (x : Expr) (a : Attr) (c : Capabilities)
...
+
 def typeOfHasAttr (ty : CedarType) (x : Expr) (a : Attr) (c : Capabilities) (env : Environment) : ResultType :=
```
cdisselkoen, line 191: curious why `typeOfHasAttr` needs adjusting, but no corresponding adjustment for `GetAttr`?

cdisselkoen (approved):

cedar-drt/fuzz/src/lib.rs
khieta: The definitional validator currently always returns a single error. But I'll switch this code to check that the list has length 1 in case we change that in the future.

cedar-lean/Cedar/Validation/Typechecker.lean
khieta: We made the same decision for [`actionUID?`](https://github.com/cedar-policy/cedar-spec/blob/6d98a57f74bdc9e694997fabdfb402902b84775e/cedar-lean/Cedar/Validation/Typechecker.lean#L135). The rationale was that an action should really be defined as something that is in the action schema, rather than an entity with a type name "Action". This is a different approach from the Rust, but it makes the formalization cleaner. Will defer further argument to @emina.

cdisselkoen: I'm good with that, that is a reasonable explanation.  And difftesting ensures the validators behave equivalently anyway

cedar-lean/Cedar/Validation/Typechecker.lean
khieta: Great question! The reason is that the Rust & Lean will both return an error if you try to access a nonexistent attribute. Currently, the Lean spec will return a `unknownEntity` error, while the Rust will return a different error (probably related to an unsafe attribute access). Since this is an error case for the Rust, no further information is checked about the Lean response.

cdisselkoen: I think my question goes the other way.  If I have this straight

- the type of `Action::"view" has foo` clearly should be `False`.  This patch was required to get that behavior, because the existing code somehow wasn't already giving `False`.
- the type of `Action::"view".foo` clearly should be <error>.  No patch was required to get that behavior, because the existing code (correctly) was already giving an error.

So I guess the question is not "why didn't we need to adjust `GetAttr`" but "why does `HasAttr` need a special case for actions" given that the validator should already have a declaration for the type `Action` that involves no attributes

emina (approved):

cedar-lean/Cedar/Validation/Typechecker.lean
cdisselkoen: Got it

mwhicks1:
> This PR adds a special unspecified entity when parsing in the schema to match how the parser currently handles an unspecified principal/resource in requests. It does not add unspecified entities to the core formalization.

I don't understand why it should not be part of the formalization. Rust Cedar's semantics includes the notion of an "unspecified" entity. I don't think it's going anywhere. Not having it in the Lean spec implies it's not part of the "essence" of Cedar, e.g., a different implementation would not have to support it. Maybe that's fine, but then we miss an opportunity to vet the design of the Rust code via proof in Lean, to make sure that it's sensible, e.g., that validation is still sound in the presence of Unspecified entities in requests.

PR #214, opened 2024-02-09
Generalize UnitTest to run in different monads.

emina:
*Description of changes:*

This PR generalizes the unit test runner to support executing tests in different monads (not just IO).

### DIFF
--- a/cedar-lean/UnitTest/IPAddr.lean
+++ b/cedar-lean/UnitTest/IPAddr.lean
@@ -166,6 +166,6 @@ def tests := [
   testsForIpNetEquality]
 
 -- Uncomment for interactive debugging
-#eval TestSuite.runAll tests
+-- #eval TestSuite.runAll tests
 
 end UnitTest.IPAddr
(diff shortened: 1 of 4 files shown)


cedar-lean/UnitTest/IPAddr.lean
```diff
@@ -164,6 +166,6 @@ def tests := [
   testsForIpNetEquality]
 
 -- Uncomment for interactive debugging
--- #eval TestSuite.runAll tests
+#eval TestSuite.runAll tests
```
cdisselkoen, line 169: did we intend to check-in this uncomment? if so, we can adjust the comment that says "Uncomment"

cdisselkoen (approved):

cedar-lean/UnitTest/IPAddr.lean
emina: Good catch; that was not intentional!

Fixing now.

khieta (approved):
Lgtm

Issue #204, opened 2024-01-24
Round trip testing for JSON policy format


khieta:
### Category

DRT target(s)

### Describe the feature you'd like to request

We currently don't have any DRT targets that test the JSON policy format. Such a target may have found the behavior described in [cedar#596](https://github.com/cedar-policy/cedar/issues/596), which shows that it is not the case that all policies that can be parsed by `Policy::from_str` can be converted to json using `Policy::to_json`.

Proposal for a new target: generate ASTs and check that, if they can be re-parsed (which is not guaranteed because auto-generated ASTs may not satisfy properties expected by CSTs), then they can be converted to JSON.

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

PR #215, opened 2024-02-11
Add basic EST round-trip testing

khieta:
*Issue #, if available:*

#204 

*Description of changes:*

This PR renames the `pp` target to `roundtrip` and adds two additional checks per policy:
1. Check that parsing a displayed policy using `parser::parse_policy_or_template_to_est` and converting to JSON produces no errors.
2. Check that converting a policy to JSON and back preserves the policy structure.

To test, I reverted [cedar#601](https://github.com/cedar-policy/cedar/pull/601) and confirmed that the underlying bug was found quickly by check 1. Check 2 didn't turn up any bugs during an overnight run, but it seems useful(?)

### DIFF
--- a/cedar-drt/fuzz/fuzz_targets/roundtrip.rs
+++ b/cedar-drt/fuzz/fuzz_targets/roundtrip.rs
@@ -38,7 +38,7 @@ struct FuzzTargetInput {
 // settings for this fuzz target
 // copy-pasted from abac.rs
 const SETTINGS: ABACSettings = ABACSettings {
-    match_types: true,
+    match_types: false,
     enable_extensions: true,
     max_depth: 7,
     max_width: 7,
--- a/cedar-drt/fuzz/fuzz_targets/roundtrip.rs
+++ b/cedar-drt/fuzz/fuzz_targets/roundtrip.rs
@@ -72,6 +72,7 @@ fn contains_unspecified_entities(p: &StaticPolicy) -> bool {
     p.condition().subexpressions().any(|e| matches!(e.expr_kind(), ExprKind::Lit(Literal::EntityUID(euid)) if matches!(*euid.entity_type(), EntityType::Unspecified)))
 }
 
+// AST --> text --> CST --> AST
 // Print a policy to a string and parse it back using the standard AST parser.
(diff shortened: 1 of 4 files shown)


cedar-drt/fuzz/fuzz_targets/roundtrip.rs
```diff
@@ -0,0 +1,201 @@
...
+            "Failed to round-trip AST: {:?}\nPretty printed form: {}\nParse error: {:?}\n",
+            p, p, err
+        )
+    })
+}
+
+// Print a policy to a string, parse it back using the EST parser, and convert to JSON.
+// Panics if any step fails. The resulting policy may be different from the input
+// (since the roundtrip is lossy), so this function only checks that the round-trip
+// succeeds without error.
```
cdisselkoen, line 89: The lossy step is EST->AST or CST->AST, neither of which is included in this roundtrip I think?  (I believe this takes AST input and goes AST->text->EST->JSON?)  So I think we actually could check that the resulting policy matches the input

cdisselkoen (approved):
See also the tests and test utilities in `est.rs`, which define three different kinds of roundtrips involving EST, one of which is always lossless.

cedar-drt/fuzz/fuzz_targets/roundtrip.rs
khieta: EST->AST would be required here because `check_policy_equivalence` expects two ASTs.

But it would be possible to define this function for ESTs too. I'll update this PR later to include more variants, inspired by the helper functions in `est.rs`.

khieta: Update: just looked at this again, and I think this step actually _should_ be lossless. I assumed the translation was supposed to be lossy since DRT failed when I compare the outputs. But looking at the failure again, it looks like it's [cedar#618](https://github.com/cedar-policy/cedar/issues/618), which should be resolved by [cedar#622](https://github.com/cedar-policy/cedar/pull/622). I'll add in the code to do the AST comparison, and test again once cedar#622 is finished.

shaobo-he-aws (approved):

PR #219, opened 2024-02-13
Deprecate Dafny formalization

khieta:
*Issue #, if available:*

*Description of changes:*

This PR completes the port to Lean proposed in [rfc#32](https://github.com/cedar-policy/rfcs/blob/main/text/0032-port-formalization-to-lean.md) 🎉 

In particular, it:
* Deletes the Dafny formalization, related CI, and build instructions
* Undoes the code duplication introduced in #170. Example: this PR deletes `abac.rs` (the Dafny target) and merges `abac_shared.rs` and `abac-lean.rs` to become the new `abac.rs`.
* Deletes the DRT targets for permissive validation, which is an experimental feature that is not yet modeled in Lean.

The Dafny formalization will continue to live on in prior `release/*` branches of `cedar-spec`.


cedar-drt/README.md
```diff
@@ -5,21 +5,20 @@ See the README in the toplevel directory `..` for instructions on how to run.
...
 
-The table below lists all available fuzz targets, including which component of the code they test and whether they perform property-based testing of the Rust code (PBT) or differential testing of the Rust code against the Dafny spec (DRT).
+The table below lists all available fuzz targets, including which component of the code they test and whether they perform property-based testing of the Rust code (PBT) or differential testing of the Rust code against the Lean spec (DRT).
 
 | Name | Component(s) tested | Type | Description |
 | ----------- | ----------- | ----------- | ----------- |
-| [`abac-type-directed`](fuzz/fuzz_targets/abac-type-directed.rs) | Evaluator | DRT | Diff test evaluator on ABAC policies using (mostly) well-typed inputs |
-| [`abac`](fuzz/fuzz_targets/abac.rs) | Evaluator | DRT | Diff test evaluator on ABAC policies |
+| [`abac-type-directed`](fuzz/fuzz_targets/abac-type-directed.rs) | Authorizer | DRT | Diff test authorizer on ABAC policies using (mostly) well-typed inputs |
+| [`abac`](fuzz/fuzz_targets/abac.rs) | Authorizer | DRT | Diff test evaluator on ABAC policies |
```
cdisselkoen, line 13:
```suggestion
| [`abac`](fuzz/fuzz_targets/abac.rs) | Authorizer | DRT | Diff test authorizer on ABAC policies |
```

cedar-drt/README.md
```diff
@@ -5,21 +5,20 @@ See the README in the toplevel directory `..` for instructions on how to run.
...
 | [`rbac-authorizer`](fuzz/fuzz_targets/rbac-authorizer.rs) | Authorizer | PBT + DRT | Test for correct authorization responses over a set of simple policies |
 | [`rbac`](fuzz/fuzz_targets/rbac.rs) | Authorizer | DRT | Diff test authorizer on sets of RBAC policies, including template instantiations |
 | [`simple-parser`](fuzz/fuzz_targets/simple-parser.rs) |  Parser | PBT | Test that parsing doesn't crash with random input strings |
-| [`strict-validation-drt-type-directed`](fuzz/fuzz_targets/strict-validation-drt-type-directed.rs) | Validator | DRT | Diff test strict validation using (mostly) well-typed inputs |
-| [`validation-drt-type-directed`](fuzz/fuzz_targets/validation-drt-type-directed.rs) | Validator | DRT | Diff test permissive validation using (mostly) well-typed inputs |
-| [`validation-drt`](fuzz/fuzz_targets/validation-drt.rs) | Validator | DRT | Diff test permissive validation |
+| [`validation-drt-type-directed`](fuzz/fuzz_targets/validation-drt-type-directed.rs) | Validator | DRT | Diff test (strict) validation using (mostly) well-typed inputs |
```
cdisselkoen, line 21: are we dropping the non-type-directed validation target?  IMO this is an important target, because many categories of errors which we want to test the validator(s) on, cannot be generated by the type-directed generators

cdisselkoen (approved):
🎉

emina (approved):

cedar-drt/README.md
shaobo-he-aws: IMO, we don't actually test much of the authorizer using this target because we only generate one policy per run.

cedar-drt/README.md
shaobo-he-aws: Another related question: Is our principle to use PBT when there's no formalization? If it is, we should remove this target.

shaobo-he-aws:
Maybe we should tag the version before completely deleting Dafny stuff. Just make it easier to look the Dafny code up.

cdisselkoen:
Personally I think the 3.0.x branch is a good place to look the Dafny code up

cedar-drt/README.md
khieta: The validation targets we currently have are:
* Type-directed DRT for permissive validation (Dafny only)
* Non-type-directed DRT for permissive validation (Dafny only)
* Type-directed DRT for strict validation (Dafny and Lean)
* Non-type-directed PBT for type soundness

The current PR proposes:
* Type-directed DRT for strict validation (Lean only)
* Non-type-directed PBT for type soundness

@cdisselkoen are you proposing we add a non-type-directed variant for strict validation? This seems fine, but I think the reason we didn't have it in the first place was that the non-directed variant didn't end up finding any bugs in the permissive case. Should we also add a type-directed variant for PBT?

@shaobo-he-aws are you suggesting we remove the PBT target? It sort of feels redundant since we have proved type soundness, but possibly this target could help find cases where the Rust validator is unsound (and thus different from the spec).

khieta:
> Maybe we should tag the version before completely deleting Dafny stuff

I also think the `release/3.0.x` branch is sufficient. There is a small diff in the json parsing code between `release/3.0.x` and the current `main`, but everything else is unchanged.

shaobo-he-aws:
Sounds good. Should we add it to the README as well?

cedar-drt/README.md
khieta: Yeah I was thinking this too as I was making the change, but it seems closer to authorization than the `eval-type-directed` target so I'm inclined to stick with "authorizer".

cedar-drt/README.md
shaobo-he-aws: I meant we should remove any DRT targets related to permissive validation since it's not proved in Lean. We should keep the corresponding PBT targets. And the PBT target for strict validation is not needed, in my opinion.

mwhicks1: If we don't drop it, I think we should change the PBT for validator soundness to be type-directed instead. It doesn't make sense to test validator soundness using (mostly) invalid inputs. The property is "if validate(P) then evaluate(P) does not error". But the non type-directed generator is mostly going to be generating "P" such that "validate(P)" is false, which is not helpful at testing the property -- those inputs will be effectively thrown away. I realize it was this way with the Dafny, and it didn't make sense then either. :) This should be an easy change to make.

But I also agree with Shaobo we should think about why we even have this property, given that we have proved it sound in Lean and have DRT to check that Lean and Rust versions agree.

mwhicks1:
It seems a shame not to be diff-testing the permissive validation in Dafny, which was proved sound, against the permissive validation we have now in Rust. It is a feature that customers can use now. Why not test it? Put another way: What's the hurry to clear away the Dafny?

khieta:
> It seems a shame not to be diff-testing the permissive validation in Dafny

We will continue to test the Rust permissive validation against the Dafny spec for versions <= 3.0. It's true that this testing won't capture new changes to the Rust permissive validator, but it seems worth the tradeoff to not have to maintain the entire Dafny formalization. If we do end up making changes to the permissive validator in an effort to stabilize it, then it's worth modeling in Lean anyways 😉

khieta:
Updated the PR to include both type-directed _and_ non-type-directed targets for validation DRT _and_ PBT.

While making this edit I introduced some additional code duplication in our targets, but we have similar duplication in other places, so I filed an issue to cleanup in the future (#220) rather than fixing in this PR.

shaobo-he-aws (approved):

shaobo-he-aws (approved):

andrewmwells-amazon (approved):

PR #234, opened 2024-02-27
Consolidate and generalize data structure proofs.

emina:
*Description of changes:*

This PR consolidates all lemmas about data structures into `Cedar.Thm.Data`.  The structure of the new folder parallels the structure of `Cedar.Data`.  So, all lemmas we've proved about `Cedar.Data.Set` can be found in `Cedar.Thm.Set.`

The PR also generalizes some of the list lemmas and adds new lemmas that have been useful for detailed proofs about Cedar's semantics.


khieta (approved):

cedar-lean/Cedar/Data/List.lean
```diff
@@ -51,19 +51,6 @@ def canonicalize [LT β] [DecidableLT β] (f : α → β) : List α → List α
   | [] => []
   | hd :: tl => insertCanonical f hd (canonicalize f tl)
 
-def Equiv {α} (a b : List α) : Prop :=
-  a ⊆ b ∧ b ⊆ a
-
-infix:50 " ≡ " => Equiv
```
cdisselkoen, line 57: don't we still need this definition in Cedar.Data?  Or is it really only used in proofs?

cedar-lean/Cedar/Data/Map.lean
```diff
@@ -33,31 +33,28 @@ deriving Repr, DecidableEq, Repr, Inhabited
...
-/- Creates a well-formed map from the given list of pairs. -/
+/-- Creates a well-formed map from the given list of pairs. -/
 def make {α β} [LT α] [DecidableLT α] (kvs : List (α × β)) : Map α β :=
   Map.mk (kvs.canonicalize Prod.fst)
 
 /-- Empty map -/
 def empty {α β} : Map α β := .mk []
 
+/-- Returns an ordered and duplicate free list provided the given map is well-formed. -/
+def toList {α : Type u} {β : Type v} (m : Map α β) : List (Prod α β) := m.kvs
```
cdisselkoen, line 48: Elsewhere we use `List (α × β)` instead of `Prod α β`

cedar-lean/Cedar/Data/Map.lean
```diff
@@ -33,31 +33,28 @@ deriving Repr, DecidableEq, Repr, Inhabited
...
+
+
 /-- Returns the keys of `m` as a set. -/
 def keys {α β} (m : Map α β) : Set α :=
   Set.mk (m.kvs.map Prod.fst) -- well-formed by construction
 
 /-- Returns the binding for `k` in `m`, if any. -/
 def find? {α β} [BEq α] (m : Map α β) (k : α) : Option β :=
-  match m.kvs.find? (fun ⟨k', _⟩ => k' == k) with
+  match m.kvs.find? (λ ⟨k', _⟩ => k' == k) with
```
cdisselkoen, line 57: TIL

cedar-lean/Cedar/Data/Map.lean
```diff
@@ -68,8 +65,8 @@ def contains {α β} [BEq α] (m : Map α β) (k : α) : Bool :=
 /-- Returns the binding for `k` in `m` or `err` if none is found. -/
 def findOrErr {α β ε} [BEq α] (m : Map α β) (k : α) (err: ε) : Except ε β :=
   match m.find? k with
-  | some v => ok v
-  | _      => error err
+  | some v => .ok v
+  | _      => .error err
```
cdisselkoen, line 69: curious why this change?  If we don't like using `ok` and `error` without dots, we shouldn't put `open Except` at the top of most of our files.  I adopted the `open Except` because it seemed like the convention we were using before I started working on the Lean

cedar-lean/Cedar/Data/Set.lean
```diff
@@ -34,17 +37,15 @@ deriving Repr, DecidableEq, Inhabited, Lean.ToJson
...
-| .mk elts => elts
-
-open Except
+abbrev elts {α : Type u} (s : Set α) := s.1
 
 ----- Definitions -----
 
--- Creates an ordered/duplicate free list froma  set provided the underlying set is well formed.
+/-- Creates an ordered/duplicate free list from a set provided the underlying
+  set is well formed. -/
```
cdisselkoen, line 45:
```suggestion
/-- 
  Creates an ordered/duplicate free list from a set provided the underlying
  set is well formed. 
-/
```

cedar-lean/Cedar/Thm/Data/Control.lean
```diff
@@ -0,0 +1,49 @@
...
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-/
+
+import Std
+
+/-!
+This file contains `simp` lemmas for proofs about code that uses the `do`
+notation together with `Except` and `Option`.
+-/
```
cdisselkoen, line 22: Did we have this file the whole time?  this would have been useful to know.  I'll have to make sure I'm always importing `Thm.Data.Control` when I'm working with `do` notation

cdisselkoen (approved):

cedar-lean/Cedar/Data/List.lean
emina: It turns out it's used only in proofs.  I was surprised too and was on the fence about leaving it in `Cedar.Data.List` but decided to stick with putting all proof-only things into `Cedar.Thm.*`.  

If we end up wanting to use it in code too, we can move it back (and probably also need to prove that `Equiv a b` is Decidable).

cedar-lean/Cedar/Data/Map.lean
emina: Yes, I also prefer `α × β` to `Prod` and have been changing it whenever I see it.  Missed this one :)

cedar-lean/Cedar/Thm/Data/Control.lean
emina: Yes, but it was named less clearly (Thm.Core.Std) and easy to miss :-\. I've renamed it to Control to match Std naming for things to related with the `do` notation.  You definitely want to import this for any proofs that deal with `do` code; you can avoid a lot of boilerplate (see eg what the first `simp` does in the proof of Cedar.Thm.Data.List.foldlM_of_assoc)

andrewmwells-amazon (approved):

PR #235, opened 2024-02-28
simplify simps

andrewmwells-amazon:
*Description of changes:*
Simplify usages of `simp` tactic in the middle of proofs. Using `simp` (as opposed to `simp only`), makes proofs more likely to break as adding the "simp" attribute to new theorems can break the proof. This is not the case at the end of a proof / case; we can freely use `simp` there.

### DIFF
--- a/cedar-lean/Cedar/Thm/Authorization/Evaluator.lean
+++ b/cedar-lean/Cedar/Thm/Authorization/Evaluator.lean
@@ -113,8 +113,8 @@ theorem ways_and_can_error {e₁ e₂ : Expr} {request : Request} {entities : En
       case bool b =>
         cases b with
         | true =>
-          simp
-          simp [h_e₁] at h₁
+          simp only [true_and]
+          simp only [h_e₁] at h₁
           cases h_e₂ : (evaluate e₂ request entities) with
           | ok val =>
             cases val <;>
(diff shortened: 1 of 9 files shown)


cedar-lean/Cedar/Thm/Authorization/Evaluator.lean
```diff
@@ -113,8 +113,8 @@ theorem ways_and_can_error {e₁ e₂ : Expr} {request : Request} {entities : En
       case bool b =>
         cases b with
         | true =>
-          simp
-          simp [h_e₁] at h₁
+          simp only [true_and]
+          simp only [h_e₁] at h₁
```
cdisselkoen, line 117: in this (and other) cases, would it be clearer to replace `simp only` with just `rw`?  What are the differences?  I think one difference is that `rw` would make you be explicit about which direction to rewrite (requires `\l` in some cases); we can decide whether we like that

andrewmwells-amazon: `simp` does quite a bit more than `rw` (https://leanprover.github.io/theorem_proving_in_lean4/tactics.html#rewriting). `rw` fails on both of these. Btw, `simp` also accepts the `\l` syntax.

emina: Thanks for making this PR!  I've been going through some other proofs and implementing the "no non-terminal simps rule", and haven't gotten this far yet :)

emina (approved):

cdisselkoen (approved):

PR #236, opened 2024-02-28
Print the correct LHS and RHS in schema roundtrip

aaronjeline:
*Issue #, if available:*
N/A
*Description of changes:*
The `schema-roundtrip` target used to print LHS for both the LHS and RHS output.


cdisselkoen (approved):
Oof

khieta (approved):
Lol

PR #241, opened 2024-03-01
Added more logging info for target roundtrip

shaobo-he-aws:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cedar-drt/fuzz/fuzz_targets/roundtrip.rs
+++ b/cedar-drt/fuzz/fuzz_targets/roundtrip.rs
@@ -25,7 +25,7 @@ use cedar_policy_generators::{
     abac::ABACPolicy, hierarchy::HierarchyGenerator, schema::Schema, settings::ABACSettings,
 };
 use libfuzzer_sys::arbitrary::{self, Arbitrary, Unstructured};
-use log::debug;
+use log::info;
 use serde::Serialize;
 use smol_str::SmolStr;
 use std::collections::HashMap;
@@ -91,12 +91,16 @@ fn round_trip_ast(p: &StaticPolicy) -> StaticPolicy {
 fn round_trip_json(p: StaticPolicy) -> StaticPolicy {
     // convert to json
     let est = est::Policy::from(ast::Policy::from(p));
+    info!("round_trip_json est: {est:?}");
     let json = serde_json::to_value(est).expect("failed to convert EST to JSON");
+    info!("round_trip_json json: {}", json.to_string());
     // read back
     let est: est::Policy = serde_json::from_value(json).expect("failed to parse EST from JSON");
+    info!("round_trip_json roundtripped est: {est:?}");
     let template = est
         .try_into_ast_template(None)
         .expect("failed to convert EST to AST");
+    info!("round_trip_json roundtripped template: {template:?}");
     template
         .try_into()
         .expect("failed to convert `Template` to `StaticPolicy`")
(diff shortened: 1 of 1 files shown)


cedar-drt/fuzz/fuzz_targets/roundtrip.rs
```diff
@@ -91,12 +91,16 @@ fn round_trip_ast(p: &StaticPolicy) -> StaticPolicy {
 fn round_trip_json(p: StaticPolicy) -> StaticPolicy {
     // convert to json
     let est = est::Policy::from(ast::Policy::from(p));
+    info!("round_trip_json est: {est:?}");
     let json = serde_json::to_value(est).expect("failed to convert EST to JSON");
+    info!("round_trip_json json: {}", json.to_string());
```
cdisselkoen, line 96:
```suggestion
    info!("round_trip_json json: {json}");
```

cedar-drt/fuzz/fuzz_targets/roundtrip.rs
```diff
@@ -112,24 +116,29 @@ fn round_trip_est(p: &StaticPolicy) -> StaticPolicy {
...
         )
     });
+    info!("round_trip_est est: {est:?}");
     let json = serde_json::to_value(est).unwrap_or_else(|err| {
         panic!(
             "Failed to convert EST to JSON: {:?}\nParse error: {:?}\n",
             p, err
         )
     });
+    info!("round_trip_est json: {}", json.to_string());
```
cdisselkoen, line 126:
```suggestion
    info!("round_trip_est json: {json}");
```

cdisselkoen (commented):
1) Will all of this logging slow down the fuzz execution any?  Or will the logging only happen during debugging, and not during real fuzzing runs?

2) Do we have a policy about what debug-logging we check in?  Other targets don't have this level of logging.  How do we decide whether to check-in this kind of code to `cedar-spec`, vs just add it locally as needed for debugging particular issues?  (I'm not saying we shouldn't take this PR, just wondering in general how we decide what logging we should and shouldn't include in our fuzz targets)

shaobo-he-aws:
> 1. Will all of this logging slow down the fuzz execution any?  Or will the logging only happen during debugging, and not during real fuzzing runs?

I think it should not if we don't enable logging, which should only be used for stats collection and debugging.

> 2. Do we have a policy about what debug-logging we check in?  Other targets don't have this level of logging.  How do we decide whether to check-in this kind of code to `cedar-spec`, vs just add it locally as needed for debugging particular issues?  (I'm not saying we shouldn't take this PR, just wondering in general how we decide what logging we should and shouldn't include in our fuzz targets)

No, I think it's kind of ad-hoc and more like demand-driven.

john-h-kastner-aws:
> > 1. Will all of this logging slow down the fuzz execution any?  Or will the logging only happen during debugging, and not during real fuzzing runs?
> 
> I think it should not if we don't enable logging, which should only be used for stats collection and debugging.

This should be configurable at run time with `RUST_LOG=debug` so that we can have fuzzing run with `RUST_LOG=error` to limit logging, but then we can re-run the failing unit with `RUST_LOG=debug` so that the logs show the more verbose output.

cdisselkoen (approved):
Personally I think it's fine to check these in, curious on others' thoughts though

khieta (approved):
Also seems fine to me to merge this. Just make sure that we are using the appropriate `RUST_LOG` level when we run DRT.

andrewmwells-amazon (approved):

shaobo-he-aws:
Closed as the code has diverged too much from main.

PR #243, opened 2024-03-05
Fix schema roundtrip equivalence property

shaobo-he-aws:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cedar-drt/fuzz/src/schemas.rs
+++ b/cedar-drt/fuzz/src/schemas.rs
@@ -64,29 +64,10 @@ fn action_type_equivalence(name: &str, lhs: ActionType, rhs: ActionType) -> Resu
     } else {
         match (lhs.applies_to, rhs.applies_to) {
             (None, None) => Ok(()),
-            (Some(lhs), Some(rhs)) if is_either_empty(&lhs) => {
-                if is_either_empty(&rhs) {
-                    Ok(())
-                } else {
-                    Err(format!(
-                        "Mismatched applies to in `{name}`. lhs : `{:?}`,rhs: `{:?}`",
-                        lhs, rhs
-                    ))
-                }
-            }
-            (Some(lhs), Some(rhs)) if is_either_empty(&rhs) => {
-                if is_either_empty(&lhs) {
-                    Ok(())
-                } else {
-                    Err(format!(
-                        "Mismatched applies to in `{name}`. lhs : `{:?}`,rhs: `{:?}`",
-                        lhs, rhs
-                    ))
-                }
-            }
-            // if neither is non-applicable, they must be equal
             (Some(lhs), Some(rhs)) => {
-                if rhs == lhs {
+                // If either of them has at least one empty appliesTo list, the other must have the same attribute.
+                // Otherwise both of them must apply to unspecified entities or non-empty entity lists, which must be equal.
+                if (either_empty(&lhs) && either_empty(&rhs)) || rhs == lhs {
                     Ok(())
                 } else {
                     Err(format!(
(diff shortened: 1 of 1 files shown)


cedar-drt/fuzz/src/schemas.rs
```diff
@@ -64,8 +64,29 @@ fn action_type_equivalence(name: &str, lhs: ActionType, rhs: ActionType) -> Resu
...
+                    Err(format!(
+                        "Mismatched applies to in `{name}`. lhs : `{:?}`,rhs: `{:?}`",
+                        lhs, rhs
+                    ))
+                }
+            }
+            // if neither is non-applicable, they must be equal
             (Some(lhs), Some(rhs)) => {
-                if empty_target(&rhs) || empty_target(&lhs) || rhs == lhs {
+                if rhs == lhs {
```
khieta, line 89: Splitting up these cases made the code difficult to read. I think it would be cleaner to have something like:
```rust
(Some(lhs), Some(rhs)) => {
    if is_either_empty(&lhs) && is_either_empty(&rhs) {
        Ok(())
    } else if rhs == lhs {
        Ok(())
    } else {
    ...
```

cedar-drt/fuzz/src/schemas.rs
```diff
@@ -85,14 +107,11 @@ fn action_type_equivalence(name: &str, lhs: ActionType, rhs: ActionType) -> Resu
...
-    spec.resource_types
-        .as_ref()
-        .map(|v| v.is_empty())
-        .unwrap_or(false)
-        || spec
-            .principal_types
-            .as_ref()
-            .map(|v| v.is_empty())
-            .unwrap_or(false)
+fn is_both_unspecified(spec: &ApplySpec) -> bool {
```
khieta, line 110: Nit: grammatically speaking, I think this should be `are_both_unspecified`. But that looks weird, so I'd suggest changing to just `both_unspecified`.

cedar-drt/fuzz/src/schemas.rs
```diff
@@ -85,14 +107,11 @@ fn action_type_equivalence(name: &str, lhs: ActionType, rhs: ActionType) -> Resu
...
+    spec.resource_types.is_none() && spec.principal_types.is_none()
+}
+
+fn is_either_empty(spec: &ApplySpec) -> bool {
```
khieta, line 114: Nit: Like above, change to `either_empty`.

cedar-drt/fuzz/src/schemas.rs
andrewmwells-amazon: Note the proposed refactor changes behavior. I'd recommend:
```
if lhs == rhs {
  Ok(())
} else if empty_target(&rhs) || empty_target(&lhs) {
   ...
}
```

andrewmwells-amazon (changes requested):

cedar-drt/fuzz/src/schemas.rs
khieta: I think the refactor is the correct behavior here: if one of the LHS or RHS has an empty `appliesTo` list, then both should.

shaobo-he-aws: I ended up adopting Kesha's refactor because it's easier to understand (please refer to the comments).

cedar-drt/fuzz/src/schemas.rs
shaobo-he-aws: Done.

khieta (approved):
Lgtm. Make sure to backport to the `release/3.1.x` branch too.

andrewmwells-amazon (approved):

cedar-drt/fuzz/src/schemas.rs
```diff
@@ -65,7 +65,9 @@ fn action_type_equivalence(name: &str, lhs: ActionType, rhs: ActionType) -> Resu
         match (lhs.applies_to, rhs.applies_to) {
             (None, None) => Ok(()),
             (Some(lhs), Some(rhs)) => {
-                if empty_target(&rhs) || empty_target(&lhs) || rhs == lhs {
+                // If either of them has at least one empty appliesTo list, the other must have the same attribute.
+                // Otherwise both of them must apply to unspecified entities or non-empty entity lists, which must be equal.
+                if (either_empty(&lhs) && either_empty(&rhs)) || rhs == lhs {
```
cdisselkoen, line 70: I don't see how the comment "the other must have the same attribute" is reflected in the code

shaobo-he-aws: I meant the other must have the same characteristics or states.

khieta: My interpretation was: either the two `appliesTo` specifications must be equal, or they both must satisfy `either_empty` (indicating that they're action groups)

PR #256, opened 2024-03-14
Use new integration test repository

khieta:
*Issue #, if available:*

*Description of changes:*

Followup to [cedar#730](https://github.com/cedar-policy/cedar/pull/730).


aaronjeline (approved):

cedar-drt/README.md
```diff
@@ -52,3 +52,20 @@ If you run into weird build issues,
 3. If the steps above don't help, then file [an issue](https://github.com/cedar-policy/cedar-spec/issues).
 
 If everything builds, but the integration tests are failing, then it may be helpful to set `RUST_BACKTRACE=1` and run `cargo test -- --nocapture` to print additional test information.
+
+## Running integration tests
+
+The integration tests are run by default in CI (e.g., as a part of each pull request), but you can also run them locally.
+In order to do this, you need to have the [`cedar-integration-tests`](https://github.com/cedar-policy/cedar-integration-tests) repository cloned in the `cedar` directory (`../cedar`).
```
cdisselkoen, line 59: why do they need to be cloned locally? could we instead use a [Git dependency](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories) in Cargo.toml?

cdisselkoen (approved):

cedar-drt/README.md
khieta: 🤔 Maybe, let me play around with that.

khieta: So the answer is "I don't think so" since the integration tests aren't a crate

aaronjeline (approved):

PR #262, opened 2024-03-21
a few more data structure lemmas

cdisselkoen:
feel free to suggest improvements to the proofs

### DIFF
--- a/cedar-lean/Cedar/Thm/Data/List.lean
+++ b/cedar-lean/Cedar/Thm/Data/List.lean
@@ -118,6 +118,67 @@ theorem filterMap_equiv (f : α → Option β) (xs ys : List α) :
   exact h₁ h₃
   exact h₂ h₃
 
+theorem filterMap_empty_iff_f_returns_none {f : α → Option β} {xs : List α} :
+  xs.filterMap f = [] ↔ ∀ x ∈ xs, f x = none
+:= by
+  constructor
+  case mp =>
+    induction xs
+    case nil => simp
+    case cons x xs h_ind =>
+      intro h₁ a h₂
+      simp only [List.filterMap_cons] at h₁
+      split at h₁ <;> try simp at h₁
+      case h_1 h₃ =>
+        rcases (List.mem_cons.mp h₂) with h₄ | h₄
+        case inl => subst h₄ ; assumption
+        case inr => apply h_ind h₁ a ; assumption
+  case mpr =>
+    intro h₁
+    induction xs
+    case nil => simp
+    case cons x xs h_ind =>
+      simp only [List.filterMap_cons]
+      split
+      case h_1 =>
+        apply h_ind ; clear h_ind
+        intro a h₂
+        apply h₁ a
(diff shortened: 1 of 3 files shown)


cdisselkoen:
Whoops there's still a sorry here, let me fix

cdisselkoen:
Fixed

emina (approved):
Looks good to me!

cedar-lean/Cedar/Thm/Data/Set.lean
```diff
@@ -280,6 +305,62 @@ theorem inter_wf {α} [LT α] [StrictLT α] [DecidableLT α] [DecidableEq α] {s
...
+  apply h₁ h₂
+
+/--
+  Like `List.subset_def`, but lifted to Sets
+-/
+theorem subset_def [DecidableEq α] {s₁ s₂ : Set α} :
+  s₁ ⊆ s₂ ↔ ∀ a, a ∈ s₁ → a ∈ s₂
+:= by
+  unfold HasSubset.Subset instHasSubsetSet subset
+  simp
```
andrewmwells-amazon, line 338: Can you do `simp?` and replace this with the output?

cedar-lean/Cedar/Thm/Data/List.lean
```diff
@@ -118,6 +118,67 @@ theorem filterMap_equiv (f : α → Option β) (xs ys : List α) :
...
+    replace ⟨x, h₁, h₃⟩ := h₁
+    specialize h₂ x h₁
+    simp [h₂, Option.isSome] at h₃
+
+theorem f_implies_g_then_subset {f g : α → Option β} {xs : List α} :
+  (∀ a b, f a = some b → g a = some b) →
+  xs.filterMap f ⊆ xs.filterMap g
+:= by
+  intro h₁
+  simp [List.subset_def]
```
andrewmwells-amazon, line 176: can this be `simp only`?

andrewmwells-amazon (approved):

cedar-lean/Cedar/Thm/Data/Set.lean
cdisselkoen: I started doing a pass for this in the 3 files touched by this PR, and it's a lot of manual work.  It seems like it could be automated.  Or even rethought: In the event that a Lean upgrade adding something to `simp` breaks a proof, would there be some way to then specifically change to `simp only` just in that place, rather than having to proactively do it everywhere (and also clutter proofs visually)?  It would be awesome if, e.g., in Lean 4.6 we could run some command like `simp4.5?` that would tell us "here's what lemmas `simp` would have used, in Lean 4.5"

andrewmwells-amazon: Seems reasonable to keep separate from this PR then.

I don't think Lean wants to be in the business of managing different versions, but what should work would be to flag the proofs that fail, revert to the previous Lean version, use `simp?` on just those proofs and then do the upgrade. Seems like a nice feature.

PR #263, opened 2024-03-26
More permissive LUB for required/optional record attr types

john-h-kastner-aws:
The least upper bound between a required and optional qualified attribute type is defined as an optional qualified type.  This definition matches the Rust implementation and the semantics in our paper.


emina (commented):
Unfortunately, this more permissive definition is not okay for analysis. 

Suppose that we have an expression `if * then context.x else context.y` where `x` has type `{name: String}` and `y` has type `{name?: String}`.  This must be rejected by the typechecker in order for the policy to be encodable.  That is, both branches of the conditional must have the exact same type.

emina:
> Well, that's annoying. We'll have to make a breaking change to the Rust typechecker then. This fixes a DRT failure we saw last week

Yeah :-\  That said, it's great that DRT caught it!  I'd consider this a bug in the (strict) Rust typechecker since it's not strict enough.

john-h-kastner-aws:
Closing in favor of https://github.com/cedar-policy/cedar/pull/769

PR #264, opened 2024-03-26
Update integration test CI

khieta:
*Issue #, if available:*

*Description of changes:*

Updates to CI to enable running the integration tests from the `cedar-integration-tests` repository.

(Same idea as [cedar#746](https://github.com/cedar-policy/cedar/pull/746).)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,7 +47,7 @@ jobs:
         run: |
               wget https://raw.githubusercontent.com/leanprover/elan/master/elan-init.sh
               bash elan-init.sh -y
-      - name: rustup
+      - name: Prepare Rust build
         run: rustup update ${{ matrix.toolchain }} && rustup default ${{ matrix.toolchain }}
       - name: cargo fmt (cedar-policy-generators)
         working-directory: ./cedar-policy-generators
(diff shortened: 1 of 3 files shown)


khieta:
Marking as ready for review. Appears to be working as expected with [cedar-integration-tests#3](https://github.com/cedar-policy/cedar-integration-tests/pull/3).

.github/workflows/ci.yml
```diff
@@ -64,22 +64,9 @@ jobs:
       - name: cargo test (cedar-policy-generators)
         working-directory: ./cedar-policy-generators
         run: cargo test --verbose
-      - name: build.sh
+      - name: Run build.sh script
         shell: bash
         run: source ~/.profile && ./build.sh
```
cdisselkoen, line 69: `build.sh` is also run as part of the `run-integration-tests` job/step.  Is this redundant?  Do we need to do it in both places?

cdisselkoen (approved):

.github/workflows/ci.yml
khieta: Technically all that's needed from that script in the `run-integration-tests` jobs is `lake build Cedar:static DiffTest:static Std:static`, but I figured running the whole script in both places would be more consistent. I think it's useful to keep the basic tests separate from the integration tests since the latter may fail for more subtle reasons.

khieta: Updated the `run-integration-tests` job & task names to avoid the confusion

shaobo-he-aws (approved):

khieta:
Added new inputs to the `run-integration-tests` job (`cedar_integration_tests_ref`, `cedar_spec_ref`) to make sure that CI is run against the correct branch in PRs to `cedar-integration-tests`.

Need to make the same change to the CI added in [cedar#746](https://github.com/cedar-policy/cedar/pull/746).

khieta:
Ok, done with edits and ready for a final review.

cdisselkoen (approved):

PR #265, opened 2024-03-27
Update generation of unspecified entities

khieta:
### TLDR

* Only produce unspecified entities with eid "principal", "action", or "resource"
* Perform an extra check during corpus test generation
* Bonus: auto-applied some clippy lints

Requires [cedar#752](https://github.com/cedar-policy/cedar/pull/752) to be merged first.

### Full description

The generators currently allow generation of unspecified entities with arbitrary eids, while the actual `cedar-policy` implementation can only produce unspecified entities with eid "principal", "action", or "resource".

I ran into this issue when re-generating corpus tests for [cedar-integration-tests#3](https://github.com/cedar-policy/cedar-integration-tests/pull/3). A couple produced tests were failing, with a note that the request was supposed to result in an error, but it did not.

The issue turned out to be a policy with a subexpression like `principal == resource && //error`. The generator had assigned both `principal` and `resource` to an unspecified entity with eid "Unspecified", so the equality was satisfied and the error was reached. But when writing out the test case, these unspecified entities were serialized as `None` (the eid was lost). When reading back the test case and running the input through `cedar-policy`, the entities were given eids "principal" and "resource" so the equality was not satisfied and short-circuiting prevented the error from being reached.

This was a little irritating to debug, since the issue was due to a loss of information when serializing the test case. To detect errors like this in the future, I added code to `dump.rs` that will try to run the test on generation. This adds some extra overhead to corpus test generation, but I don't think that's a big issue. Lmk if you disagree.

### DIFF
--- a/cedar-policy-generators/src/hierarchy.rs
+++ b/cedar-policy-generators/src/hierarchy.rs
@@ -95,7 +95,7 @@ impl Hierarchy {
         }
     }
 
-    /// Generate an arbitrary uid based on the hierarchy. If `unspecified_location`
+    /// Generate an arbitrary uid based on the hierarchy. If `request_field`
     /// is `Some(var)` then the generated uid may be unspecified with eid `var`.
     /// Otherwise, the uid is guaranteed to be specified.
     pub fn arbitrary_uid(
--- a/cedar-policy-generators/src/hierarchy.rs
+++ b/cedar-policy-generators/src/hierarchy.rs
@@ -95,8 +95,14 @@ impl Hierarchy {
         }
     }
 
-    /// generate an arbitrary uid based on the hierarchy
-    pub fn arbitrary_uid(&self, u: &mut Unstructured<'_>) -> Result<EntityUID> {
+    /// Generate an arbitrary uid based on the hierarchy. If `unspecified_location`
+    /// is `Some(var)` then the generated uid may be unspecified with eid `var`.
+    /// Otherwise, the uid is guaranteed to be specified.
+    pub fn arbitrary_uid(
+        &self,
+        u: &mut Unstructured<'_>,
+        request_field: Option<ast::Var>,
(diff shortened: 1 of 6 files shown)


cedar-policy-generators/src/hierarchy.rs
```diff
@@ -95,17 +95,34 @@ impl Hierarchy {
         }
     }
 
-    /// generate an arbitrary uid based on the hierarchy
-    pub fn arbitrary_uid(&self, u: &mut Unstructured<'_>) -> Result<EntityUID> {
+    /// Generate an arbitrary uid based on the hierarchy. If `unspecified_location`
```
cdisselkoen, line 98:
```suggestion
    /// Generate an arbitrary uid based on the hierarchy. If `request_field`
```

cdisselkoen (approved):

cedar-policy-generators/src/hierarchy.rs
```diff
@@ -95,17 +95,34 @@ impl Hierarchy {
...
+        request_field: Option<ast::Var>,
+    ) -> Result<EntityUID> {
         // UID that exists or doesn't. 90% of the time pick one that exists
         if u.ratio::<u8>(9, 10)? {
             let uid = u
                 .choose(&self.uids)
                 .map_err(|e| while_doing("getting an arbitrary uid".into(), e))?;
             Ok(uid.clone())
         } else {
-            // Note: may generate an unspecified entity
```
aaronjeline, line 107: Did we want to drop this note?

aaronjeline (approved):
Seems good. The extra overhead doesn't seem like a big deal since this just test generation.

cedar-policy-generators/src/hierarchy.rs
khieta: I think this is covered by other comments added.

PR #275, opened 2024-04-04
Fix docker command in readme

john-h-kastner-aws:
Apparently `--rm` has to come first


khieta (approved):
Thanks for testing this 🙂

khieta:
Also, since you're already here: can you remove the note about having to run `source /root/.profile && source ./set_env_vars.sh` in the Docker container? It's no longer necessary with the .rcfile

khieta (approved):

PR #278, opened 2024-04-09
Andrewmwells/lean upgrade

andrewmwells-amazon:
*Issue #, if available:*

*Description of changes:*
Upgrade to Lean 4.7.0

### DIFF
--- a/cedar-lean/Cedar/Data/List.lean
+++ b/cedar-lean/Cedar/Data/List.lean
@@ -56,9 +56,17 @@ theorem sizeOf_snd_lt_sizeOf_list {α : Type u} {β : Type v} [SizeOf α] [SizeO
   x ∈ xs → sizeOf x.snd < 1 + sizeOf xs
 := by
   intro h
-  have := List.sizeOf_lt_of_mem h
-  have : sizeOf x = 1 + sizeOf x.1 + sizeOf x.2 := rfl
-  omega
+  rw [Nat.add_comm]
+  apply Nat.lt_add_right
+  apply @Nat.lt_trans (sizeOf x.snd) (sizeOf x) (sizeOf xs)
+  {
+    simp [Prod._sizeOf_inst, Prod._sizeOf_1]
+    rw [Nat.add_comm]
+    apply Nat.lt_add_of_pos_right
+    apply Nat.add_pos_left
+    apply Nat.one_pos
+  }
+  { apply List.sizeOf_lt_of_mem; exact h }
 
 def attach₂ {α : Type u} {β : Type v} [SizeOf α] [SizeOf β] (xs : List (α × β)) :
 List { x : α × β // sizeOf x.snd < 1 + sizeOf xs } :=
--- a/cedar-lean/Cedar/Data/List.lean
+++ b/cedar-lean/Cedar/Data/List.lean
@@ -60,7 +60,7 @@ theorem sizeOf_snd_lt_sizeOf_list {α : Type u} {β : Type v} [SizeOf α] [SizeO
   apply Nat.lt_add_right
   apply @Nat.lt_trans (sizeOf x.snd) (sizeOf x) (sizeOf xs)
   {
-    simp [Prod._sizeOf_inst, Prod._sizeOf_1]
+    simp only [sizeOf, Prod._sizeOf_1]
     rw [Nat.add_comm]
     apply Nat.lt_add_of_pos_right
     apply Nat.add_pos_left
--- a/cedar-lean/Cedar/Spec/Wildcard.lean
+++ b/cedar-lean/Cedar/Spec/Wildcard.lean
@@ -70,7 +70,7 @@ def wildcardMatchIdx (text : List Char) (pattern : Pattern) (i j : Nat)
          (← wildcardMatchIdx text pattern (i + 1) (j + 1) (by omega) (by omega))
   modifyGet λ cache => (r, cache.insert (i, j) r)
 termination_by
-  wildcardMatchIdx text pattern i j _ _=> (text.length - i) + (pattern.length - j)
+  (text.length - i) + (pattern.length - j)
 decreasing_by
   all_goals { simp_wf ; omega }
 
--- a/cedar-lean/Cedar/Thm/Authorization/Evaluator.lean
+++ b/cedar-lean/Cedar/Thm/Authorization/Evaluator.lean
@@ -143,29 +143,29 @@ theorem and_produces_bool_or_error {e₁ e₂ : Expr} {request : Request} {entit
 := by
   cases h : evaluate (Expr.and e₁ e₂) request entities <;> simp
   case ok val =>
-    cases val <;> simp [evaluate, Result.as, Coe.coe, Value.asBool, Lean.Internal.coeM, pure, Except.pure] at h <;>
+    cases val <;> simp only [evaluate._eq_4, Result.as, Coe.coe, Value.asBool, Bool.not_eq_true',
+      Lean.Internal.coeM, pure, Except.pure] at h <;>
     generalize (evaluate e₁ request entities) = r₁ at h <;>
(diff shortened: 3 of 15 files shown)


cedar-lean/Cedar/Data/List.lean
```diff
@@ -56,9 +56,17 @@ theorem sizeOf_snd_lt_sizeOf_list {α : Type u} {β : Type v} [SizeOf α] [SizeO
   x ∈ xs → sizeOf x.snd < 1 + sizeOf xs
 := by
   intro h
-  have := List.sizeOf_lt_of_mem h
-  have : sizeOf x = 1 + sizeOf x.1 + sizeOf x.2 := rfl
-  omega
```
andrewmwells-amazon, line 61: Issue with `omega`: https://github.com/leanprover/lean4/issues/3848

cedar-lean/Cedar/Thm/Data/Map.lean
```diff
@@ -139,6 +139,16 @@ theorem sizeOf_lt_of_value [SizeOf α] [SizeOf β] {m : Map α β} {k : α} {v :
   have _ : sizeOf m.1 < sizeOf m := by
     simp only [sizeOf, _sizeOf_1]
     omega
-  omega
```
andrewmwells-amazon, line 142: I'll ask on Zulip, but surely a nicer proof is possible.

cedar-lean/Cedar/Data/List.lean
emina: Interesting.  Do we know why this is a problem in the current version of Lean and wasn't in the older one we were using?  It sounds like this shouldn't have worked in the older version either but somehow did.

cedar-lean/Cedar/Spec/Wildcard.lean
```diff
@@ -70,7 +70,7 @@ def wildcardMatchIdx (text : List Char) (pattern : Pattern) (i j : Nat)
          (← wildcardMatchIdx text pattern (i + 1) (j + 1) (by omega) (by omega))
   modifyGet λ cache => (r, cache.insert (i, j) r)
 termination_by
-  wildcardMatchIdx text pattern i j _ _=> (text.length - i) + (pattern.length - j)
+  (text.length - i) + (pattern.length - j)
```
emina, line 73: Ah, do we have new `termination_by` syntax in this version of Lean?

cedar-lean/Cedar/Thm/Authorization/Evaluator.lean
```diff
@@ -143,29 +143,29 @@ theorem and_produces_bool_or_error {e₁ e₂ : Expr} {request : Request} {entit
 := by
   cases h : evaluate (Expr.and e₁ e₂) request entities <;> simp
   case ok val =>
-    cases val <;> simp [evaluate, Result.as, Coe.coe, Value.asBool, Lean.Internal.coeM, pure, Except.pure] at h <;>
+    cases val <;> simp only [evaluate._eq_4, Result.as, Coe.coe, Value.asBool, Bool.not_eq_true',
```
emina, line 146: We should not use internally generated names like `evaluate._eq_4` in `simp only`.  They are flaky.  

https://github.com/leanprover/lean4/issues/3547

I manually change all of them to just the externally visible names when they are suggested, in this case, `evaluate`.

cedar-lean/Cedar/Thm/Validation/Typechecker/UnaryApp.lean
```diff
@@ -122,7 +122,7 @@ theorem type_of_neg_is_sound {x₁ : Expr} {c₁ c₂ : Capabilities} {env : Env
       have ⟨i, h₈⟩ := instance_of_int_is_int h₇
       subst h₈
       simp [apply₁, intOrErr]
-      cases h₉ : i.neg?
+      cases _h₉ : i.neg?
```
emina, line 125: Curious:  why do we need the underscore before `h9` here?

emina (changes requested):

cedar-lean/Cedar/Spec/Wildcard.lean
andrewmwells-amazon: Yeah, https://lean-lang.org/blog/2024-2-29-lean-460/

cedar-lean/Cedar/Data/List.lean
andrewmwells-amazon: On zulip, Joachim said: "There was a change where omega would no longer check atoms for definitional equality, merely syntactic equal."
But it sounds like there's also an issue with `omega`. I don't know enough to pretend to understand the root cause.

cedar-lean/Cedar/Thm/Validation/Typechecker/UnaryApp.lean
andrewmwells-amazon: To suppress a warning about it being unused. I'm sure there's some syntax to just omit it, but `_` doesn't work.

emina: Got it.

emina (approved):

shaobo-he-aws (approved):

cedar-lean/Cedar/Thm/Data/LT.lean
```diff
@@ -55,7 +55,7 @@ theorem IPNet.lt_conn {a₁ a₂ p₁ p₂ : Nat}
 
 instance IPNet.strictLT : StrictLT Ext.IPAddr.IPNet where
   asymmetric a b   := by
-    cases a <;> cases b <;> simp [LT.lt, Ext.IPAddr.IPNet.lt]
+    cases a <;> cases b <;> simp only [LT.lt] <;> simp only [Ext.IPAddr.IPNet.lt, UInt32.val_val_eq_toNat, Nat.lt_eq, Bool.or_eq_true, decide_eq_true_eq, Bool.and_eq_true, not_false_eq_true, not_true_eq_false, imp_self]
```
khieta, line 58: I wonder if we should be defining some custom tactics for long `simp only` clauses. It seems like the instances in this file at least have a lot of overlap.

cedar-lean/Cedar/Spec/Wildcard.lean
khieta, line 73: Nit: since these clauses are so much shorter with the update, I think they'd look nicer on one line.

emina:
> Lgtm!
> 
> Are the style/maintainability conventions we've been using (e.g., `simp only` instead of `simp`) documented somewhere? I'd feel bad if I made edits later that undid your, @cdisselkoen's, and @emina's hard work.

Good question.  I don't think we've updated the README etc to add these conventions.  (We adopted them from Mathlib.)

Here are the three that come to mind, and @cdisselkoen may remember more:

* Use `simp only` instead of `simp`.  It's okay to use `simp` to close a goal, but I personally just use `simp only` everywhere.  (Exceptions mean I forgot to do it).
* Use `have` to deconstruct values.  Use `rcases` only to split disjunctions. 
* Use `exact` instead of `apply` whenever possible.
* Fully spell out types in function and theorem declarations.  (You can often omit types and even variables, but I find things much easier to read quickly when the type declarations are spelled out.)

cedar-lean/Cedar/Thm/Data/LT.lean
andrewmwells-amazon: Seems like a good idea but I don't know how to do it quickly and want to get this merged now so I'll open an issue: https://github.com/cedar-policy/cedar-spec/issues/279

shaobo-he-aws (approved):

cedar-lean/Cedar/Thm/Validation/Typechecker/UnaryApp.lean
cdisselkoen: `cases i.neg?` should work

PR #280, opened 2024-04-10
Update guide and set default target

andrewmwells-amazon:
*Issue #, if available:*

*Description of changes:*
Update Lean style guide and set Cedar as the default target.

### DIFF
--- a/cedar-lean/GUIDE.md
+++ b/cedar-lean/GUIDE.md
@@ -60,3 +60,12 @@ To break up implications across multiple lines:
 ## Comments
 
 Main theorems should have a docstring comment `/-- ... -/` explaining what is proven. See [here](https://leanprover-community.github.io/contribute/doc.html) for more information on the Mathlib documentation style.
+
+
+## Proof stability
+To make version upgrades easier, strive to follow these guidelines:
+
+- Use `simp only` instead of `simp`. It's okay to use `simp` to close a goal.
+- Use `have` to deconstruct values. Use `rcases` only to split disjunctions.
+- Use `exact` instead of `apply` whenever possible.
+- Fully spell out types in function and theorem declarations.
--- a/cedar-lean/lakefile.lean
+++ b/cedar-lean/lakefile.lean
@@ -22,6 +22,7 @@ require std from git
 
 package Cedar
 
+@[default_target]
 lean_lib Cedar where
   defaultFacets := #[LeanLib.staticFacet]
 


emina (approved):

khieta (approved):

cedar-lean/GUIDE.md
```diff
@@ -60,3 +60,12 @@ To break up implications across multiple lines:
 ## Comments
 
 Main theorems should have a docstring comment `/-- ... -/` explaining what is proven. See [here](https://leanprover-community.github.io/contribute/doc.html) for more information on the Mathlib documentation style.
+
+
+## Proof stability
+To make version upgrades easier, strive to follow these guidelines:
+
+- Use `simp only` instead of `simp`. It's okay to use `simp` to close a goal.
```
cdisselkoen, line 68: I'm still not the biggest fan of this advice.  It creates a burden on all uses of `simp` in the codebase, even though I claim that most of them will never be broken by any version upgrade.  It also reduces readability of our proofs.  An alternate convention would be to convert to `simp only` only when a version upgrade problem is actually encountered.  However, I don't feel strongly about this and am happy to follow the `simp only` convention if everyone else agrees on it.

cdisselkoen (approved):

cedar-lean/GUIDE.md
cdisselkoen: Somewhat relatedly: should we encode a guideline that we prefer `rw` to `simp only` when possible?  One advantage of `rw` is that it has a clear failure state if the given rewrite is not possible.  `simp only [h]` can still succeed without using `h`, which can lead to confusion.

andrewmwells-amazon: It's a convention in `Mathlib`, which we roughly try to follow. Agree it doesn't look as nice, but as Kesha pointed out, maybe we should create custom tactics for commonly used sets of `simp` rules. In support of following Mathlib, the authors have more experience than us. OTOH, Mathlib is much larger so the same rules may not apply. Anecdotally, I think a lot (I'd guess more than 25%) of our broken proofs could be fixed by this.

andrewmwells-amazon: I'm also ok with using `rw` when possible, but I don't know of a tool that tells you when it's possible, so I'm not sure how we'd easily check this.

emina: I'm happy to follow `simp only` based on Mathlib's experience. 

I use `rw` when necessary, but in general, I find it too tedious and slow to use compared to `simp only`.  My workflow tends to be to use `simp` and add all the custom lemmas needed to reduce to the desired shape.  Then, I replace this `simp` with `simp only`.  It would be slower (and longer) to replace that with `rw` in general, since `simp only` will also use, e.g., commutativity of equality to match up goals and lemma statements.  With `rw`, you'd have to apply `eq_comm` in all such places manually. I end up using `rw` very selectively, when I want to rewrite a specific part of the goal into a specific shape.

cdisselkoen: That's good enough for me, we can recommend `simp only` over `simp` and `rw`.

I remembered another convention we established a while ago: don't `open Except`

PR #281, opened 2024-04-11
add implementation of partial evaluation

cdisselkoen:
Much more coming, but I thought I'd try to split this up into smaller PRs for review.

This PR contains _only_ the implementation code for the partial evaluator.  It does _not_ contain the partial authorizer, nor does it contain basically any of the proofs.  If it's unacceptable to take this PR without the accompanying proofs, I can close this and re-post a (significantly) larger PR once the proofs are ready (enough).

~I particularly don't have confidence in `Decimal.unParse` and `IPNet.unParse` as I haven't tested them and don't have any proofs about them even on my own branch.  Review appreciated.~ Edit: tested 

~Looking for suggestions on how best to finish off the termination proofs that are currently using `sorry`.~ Edit: thanks @emina for help


cedar-lean/Cedar/Spec/PartialExpr.lean
```diff
@@ -0,0 +1,212 @@
...
+  | .ext (.decimal d) => .call ExtFun.decimal [PartialExpr.lit (.string d.unParse)]
+  | .ext (.ipaddr ip) => .call ExtFun.ip [PartialExpr.lit (.string (Cedar.Spec.Ext.IPAddr.unParse ip))]
+termination_by v.size
+
+/--
+  A version of `PartialExpr`, but only allows "restricted expressions" -- no
+  vars, no expressions that require entity data to evaluate, no operators at all,
+  just literals, unknowns, extension values, and sets/records of those things
+-/
+inductive RestrictedPartialExpr where
```
emina, line 68: Curious:  why do we need `RestrictedPartialExpr`s?

Depending on what they are used for (e.g., just proofs?), it may be easier to define an inductive predicate that restricts `PartialExpr`s rather than defining a whole new type (and presumably translating back and forth between regular and restricted partial exprs).

cedar-lean/Cedar/Spec/PartialExpr.lean
```diff
@@ -0,0 +1,212 @@
...
+  | record (map : List (Attr × RestrictedPartialExpr))
+  | call (xfn : ExtFun) (args : List Value) -- this requires that all arguments to extension functions in RestrictedPartialExpr are concrete. TODO do we need to relax this?
+  | unknown (name : String)
+
+deriving instance Repr, Inhabited for RestrictedPartialExpr
+
+def Value.asRestrictedPartialExpr (v : Value) : RestrictedPartialExpr :=
+  match v with
+  | .prim p => .lit p
+  | .set s => .set (s.elts.map₁ λ ⟨v, h₁⟩ =>
```
emina, line 80: I'm hoping we can simplify this, to get rid of the calls to the termination lemmas.  I'll play it with a bit offline to see what we can do.

cedar-lean/Cedar/Spec/PartialExpr.lean
```diff
@@ -0,0 +1,212 @@
...
+  | and (a : PartialExpr) (b : PartialExpr)
+  | or (a : PartialExpr) (b : PartialExpr)
+  | unaryApp (op : UnaryOp) (expr : PartialExpr)
+  | binaryApp (op : BinaryOp) (a : PartialExpr) (b : PartialExpr)
+  | getAttr (expr : PartialExpr) (attr : Attr)
+  | hasAttr (expr : PartialExpr) (attr : Attr)
+  | set (ls : List PartialExpr)
+  | record (map : List (Attr × PartialExpr))
+  | call (xfn : ExtFun) (args : List PartialExpr)
+  | unknown (name : String)
```
emina, line 46: I'd recommend defining an abbreviation for unknown and using that everywhere instead of `String`, just in case we change our mind about how to model unknowns.

```
abbrev Unknown := String

...

  |  unknown (u : Unknown)
```

cedar-lean/Cedar/Spec/PartialRequest.lean
```diff
@@ -0,0 +1,62 @@
...
+This file defines Cedar partial requests.
+-/
+
+namespace Cedar.Spec
+
+open Cedar.Data
+
+inductive UidOrUnknown where
+  | known (uid : EntityUID)
+  | unknown (name : String)
```
emina, line 33: Replace `String` with `Unknown` (abbrev).

cedar-lean/Cedar/Spec/PartialRequest.lean
```diff
@@ -0,0 +1,62 @@
...
+instance : Coe UidOrUnknown PartialValue where
+  coe x := match x with
+  | .known uid    => .value uid
+  | .unknown name => .residual (PartialExpr.unknown name)
+
+structure PartialRequest where
+  principal : UidOrUnknown
+  action : UidOrUnknown
+  resource : UidOrUnknown
+  context : Map Attr RestrictedPartialValue -- allows individual context attributes to have unknown values, but does not allow it to be unknown whether a context attribute exists at all
```
emina, line 46: I don't follow this comment:  how is the use of RestrictedPartialValue (vs PartialValue) related to the presence of an attribute/value binding in this map?

(Or maybe "attribute exists" here doesn't refer to the map's contents?  If so, what does it refer to?)

cedar-lean/Cedar/Thm/Termination/Expr.lean
```diff
@@ -0,0 +1,118 @@
...
+ limitations under the License.
+-/
+
+import Cedar.Spec.Expr
+
+namespace Cedar.Spec
+
+open Cedar.Data
+
+def Expr.size : Expr -> Nat
```
emina, line 23: You shouldn't need to define this `size` function.  Lean automatically generates these for you, so you can write `sizeOf e` when `e` is an Expr.

emina (changes requested):

cedar-lean/Cedar/Spec/Evaluator.lean
```diff
@@ -78,43 +78,41 @@ def bindAttr (a : Attr) (res : Result Value) : Result (Attr × Value) := do
...
-  | .lit l       => .ok l
-  | .var var     =>
-    match var with
-    | .principal => .ok req.principal
-    | .action    => .ok req.action
-    | .resource  => .ok req.resource
-    | .context   => .ok req.context
-  | .ite x₁ x₂ x₃ => do
+  | .lit l           => .ok l
+  | .var .principal  => .ok req.principal
```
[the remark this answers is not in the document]
cdisselkoen: Fixed

cedar-lean/Cedar/Spec/Ext/Decimal.lean
```diff
@@ -58,6 +58,11 @@ def parse (str : String) : Option Decimal :=
     else .none
   | _ => .none
 
+def unParse (d : Decimal) : String :=
```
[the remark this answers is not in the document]
cdisselkoen: Fixed. Thanks for the catch

cedar-lean/Cedar/Spec/Ext/Decimal.lean
```diff
@@ -58,6 +58,11 @@ def parse (str : String) : Option Decimal :=
...
+  let left  := d / (Int.pow 10 DECIMAL_DIGITS)
+  let right := d % (Int.pow 10 DECIMAL_DIGITS)
+  ToString.toString left ++ "." ++ ToString.toString right
```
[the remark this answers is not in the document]
cdisselkoen: TIL about Lean string interpolation.  Fixed

cedar-lean/Cedar/Spec/Ext/IPAddr.lean
```diff
@@ -246,6 +246,37 @@ def parse (str : String) : Option IPNet :=
   let ip := parseIPv4Net str
   if ip.isSome then ip else parseIPv6Net str
 
+def unParse (ip : IPNet) : String :=
```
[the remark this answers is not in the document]
cdisselkoen: I looked for a way to convert to hex and didn't find one -- any pointers?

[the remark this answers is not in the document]
cdisselkoen: Wrote my own using the builtin `hexDigitRepr`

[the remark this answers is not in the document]
cdisselkoen: (and resolved the correctness issues -- unless you see more!)

cedar-lean/Cedar/Spec/PartialEntities.lean
```diff
@@ -0,0 +1,82 @@
...
+-/
+
+import Cedar.Spec.Entities
+import Cedar.Spec.PartialValue
+
+/-!
+This file defines Cedar partial-entities structures.
+-/
+
+namespace Cedar.Spec
```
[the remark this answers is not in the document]
cdisselkoen: Moved

cedar-lean/Cedar/Spec/ExtFun.lean
```diff
@@ -42,7 +42,7 @@ def res {α} [Coe α Ext] : Option α → Result Value
   | some v => .ok v
   | none   => .error .extensionError
 
-def call : ExtFun → List Value → Result Value
+def ExtFun.call : ExtFun → List Value → Result Value
```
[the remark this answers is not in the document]
cdisselkoen: I think it was for syntactic sugar -- if you have `xfn : ExtFun`, now you can write `xfn.call args`, which looks pleasing.  But this PR is not using that sugar, and has no other need for the qualifier.  I will remove it from this PR.

cedar-lean/Cedar/Spec/PartialRequest.lean
cdisselkoen: The comment is not about the use of RestrictedPartialValue vs PartialValue.  It is generally about the definition as `Map Attr RestrictedPartialValue` -- that is, each `Attr` is either in the map or not, and if it is in the map, it definitely has a (partial) value.  An alternate definition that would allow attributes to have unknown existence would be something like `Map Attr (Option RestrictedPartialValue)`, or something isomorphic

cedar-lean/Cedar/Spec/PartialRequest.lean
cdisselkoen: Done

cedar-lean/Cedar/Spec/PartialExpr.lean
cdisselkoen: They are used as the residual type for `RestrictedPartialValue`.  Obviously, that leads to the question "why do we need `RestrictedPartialValue`.

`RestrictedPartialValue` is the type of attributes in a `PartialEntityData` and in the `context` of a `PartialRequest`.  It is clear that we cannot allow fully general `PartialValue` in these locations, because that would allow entity/context attribute values to be residual expressions that depend on the values of other entity/context attributes, or of the variables `principal`, `action`, `resource`.  However, you make a good suggestion that we could simply require some well-formedness predicate on `PartialEntityData` and `PartialRequest` that requires that any residuals must only use one of the allowed constructors/variants.  (Indeed my proofs already need well-formedness predicates on these, eg ensuring that the underlying `Map`s are `WellFormed`, so it would be easy to add another requirement to the definitions of well-formedness there.)

Similarly, `RestrictedPartialValue` is the type allowed in a substitution-for-unknowns operation -- when providing the value for an `unknown`, you can provide any `RestrictedPartialValue`, but it clearly does not make sense to allow providing a fully general `PartialValue` and thus allowing the value of the `unknown` to depend on Cedar residual expressions involving entity or request data.  However, this could be expressed as a precondition instead of at the type level.

cdisselkoen: I guess the higher-level design question is, in situations like this when we need to establish constraints on the inputs users provide (in this case, user-provided partial values for attribute values, context values, and unknown substitutions), do we prefer expressing this at the type level (so any value of the given type is valid), or via preconditions on the top-level methods.

cedar-lean/Cedar/Spec/PartialExpr.lean
```diff
@@ -0,0 +1,212 @@
...
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-/
+
+import Cedar.Data
+import Cedar.Spec.Expr
+import Cedar.Thm.Data.Map -- necessary for a termination argument
```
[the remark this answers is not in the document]
cdisselkoen: Feels bad to have general-purpose lemmas about `Map` that live in PE code.  Does it make sense to have an exception for `Cedar.Thm.Data`?  I totally understand not wanting dependencies from `Spec` (or the new `Partial`) to other parts of `Thm`

cedar-lean/Cedar/Spec/Ext/IPAddr.lean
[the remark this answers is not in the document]
emina: Yeah, my suggestion would have been to roll your own hex conversion (we've done so for the parser).  I couldn't find any built-in ones.

cedar-lean/Cedar/Spec/PartialExpr.lean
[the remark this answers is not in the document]
emina: Yeah, if we can't find a way to get around including these termination proofs, I'd probably suggest putting the relevant `Map` lemmas into the `Cedar.Data.Map` file.  Not entirely satisfying either though.

cedar-lean/Cedar/Spec/PartialExpr.lean
emina: It depends on a couple of things.  Here is my rough mental checklist for deciding on one vs the other:

(1) Is expressing this at the type level making the specification clearer to a reader?
(2) Is expressing this at the type level making the algorithm (in this case PE) shorter / easier to read / etc. in some way?  For example, does having this type help you return `.ok` (vs `.error`) in more places?
(3) Does the algorithm make sense without the type-level constraint?

These are pretty subjective, but if the answer is "yes"-ish for 1-2 and "no"-ish for 3, then I keep it at the type level.  Otherwise, it becomes a well-formedness (WF) predicate. 

For example, in the symbolic compilation algorithm, we use WF predicates because (a) the WF constraints are subtle and cannot be expressed as ADT types (we'd have to use dependent types), and (b) the algorithm does _something_ on any input, and we prove that it does what we want on WF inputs.

cedar-lean/Cedar/Partial/Evaluator.lean
```diff
@@ -0,0 +1,159 @@
...
+
+def getAttr (v : Spec.Value) (a : Attr) (es : Partial.Entities) : Result Partial.RestrictedValue := do
+  let r ← Partial.attrsOf v es.attrs
+  r.findOrErr a attrDoesNotExist
+
+def bindAttr (a : Attr) (res : Result Partial.Value) : Result (Attr × Partial.Value) := do
+  let v ← res
+  .ok (a, v)
+
+def evaluate (x : Partial.Expr) (req : Partial.Request) (es : Partial.Entities) : Result Partial.Value :=
```
emina, line 73: In my experience, proofs are easier to structure if the code for each of these big match branches is in a separate function, whenever possible.  (This is why we broke up the concrete evaluator and typechecker code in this way.)

It becomes easier to write "inversion" lemmas that describe the behavior of the evaluation on a given case declaratively.  And these lemmas in turn make proofs easier to do.

YMMV of course :)

cedar-lean/Cedar/Spec/Ext/IPAddr.lean
```diff
@@ -264,6 +291,15 @@ instance : LT IPNet where
 instance IPNet.decLt (d₁ d₂ : IPNet) : Decidable (d₁ < d₂) :=
 if h : IPNet.lt d₁ d₂ then isTrue h else isFalse h
 
+theorem test1 : unParse ((parse "192.168.0.1/32").get!) = "192.168.0.1/32" := by decide
```
emina, line 294: I suggest moving these to `UnitTest.IPAddr`: https://github.com/cedar-policy/cedar-spec/blob/main/cedar-lean/UnitTest/IPAddr.lean

cedar-lean/Cedar/Spec/Ext/Decimal.lean
```diff
@@ -58,6 +58,28 @@ def parse (str : String) : Option Decimal :=
...
+  let right := d % (Nat.pow 10 DECIMAL_DIGITS)
+  let right :=
+    -- this is not generalized for arbitrary DECIMAL_DIGITS
+    if right < 10 then s!".000{right}"
+    else if right < 100 then s!".00{right}"
+    else if right < 1000 then s!".0{right}"
+    else s!".{right}"
+  s!"{neg}{left}{right}"
+
+theorem test1 : unParse ((parse "3.14").get!) = "3.1400" := by decide
```
emina, line 74: I suggest moving these to `UnitTest.Decimal`: https://github.com/cedar-policy/cedar-spec/blob/main/cedar-lean/UnitTest/Decimal.lean

cedar-lean/Cedar/Spec/PartialExpr.lean
emina: Turns out we can avoid these explicit proofs :)

Here is the one for `Expr.asPartialExpr`: https://github.com/cedar-policy/cedar-spec/blob/7d05c98c753b1208aa803894842180d83b83e3a9/cedar-lean/Cedar/Partial/Expr.lean#L167

Others should be similar.

The key is to use `.attach₂` to map over records in a way that ensures termination.

cdisselkoen: Thanks for this, this is helpful

cedar-lean/Cedar/Spec/Ext/Decimal.lean
cdisselkoen: Moved (and the ip tests)

cedar-lean/Cedar/Spec/Ext/IPAddr.lean
cdisselkoen: Moved (and the Decimal tests)

cedar-lean/Cedar/Spec/PartialExpr.lean
cdisselkoen: I think that regardless of the answer in the full context of existing and planned PE proofs, it's certainly clear that the need for the `Restricted*` types is not justified in the current PR.  So I will take them out here, and if/when they are needed, we can reopen this discussion in the future PR where we have a concrete justification.

cdisselkoen: Removed from the current PR.

cedar-lean/Cedar/Partial/Evaluator.lean
cdisselkoen: Noted.  But let's take the current PR as-is and refactor in a future PR if it makes proofs easier?

cedar-lean/Cedar/Spec/PartialExpr.lean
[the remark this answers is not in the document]
cdisselkoen: Latest version of this PR has termination proofs that do not require dependencies on Cedar.Thm.Data.  Resolving this.

emina (approved):

cedar-lean/Cedar/Partial.lean
```diff
@@ -0,0 +1,21 @@
+/-
+ Copyright 2022-2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
andrewmwells-amazon, line 2:
```suggestion
 Copyright Cedar Contributors. All Rights Reserved.
```

cedar-lean/Cedar/Partial/Entities.lean
```diff
@@ -0,0 +1,90 @@
+/-
+ Copyright 2022-2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
andrewmwells-amazon, line 2:
```suggestion
 Copyright Cedar Contributors. All Rights Reserved.
```

cedar-lean/Cedar/Partial/Evaluator.lean
```diff
@@ -0,0 +1,159 @@
+/-
+ Copyright 2022-2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
andrewmwells-amazon, line 2:
```suggestion
 Copyright Cedar Contributors. All Rights Reserved.
```

shaobo-he-aws (approved):

cedar-lean/Cedar/Partial/Evaluator.lean
andrewmwells-amazon: Was about to say this would be easier to read with functions for each case, but luckily I thought I'd check what this comment said first 😄

cedar-lean/Cedar/Partial/Request.lean
```diff
@@ -0,0 +1,66 @@
+/-
+ Copyright 2022-2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
andrewmwells-amazon, line 2:
```suggestion
 Copyright Cedar Contributors. All Rights Reserved.
```

cedar-lean/Cedar/Partial/Value.lean
```diff
@@ -0,0 +1,37 @@
+/-
+ Copyright 2022-2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
andrewmwells-amazon, line 2:
```suggestion
 Copyright Cedar Contributors. All Rights Reserved.
```

cedar-lean/Cedar/Spec/Ext/Decimal.lean
```diff
@@ -58,6 +58,19 @@ def parse (str : String) : Option Decimal :=
     else .none
   | _ => .none
 
+def unParse (d : Decimal) : String :=
```
andrewmwells-amazon, line 61: Why `unParse` and not `toString` or similar?

cedar-lean/Cedar/Partial/Expr.lean
```diff
@@ -0,0 +1,158 @@
+/-
+ Copyright 2022-2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
andrewmwells-amazon, line 2:
```suggestion
 Copyright Cedar Contributors. All Rights Reserved.
```

andrewmwells-amazon (approved):

cedar-lean/Cedar/Partial/Expr.lean
```diff
@@ -0,0 +1,158 @@
...
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-/
+
+import Cedar.Data
+import Cedar.Spec.Expr
+
+/-! This file defines abstract syntax for Cedar expressions. -/
```
khieta, line 20:
```suggestion
/-! This file defines abstract syntax for Cedar partial expressions. -/
```

cedar-lean/Cedar/Spec/Ext/Decimal.lean
cdisselkoen: No reason lol

cdisselkoen:
The old 2022-2023 copyright line is still used extensively throughout cedar-lean (not just in my new files).  I will merge this and then do a followup PR to change the copyright everywhere

cedar-lean/Cedar/Partial/Entities.lean
```diff
@@ -0,0 +1,90 @@
...
+/--
+Represents the information about one entity.
+
+Currently, this allows attrs to be known-to-exist-but-unknown-value,
+but does not allow attrs to be unknown-whether-they-exist-entirely.
+(the result of `e has attr` is never a residual for an `e` that is known to exist.)
+
+Currently, this does not allow any unknowns about ancestor information.
+All ancestor information must be fully concrete.
+-/
```
khieta, line 38: Is this also the current state of the Rust implementation?

cedar-lean/Cedar/Partial/Evaluator.lean
```diff
@@ -0,0 +1,159 @@
...
+
+import Cedar.Partial.Entities
+import Cedar.Partial.Expr
+import Cedar.Partial.Request
+import Cedar.Partial.Value
+import Cedar.Spec.Evaluator
+import Cedar.Spec.ExtFun
+import Cedar.Spec.Value
+
+/-! This file defines the semantics of Cedar partial evaluation. -/
```
khieta, line 25: Would be useful to have comments here about which functions are analogous to the ones in `Spec.Evaluator`, and which ones are different.

khieta (approved):
I see that my review was too slow & this was already merged 😄 Still a useful exercise for me to read through this addition to the formalization.

cedar-lean/Cedar/Spec/Ext/Decimal.lean
cdisselkoen: I'll do the ToString instance as a followup

cedar-lean/Cedar/Partial/Entities.lean
cdisselkoen: I believe so, but @aaronjeline is the authority on that

cedar-lean/Cedar/Partial/Evaluator.lean
cdisselkoen: Will do in a followup

cedar-lean/Cedar/Partial/Evaluator.lean
cdisselkoen: Alright I that's several votes, so I will split it in a followup

cedar-lean/Cedar/Spec/Ext/Decimal.lean
cdisselkoen: #286

cedar-lean/Cedar/Partial/Evaluator.lean
cdisselkoen: #286

cedar-lean/Cedar/Partial/Expr.lean
cdisselkoen: Fixed in #286

PR #283, opened 2024-04-11
Restrict Std imports to what we actually need

andrewmwells-amazon:
*Issue #, if available:*

*Description of changes:*
Restrict Std imports to what we actually need

### DIFF
--- a/cedar-lean/Cedar/Spec/Wildcard.lean
+++ b/cedar-lean/Cedar/Spec/Wildcard.lean
@@ -14,7 +14,7 @@
  limitations under the License.
 -/
 
-import Std
+import Std.Data.HashMap
 
 namespace Cedar.Spec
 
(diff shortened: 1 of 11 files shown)


cdisselkoen (approved):

cedar-lean/Cedar/Spec/Wildcard.lean
```diff
@@ -46,7 +47,7 @@ def wildcard : PatElem → Bool
   | .star => true
   | _     => false
 
-abbrev Cache := HashMap (Nat × Nat) Bool
+abbrev Cache := Std.HashMap (Nat × Nat) Bool
```
emina, line 50: Do you know why we have to write `Std.HashMap` here even though we have `open Std` at the top?

Would be nicer to have just `HashMap`.

emina (approved):

emina (approved):

PR #291, opened 2024-04-24
Proof that PE on concrete inputs is equivalent to concrete eval

cdisselkoen:
The first (of many) proofs about partial evaluation: namely, that partial evaluation applied to concrete inputs gives the same answers as concrete evaluation.

Looking for advice or help on a few unfinished bits in this PR:
* ~In `mapM_eq_some` in Thm/Data/List.lean, I think we need to somehow perform induction over two lists simultaneously.  (We know or can prove that the lists are equal length.)  I'm not sure how to do this, or maybe some alternate approach is better/easier/clearer.~
    * Edit: solved this after I discovered `induction xs generalizing ys`
* ~Termination argument for the main theorem, `on_concrete_eqv_concrete_eval` in Thm/Partial/Evaluation.lean, which I have working for all cases except the Record case.  Currently, Lean complains it is unable to prove termination _even if_ we admit the property that should be necessary with `sorry` (a line which the current PR includes just for clarity).  @emina you were very helpful with termination arguments previously; any advice here?~
    * Edit: fixed with some advice from @emina, thanks!
* How to prove some simple theorems that I believe should be consequences of well-formedness in maps: (or alternately, rework the proofs that currently use these, so that these are not needed)
    * `key_maps_to_one_value` in Thm/Data/Map.lean
    * `equal_keys_same_value` in Thm/Data/Map.lean
* How to prove well-formedness for the results of `Map.mapOnValues` and `Map.mapMOnValues`
    * `mapOnValues_wf` in Thm/Data/Map.lean
    * `mapMOnValues_wf` in Thm/Data/Map.lean
* And one lemma involving `List.Equiv` and `Map.make` i.e. `List.canonicalize Prod.fst`:
    * `make_eqv` in Thm/Data/Map.lean


cedar-lean/Cedar/Thm/Authorization/Authorizer.lean
```diff
@@ -269,16 +269,6 @@ theorem satisfied_implies_resource_scope {policy : Policy} {request : Request} {
         (and_true_implies_right_true h₁))
   exact resource_eval_ok_means_resource_in_uid h₁ h₂
 
-/--
-  A generic lemma that relates List.mapM to List.map. Not in Std AFAICT.
--/
-theorem if_f_produces_pure_then_mapM_f_is_pure_map {α β} [Monad m] [LawfulMonad m] {f : α → β} {list : List α} :
```
cdisselkoen, line 275: Moved to Thm/Data/List.lean and renamed `mapM_pure`

cedar-lean/Cedar/Thm/Authorization/Authorizer.lean
```diff
@@ -310,80 +300,37 @@ theorem mapM_asEntityUID_of_uid {uids : List EntityUID} :
...
-  case cons x xs h_ind =>
+  case cons x xs ih =>
     rw [List.map_cons, List.mapM_cons]
-    simp [pure, Except.pure, asEntityUID_of_uid, h_ind]
+    simp [pure, Except.pure, asEntityUID_of_uid, ih]
 
-/--
-  A generic lemma about the behavior of List.mapM' in the Except monad
--/
-theorem mapM'_ok_iff_f_ok_on_all_elements {f : α → Except ε β} {list : List α} :
```
cdisselkoen, line 320: Moved to Thm/Data/List.lean, reformulated in terms of `mapM` instead of `mapM'`, and renamed `isOk_mapM`

cedar-lean/Cedar/Thm/Data/List.lean
```diff
@@ -114,70 +114,9 @@ theorem filterMap_equiv (f : α → Option β) (xs ys : List α) :
...
-        apply h₁ a
-        exact List.mem_cons_of_mem x h₂
-      case h_2 b h₂ =>
-        exfalso
-        specialize h₁ x
-        simp at h₁
-        simp [h₁] at h₂
-
-theorem filterMap_nonempty_iff_exists_f_returns_some {f : α → Option β} {xs : List α} :
-  xs.filterMap f ≠ [] ↔ ∃ x ∈ xs, (f x).isSome
```
cdisselkoen, line 155: Moved to a dedicated section for `filterMap`, and renamed `filterMap_nonempty_iff_exists`

cedar-lean/Cedar/Thm/Data/List.lean
```diff
@@ -114,70 +114,9 @@ theorem filterMap_equiv (f : α → Option β) (xs ys : List α) :
...
-  case mpr =>
-    intro h₁ h₂
-    rw [filterMap_empty_iff_f_returns_none] at h₂
-    replace ⟨x, h₁, h₃⟩ := h₁
-    specialize h₂ x h₁
-    simp [h₂, Option.isSome] at h₃
-
-theorem f_implies_g_then_subset {f g : α → Option β} {xs : List α} :
-  (∀ a b, f a = some b → g a = some b) →
-  xs.filterMap f ⊆ xs.filterMap g
```
cdisselkoen, line 173: Moved to a dedicated section for `filterMap`

cedar-lean/Cedar/Thm/Data/List.lean
```diff
@@ -718,21 +656,117 @@ theorem any_of_mem {f : α → Bool} {x : α} {xs : List α}
...
+  (∀ x ∈ l, f x = g x) → map f l = map g l
+  | [], _ => rfl
+  | a :: l, h => by
+    let ⟨h₁, h₂⟩ := forall_mem_cons.1 h
+    rw [map, map, h₁, map_congr h₂]
+
+/--
+  Copied from Mathlib. We can delete this if it gets added to Std.
+-/
+theorem map_pmap_subtype
```
cdisselkoen, line 683: Moved here from below. Instead of a dedicated section for "theorems lifted from Mathlib", I put them in the sections where the logically belong (in this case, the "map and map1" section)

cedar-lean/Cedar/Thm/Data/List.lean
```diff
@@ -755,21 +789,170 @@ theorem mapM_implies_nil {f : α → Except β γ} {as : List α}
...
+    replace ⟨v, h₁, h₂⟩ := h₁
+    rw [← Option.not_isSome_iff_eq_none]
+    intro h₃
+    rw [isSome_mapM] at h₃
+    specialize h₃ v h₁
+    simp [h₂] at h₃
+
+/--
+  Possibly the `mp` direction of `isSome_mapM` could be a corollary of this?
+-/
```
cdisselkoen, line 884: For the corresponding theorems in Thm/Data/Map.lean, I did indeed prove the `mp` direction of `isSome_mapMOnValues` as a corollary of `mapMOnValues_eq_some`.  I held off on doing that refactoring here for the moment because (a) we do have a perfectly valid proof of `isSome_mapM` already, without `mapM_eq_some`, and (b) I wanted to wait until we completed the proof of `mapM_eq_some` before introducing more dependencies on it, just in case it turns out to be invalid or the statement need tweaking

cedar-lean/Cedar/Thm/Data/Map.lean
```diff
@@ -54,76 +166,271 @@ theorem not_contains_of_empty {α β} [BEq α] (k : α) :
...
 := by
-  simp [Map.WellFormed] at h₁
-  simp [Map.mapOnValues, Map.toList, Map.kvs, Map.make] at *
-  rw [h₁] ; simp only ; rw [eq_comm]
-  have h₂ : Prod.map id f = (fun (x : α × β) => (x.fst, f x.snd)) := by unfold Prod.map ; simp only [id_eq]
-  simp only [← h₂, ← List.canonicalize_of_map_fst, List.canonicalize_idempotent]
-
-theorem find?_mem_toList {α β} [LT α] [DecidableLT α] [DecidableEq α] {m : Map α β} {k : α} {v : β}
-  (h₁ : m.find? k = .some v) :
-  (k, v) ∈ m.toList
```
cdisselkoen, line 109: I needed the converse of this as well, so I reformulated as an `iff`, renamed to `in_list_some_find?`, and reformulated using `.kvs` instead of `.toList` (which are synonymous, I'm not sure why we have both).  I needed to introduce a well-formedness precondition, but that wasn't a problem at any existing usage sites.

cedar-lean/Cedar/Thm/Data/List.lean
cdisselkoen: Since I've now resolved the problem with `mapM_eq_some`, I also refactored so `isSome_mapM.mp` is a corollary of it, which simplifies the proof substantially.

cedar-lean/Cedar/Thm/Data/Map.lean
```diff
@@ -54,76 +169,274 @@ theorem not_contains_of_empty {α β} [BEq α] (k : α) :
 theorem make_wf [LT α] [StrictLT α] [DecidableLT α] (xs : List (α × β)) :
   WellFormed (Map.make xs)
 := by
-  simp only [WellFormed, make, toList, kvs, List.canonicalize_idempotent]
+  simp [WellFormed, make, toList, kvs, List.canonicalize_idempotent]
 
-theorem make_mem_list_mem [LT α] [StrictLT α] [DecidableLT α] {xs : List (α × β)} :
```
emina, line 59: Let's leave the old signatures here.  

If a new theorem is necessary or helpful, it can be added as well, and it can use the old one in the proof.

cedar-lean/Cedar/Thm/Data/Map.lean
```diff
@@ -54,76 +169,274 @@ theorem not_contains_of_empty {α β} [BEq α] (k : α) :
...
+-/
+theorem make_cons [LT α] [DecidableLT α] {xs ys : List (α × β)} {ab : α × β} :
+  make xs = make ys → make (ab :: xs) = make (ab :: ys)
+:= by
+  simp only [make, mk.injEq]
+  apply List.canonicalize_cons
+
 /-! ### find? and mapOnValues -/
 
-theorem find?_mapOnValues {α β γ} [LT α] [DecidableLT α] [DecidableEq α] (f : β → γ) (m : Map α β) (k : α)  :
```
emina, line 74: Let's not change this signature.  Again, it will break internal proofs. 

Independently of that, if there are no preconditions to constrain the implicit variables (as in this case), they are usually made explicit.

cedar-lean/Cedar/Thm/Data/Map.lean
```diff
@@ -54,76 +169,274 @@ theorem not_contains_of_empty {α β} [BEq α] (k : α) :
...
+theorem find?_mapOnValues {α β γ} [LT α] [DecidableLT α] [DecidableEq α] {f : β → γ} {m : Map α β} {k : α}  :
   (m.find? k).map f = (m.mapOnValues f).find? k
 := by
-  simp [Map.find?, Map.mapOnValues, Map.kvs, ←List.find?_pair_map]
-  cases __ : List.find? (fun x => x.fst == k) m.1 <;>
-  simp only [Option.map_none', Option.map_some']
+  simp only [find?, kvs, mapOnValues, ← List.find?_pair_map]
+  cases m.1.find? (λ x => x.fst == k) <;> simp
 
-theorem find?_mapOnValues_some {α β γ} [LT α] [DecidableLT α] [DecidableEq α] (f : β → γ) {m : Map α β} {k : α} {v : β} :
```
emina, line 81: Same comment as above:  let's not make breaking changes to the shared modules unless absolutely necessary.

cedar-lean/Cedar/Thm/Data/Map.lean
```diff
@@ -54,76 +169,274 @@ theorem not_contains_of_empty {α β} [BEq α] (k : α) :
...
+theorem find?_mapOnValues_some {α β γ} [LT α] [DecidableLT α] [DecidableEq α] {f : β → γ} {m : Map α β} {k : α} {v : β} :
   m.find? k = .some v →
   (m.mapOnValues f).find? k = .some (f v)
 := by
   intro h₁
-  rw [← find?_mapOnValues f m k]
+  rw [← find?_mapOnValues]
   simp [Option.map, h₁]
 
-theorem find?_mapOnValues_none {α β γ} [LT α] [DecidableLT α] [DecidableEq α] (f : β → γ) {m : Map α β} {k : α} :
```
emina, line 89: This is a breaking change. Please revert.

cedar-lean/Cedar/Thm/Data/Map.lean
```diff
@@ -54,76 +169,274 @@ theorem not_contains_of_empty {α β} [BEq α] (k : α) :
...
+theorem find?_mapOnValues_none {α β γ} [LT α] [DecidableLT α] [DecidableEq α] {f : β → γ} {m : Map α β} {k : α} :
   m.find? k = .none →
   (m.mapOnValues f).find? k = .none
 := by
   intro h₁
-  rw [← find?_mapOnValues f m k]
+  rw [← find?_mapOnValues]
   simp [Option.map, h₁]
 
-theorem mapOnValues_eq_make_map {α β γ} [LT α] [StrictLT α] [DecidableLT α] (f : β → γ) {m : Map α β}
```
emina, line 97: This is a breaking change.

cedar-lean/Cedar/Thm/Data/Map.lean
emina: It breaks a large internal development.  Let's keep the old definitions.  

As a general rule, we want to keep the shared files as stable as possible, and not make breaking changes unless really necessary.

We can do a cleanup pass once our proofs are stable :)

cedar-lean/Cedar/Thm/Data/Map.lean
```diff
@@ -54,76 +169,274 @@ theorem not_contains_of_empty {α β} [BEq α] (k : α) :
...
-:= by
-  simp [Map.toList, Map.kvs, Map.find?] at *
-  split at h₁ <;> simp at h₁
-  subst h₁
-  rename_i h₂
-  have h₃ := List.find?_some h₂
-  simp at h₃ ; subst h₃
-  exact List.mem_of_find?_eq_some h₂
-
-theorem mapOnValues_contains {α β γ} [LT α] [DecidableLT α] [DecidableEq α] (f : β → γ) {m : Map α β} {k : α} :
```
emina, line 119: This is a breaking change.

cedar-lean/Cedar/Thm/Data/Map.lean
```diff
@@ -54,76 +169,274 @@ theorem not_contains_of_empty {α β} [BEq α] (k : α) :
...
+  (m₁.mapMOnValues f = some m₂) →
+  m₂.WellFormed
+:= by
+  unfold WellFormed toList
+  intro wf h₁
+  simp only [mapMOnValues, Option.pure_def, Option.bind_eq_bind, Option.bind_eq_some,
+    Option.some.injEq] at h₁
+  replace ⟨xs, h₁, h₂⟩ := h₁
+  subst h₂
+  sorry
```
emina, line 336: I think this would also benefit from some List lemmas I proved internally (and would have sworn I pushed here, but clearly not!):

```
theorem mapM'_some_eq_filterMap {α β} {f : α → Option β} {xs : List α} {ys : List β} :
  List.mapM' f xs = .some ys →
  List.filterMap f xs = ys
:= by
```
It should be easier to prove with this.  You turn WellFormed into SortedBy on lists, and if the initial list was sortedBy, then its filterMap is too, which then gives the well-formedness of the result.

cedar-lean/Cedar/Thm/Partial/Evaluation.lean
```diff
@@ -0,0 +1,113 @@
...
+    have ih : ∀ x ∈ xs, Partial.evaluate x request entities = (Spec.evaluate x request entities).map Partial.Value.value := by
+      intro x h₁
+      apply @on_concrete_eqv_concrete_eval x request entities wf
+    exact Set.on_concrete_eqv_concrete_eval ih
+  case record attrs =>
+    -- rw [List.map_attach₂ (λ x => (x.fst, Spec.Expr.asPartialExpr x.snd))]
+    have ih : ∀ kv ∈ attrs, Partial.evaluate kv.snd request entities = (Spec.evaluate kv.snd request entities).map Partial.Value.value := by
+      intro kv h₁
+      have h₂ : sizeOf kv.snd <= sizeOf kv := by simp only [sizeOf, Prod._sizeOf_1] ; omega
+      have : sizeOf kv.snd < sizeOf (Spec.Expr.record attrs) := by sorry
```
emina, line 88: I think you need an explicit `terminationBy` and then `decreasingBy` proof for this theorem.  

This will likely trigger two proof obligations:  one for sets and one for records.  But you'll be able to push it through.  I've seen this before too, where the addition of the missing proof obligation to the proof context is not enough.  The explicit proof will work in those cases.

emina (changes requested):

cedar-lean/Cedar/Thm/Data/Map.lean
```diff
@@ -27,20 +27,135 @@ This file proves useful properties of canonical list-based maps defined in
...
+    exact make_eqv h₁
+  case mpr =>
+    intro h₁
+    subst h₁
+    exact List.Equiv.refl
+
+/-! ### contains, mem, and find? -/
 
-theorem in_list_in_map {α : Type u} (k : α) (v : β) (m : Map α β) :
+theorem in_list_in_map {α : Type u} {k : α} {v : β} {m : Map α β} :
```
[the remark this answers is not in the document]
cdisselkoen: Emina, I thought you would (like me) be bugged by how inconsistent `Thm/Data/Map.lean` currently is about whether these `k`, `v`, and `m` arguments are `{}` or not.  (Compare to `contains_iff_some_find?`, immediately below this theorem.)

Adapting to this change is extremely trivial.  In most cases you can just remove the explicit `k`, `v`, and `m` arguments from callsites, because Lean can infer these arguments from the hypothesis argument.  If not, you can just add `@` and it will work exactly the same as before.

I think the consistency is worth the minor pain of updating usage sites of this theorem.

cedar-lean/Cedar/Thm/Data/Map.lean
cdisselkoen: I'll revert.

cdisselkoen: Reverted and revised this PR

cedar-lean/Cedar/Thm/Data/Map.lean
```diff
@@ -27,20 +27,135 @@ This file proves useful properties of canonical list-based maps defined in
...
+/-- kinda the converse of `in_kvs_in_values` -/
+theorem in_values_exists_key {m : Map α β} {v : β} :
+  v ∈ m.values → ∃ k, (k, v) ∈ m.kvs
+:= by
+  simp only [values, List.mem_map, forall_exists_index, and_imp]
+  intro kv h₁ h₂
+  subst h₂
+  exists kv.fst
+
+theorem in_list_some_find? [DecidableEq α] [LT α] [DecidableLT α] {k : α} {v : β} {m : Map α β} :
```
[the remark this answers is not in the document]
cdisselkoen: Done

cedar-lean/Cedar/Thm/Data/Map.lean
cdisselkoen: Can you explain more why we would make the implicit variables explicit here?  It seems to be more ergonomic to leave them implicit.  Lean is able to infer them at all of the current usage sites for this theorem.

cedar-lean/Cedar/Thm/Data/Map.lean
cdisselkoen: See [above comment](https://github.com/cedar-policy/cedar-spec/pull/291#discussion_r1583122381).  This is a trivial breaking change and improves consistency.

cedar-lean/Cedar/Thm/Data/Map.lean
cdisselkoen: See [above comment](https://github.com/cedar-policy/cedar-spec/pull/291#discussion_r1583122381)

cedar-lean/Cedar/Thm/Data/Map.lean
cdisselkoen: Will revert, but I'd like to understand better how this is a breaking change.

I think the only case this would be breaking is if some usage site had `mapOnValues_eq_make_map (h1 := ...)` instead of just `mapOnValues_eq_make_map h1`?  The latter is much more concise and is not broken by this change.  Am I missing something?

cedar-lean/Cedar/Thm/Partial/Evaluation.lean
cdisselkoen: Thanks for the tip.  Termination proof is now passing.

cedar-lean/Cedar/Thm/Data/Map.lean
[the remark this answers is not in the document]
emina: Yeah, it's definitely bugging me, and we should fix it :).  

I'm hoping we can do that as part of a separate future "breaking changes" PR where all / most breaking changes we want to make are included.  (We talked about a few others offline---e.g., `toList` vs `kvs` and `elts` for sets.  I think you also also wanted to change how Map membership is defined, which is a good idea too.)

Handling them incrementally is more disruptive, at least on my end.

Should we create an issue on this repo to track all the changes we want to make and put a deadline on it?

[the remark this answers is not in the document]
cdisselkoen: Separate issue sounds good.  However, I think we should do it sooner than later, perhaps before merging this PR itself, so as not to accumulate more tech debt (more usage sites that might need changes).  I can split off these changes into a separate PR for discussion, hopefully today.

I disagree that we should have one big "breaking changes" PR including all / most breaking changes.  Focused PRs are much easier to review.  If you want to batch all the breaking changes on your end, you can delay applying the focused PRs until all of them are merged, and then handle them all at once.  But I'd imagine your job would be easier too if you handled one change at a time, rather than many unrelated changes concurrently.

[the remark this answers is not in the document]
cdisselkoen: In the meantime I'll revert these in favor of the separate PR

cedar-lean/Cedar/Thm/Data/Map.lean
cdisselkoen: But I will put this change in a separate PR, reverting it here

cedar-lean/Cedar/Thm/Data/Map.lean
cdisselkoen: Putting this in a separate PR and reverting it here

cedar-lean/Cedar/Thm/Data/Map.lean
cdisselkoen: Putting this in a separate PR and reverting it here

cedar-lean/Cedar/Thm/Data/Map.lean
cdisselkoen: Putting this in a separate PR and reverting it here

cedar-lean/Cedar/Thm/Data/Map.lean
cdisselkoen: Putting this in a separate PR and reverting it here

cedar-lean/Cedar/Thm/Data/Map.lean
emina: I think it's fine for the two theorems to have asymmetric names. But if we want symmetry, value operations / theorems can be wordier versions of the key operations' names, something like `make_mem_values_mem`.

cedar-lean/Cedar/Thm/Data/Map.lean
```diff
@@ -27,20 +27,135 @@ This file proves useful properties of canonical list-based maps defined in
...
+
+  (Organizationally belongs in the `make` section, but is needed before that)
+-/
+theorem make_eqv [LT α] [DecidableLT α] {xs ys : List (α × β)}:
+  xs ≡ ys →
+  Map.make xs = Map.make ys
+:= by
+  unfold make
+  intro h₁
+  sorry
```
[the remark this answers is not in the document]
cdisselkoen: The Map analogue of `Set.eq_means_eqv` is `eq_iff_kvs_equiv` (slightly below this), and proving that is actually why I need `make_eqv` in the first place.

I tried taking inspiration from `Set.eq_means_eqv` to make a (different, better) proof of `Map.eq_iff_kvs_equiv`.  Unfortunately, the `Set.eq_means_eqv` proof relies on `Set.make_of_make_eqv` which relies on `List.canonicalize_equiv` which explicitly does not apply for canonicalization of Maps, only Sets.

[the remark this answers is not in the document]
cdisselkoen: Nonetheless, I feel that `Map.make_eqv` should still be valid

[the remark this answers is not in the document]
emina: Yes, this should be valid.  If it's not urgent, I can take a look in a few days :)

cedar-lean/Cedar/Thm/Authorization/Authorizer.lean
cdisselkoen: Conversation moved to https://github.com/cedar-policy/cedar-spec/pull/305#discussion_r1587717976

cedar-lean/Cedar/Thm/Authorization/Authorizer.lean
cdisselkoen: Superseded by https://github.com/cedar-policy/cedar-spec/pull/305#discussion_r1587719350

cedar-lean/Cedar/Thm/Data/List.lean
```diff
@@ -114,70 +114,9 @@ theorem filterMap_equiv (f : α → Option β) (xs ys : List α) :
   apply And.intro <;>
   intro b a h₃ h₄ <;>
   exists a <;>
-  simp [h₄]
-  exact h₁ h₃
-  exact h₂ h₃
-
-theorem filterMap_empty_iff_f_returns_none {f : α → Option β} {xs : List α} :
-  xs.filterMap f = [] ↔ ∀ x ∈ xs, f x = none
```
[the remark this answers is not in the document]
cdisselkoen: Conversation moved to https://github.com/cedar-policy/cedar-spec/pull/305#discussion_r1587725177

cedar-lean/Cedar/Thm/Data/List.lean
cdisselkoen: Conversation moved to https://github.com/cedar-policy/cedar-spec/pull/305#discussion_r1587727061

cedar-lean/Cedar/Thm/Data/List.lean
cdisselkoen: Conversation moved to https://github.com/cedar-policy/cedar-spec/pull/305#discussion_r1587727756

cedar-lean/Cedar/Thm/Data/List.lean
cdisselkoen: Conversation moved to https://github.com/cedar-policy/cedar-spec/pull/305#discussion_r1587739841

cedar-lean/Cedar/Thm/Data/Map.lean
```diff
@@ -27,20 +27,135 @@ This file proves useful properties of canonical list-based maps defined in
...
+-/
+theorem key_maps_to_one_value [BEq α] [LT α] [DecidableLT α] (k : α) (v₁ v₂ : β) (m : Map α β) :
+  m.WellFormed →
+  (k, v₁) ∈ m.kvs →
+  (k, v₂) ∈ m.kvs →
+  v₁ = v₂
+:= by
+  unfold WellFormed toList
+  intro wf h₁ h₂
+  sorry
```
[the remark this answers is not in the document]
cdisselkoen: For posterity, @emina's PR she is referring to is #297.

#305 uses #297 to prove this lemma.

cedar-lean/Cedar/Thm/Data/Map.lean
```diff
@@ -27,20 +27,135 @@ This file proves useful properties of canonical list-based maps defined in
...
+-/
+theorem equal_keys_same_value [BEq α] [LT α] [DecidableLT α] (k₁ k₂ : α) (v : β) (m : Map α β) :
+  m.WellFormed →
+  (k₁, v) ∈ m.kvs →
+  k₁ == k₂ →
+  (k₂, v) ∈ m.kvs
+:= by
+  unfold WellFormed toList
+  intro wf h₁ h₂
+  sorry
```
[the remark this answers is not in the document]
cdisselkoen: #305 uses #297 to prove this lemma.

cedar-lean/Cedar/Thm/Data/Map.lean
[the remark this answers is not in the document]
cdisselkoen: Conversation continues at https://github.com/cedar-policy/cedar-spec/pull/305#discussion_r1587742127

cedar-lean/Cedar/Thm/Data/Map.lean
cdisselkoen: #305 has a proof that does use `wf_iff_sorted`, but doesn't directly follow the above advice haha.  I should probably have reread this comment more carefully.  Let me try following this advice more directly

cdisselkoen: Yep, this advice does lead to a good proof.  I added it as an "alternate proof" in #305.  I'm not sure which proof is better, this or #305's original proof.  #305's original proof is technically even simpler, because it relies on `mapMOnValues_preserves_keys` which does most of the work (and has a more complicated proof).  In contrast, the new alt proof following the above advice relies on `List.mapM_some_eq_filterMap` (and `List.filterMap_sortedBy`) to do most of the work.

Let's continue this conversation at https://github.com/cedar-policy/cedar-spec/pull/305#discussion_r1587795641

cdisselkoen:
After a long hiatus while we discussed #297 and #305, I've rebased this PR now that those two are merged.

cdisselkoen:
CI is passing and there are no `sorry`s left.  Ready for re-review.

cedar-lean/Cedar/Thm/Partial/Evaluation/GetAttr.lean
```diff
@@ -0,0 +1,126 @@
...
+  if `entities.attrs uid` is `ok` with some attrs, those attrs are a
+  well-formed `Map`
+-/
+theorem partialEntities_attrs_wf {entities : Partial.Entities} {uid : EntityUID} {attrs: Map String Partial.Value} :
+  entities.AllWellFormed →
+  entities.attrs uid = .ok attrs →
+  attrs.WellFormed
+:= by
+  unfold Partial.Entities.attrs Partial.Entities.AllWellFormed Partial.EntityData.WellFormed
+  intro wf h₁
```
emina, line 40: Minor style comment---definitely no need to change this anytime soon:  hypotheses names start with `h` in all the Lean code I've looked at (e.g., Std).  It's unusual to see one that doesn't, except for `ih`.  I tend to write `hwf` for the well-formedness ones.

emina (approved):

cedar-lean/Cedar/Thm/Partial/Evaluation/GetAttr.lean
```diff
@@ -0,0 +1,126 @@
...
+    apply (wf_edata _ _).left
+    have h₃ := Map.in_values_iff_findOrErr_ok (v := attrs) (e := Error.entityDoesNotExist) wf_m
+    simp only [h₃]
+    exists uid
+
+/--
+  if `Partial.attrsOf` returns `ok` with some attrs, those attrs are a
+  well-formed `Map`
+-/
+theorem attrsOf_wf {entities : Partial.Entities} {v : Spec.Value} {attrs : Map String Partial.Value} :
```
khieta, line 55: Maybe I didn't review carefully enough -- but it doesn't look like this is used anywhere? In general it seems like the "WellFormed" requirement is only used in the `Var` case?

khieta (approved):
Left some minor comments on style, which you can take or leave. PR looks good to me -- no `sorry`s and the structure is consistent with the proofs in `Thm/Validation`.

cedar-lean/Cedar/Thm/Partial/Evaluation.lean
```diff
@@ -0,0 +1,121 @@
...
+/--
+  Corollary to the above: partial evaluation with concrete inputs gives a
+  concrete value (or an error)
+-/
+theorem on_concrete_gives_concrete (expr : Spec.Expr) (request : Spec.Request) (entities : Spec.Entities)
+  (wf : request.WellFormed) :
+  match Partial.evaluate expr request entities with
+  | .ok (.value _) => true
+  | .ok (.residual _) => false
+  | .error _ => true
```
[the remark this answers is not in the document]
cdisselkoen: Pushed a version with this named `isValueOrError`. Lmk if it looks better to you

khieta (approved):

cedar-lean/Cedar/Thm/Partial/Evaluation.lean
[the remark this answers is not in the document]
khieta: Yeah, I prefer the change.

cedar-lean/Cedar/Thm/Partial/Evaluation.lean
```diff
@@ -0,0 +1,121 @@
...
+open Cedar.Partial (Unknown)
+open Cedar.Spec (Error Result)
+
+/--
+  Partial evaluation with concrete inputs gives the same output as
+  concrete evaluation with those inputs
+-/
+theorem on_concrete_eqv_concrete_eval (expr : Spec.Expr) (request : Spec.Request) (entities : Spec.Entities)
+  (wf : request.WellFormed) :
+  Partial.evaluate expr request entities = (Spec.evaluate expr request entities).map Partial.Value.value
```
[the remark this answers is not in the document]
cdisselkoen: Made this change and it's very nice.  Thanks for the suggestion

PR #301, opened 2024-05-01
Tweak build scripts to hopefully avoid commmon confusion

john-h-kastner-aws:
* Use `expect` instead of `unwrap` for better errors from `build.rs` when you forget to `source set-env-vars.sh`
* Move `lean-toolchain` symlink from `cedar-drt` to the repository root. With this change `lean --print-libdir` detects the correct toolchain when run in root or any subdirectory of the repository, hopefully avoiding the issue mentioned in the readme update here https://github.com/cedar-policy/cedar-spec/pull/277
* Add check to rebuild DRT if lean build dir has changed. Without this a changing and recompiling the lean did not reflect if cargo didn't have to recompile for some other reason.

### DIFF
--- a/cedar-drt/build.rs
+++ b/cedar-drt/build.rs
@@ -17,7 +17,7 @@
 use std::env;
 fn main() {
     let lean_dir = env::var("LEAN_LIB_DIR").expect(
-        "`LEAN_LIB_DIR` environment variable is not set! Try running `source set-env-vars.sh`",
+        "`LEAN_LIB_DIR` environment variable is not set! Try running `source set_env_vars.sh`",
     );
     println!("cargo:rustc-link-search=native=../cedar-lean/.lake/build/lib");
     println!("cargo:rustc-link-search=native={lean_dir}");
--- a/cedar-drt/build.rs
+++ b/cedar-drt/build.rs
@@ -22,4 +22,5 @@ fn main() {
     println!("cargo:rustc-link-search=native=../cedar-lean/.lake/build/lib");
     println!("cargo:rustc-link-search=native={lean_dir}");
     println!("cargo:rustc-link-search=native=../cedar-lean/.lake/packages/std/.lake/build/lib");
+    println!("cargo:rerun-if-changed==../cedar-lean/.lake/build/lib");
 }
--- a/cedar-drt/build.rs
+++ b/cedar-drt/build.rs
@@ -16,7 +16,9 @@
 
 use std::env;
 fn main() {
-    let lean_dir = env::var("LEAN_LIB_DIR").unwrap();
+    let lean_dir = env::var("LEAN_LIB_DIR").expect(
+        "`LEAN_LIB_DIR` environment variable is not set! Try running `source set-env-vars.sh`",
+    );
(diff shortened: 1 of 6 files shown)


cedar-drt/build.rs
```diff
@@ -16,7 +16,9 @@
 
 use std::env;
 fn main() {
-    let lean_dir = env::var("LEAN_LIB_DIR").unwrap();
+    let lean_dir = env::var("LEAN_LIB_DIR").expect(
+        "`LEAN_LIB_DIR` environment variable is not set! Try running `source set-env-vars.sh`",
```
cdisselkoen, line 20:
```suggestion
        "`LEAN_LIB_DIR` environment variable is not set! Try running `source set_env_vars.sh`",
```

cedar-drt/fuzz/build.rs
```diff
@@ -16,7 +16,9 @@
 
 use std::env;
 fn main() {
-    let lean_dir = env::var("LEAN_LIB_DIR").unwrap();
+    let lean_dir = env::var("LEAN_LIB_DIR").expect(
+        "`LEAN_LIB_DIR` environment variable is not set! Try running `source set-env-vars.sh`",
```
cdisselkoen, line 20:
```suggestion
        "`LEAN_LIB_DIR` environment variable is not set! Try running `source set_env_vars.sh`",
```

cdisselkoen (changes requested):

khieta (commented):
I'm sure the `expect` error message would have saved me much confusion 😄 

Can you also update the README build instructions to match the new location of `lean-toolchain`? Possibly moving the symlink means you'll no longer be able to run the setup script in the `cedar-drt` directory, which the instructions may suggest.

john-h-kastner-aws:
>  Possibly moving the symlink means you'll no longer be able to run the setup script in the `cedar-drt` directory, which the instructions may suggest.

The intent of this change is that `source set_env_vars.sh` should work from anywhere inside the `cedar-spec` repo.

cdisselkoen (approved):

john-h-kastner-aws:
That's what it looks like. `lean --print-libdir` prints a directory corresponding to the toolchain as long as a parent directory has the toolchain file.

khieta:
Ack. Then I'll update my request to reverting #277

khieta (approved):

cedar-drt/build.rs
```diff
@@ -16,8 +16,11 @@
...
 use std::env;
 fn main() {
-    let lean_dir = env::var("LEAN_LIB_DIR").unwrap();
+    let lean_dir = env::var("LEAN_LIB_DIR").expect(
+        "`LEAN_LIB_DIR` environment variable is not set! Try running `source set_env_vars.sh`",
+    );
     println!("cargo:rustc-link-search=native=../cedar-lean/.lake/build/lib");
     println!("cargo:rustc-link-search=native={lean_dir}");
     println!("cargo:rustc-link-search=native=../cedar-lean/.lake/packages/std/.lake/build/lib");
+    println!("cargo:rerun-if-changed==../cedar-lean/.lake/build/lib");
```
khieta, line 25: This is much better than my solution of just always running `cargo clean` 😅

PR #297, opened 2024-04-26
Add more data structure lemmas.

emina:
This PR adds more useful data structure lemmas.

@cdisselkoen I think the following will be useful in discharging some of the `Map` proofs from #291:

* `List.filterMap_sortedBy`
* `List.mapM'_some_eq_filterMap`
* `List.mem_of_sortedBy_unique`

PR #305, opened 2024-05-02
more data structure lemmas

cdisselkoen:
A bunch more data structure lemmas.  These changes were split off #291, but without most of the breaking changes to the Cedar/Thm/Data files, and also taking advantage of the new lemmas in #297 when appropriate.  There are still a few breaking changes in this PR, but only to lemmas introduced in #297, as a compromise due to the simultaneous/duplicated work in #291 and #297.  Also, the changes are isolated to this PR rather than being included with #291.

### DIFF
--- a/cedar-lean/Cedar/Thm/Authorization/Authorizer.lean
+++ b/cedar-lean/Cedar/Thm/Authorization/Authorizer.lean
@@ -145,13 +145,12 @@ theorem sound_policy_slice_is_equierror (request : Request) (entities : Entities
   simp [List.mem_filter] <;>
   intro h₄ h₅ <;>
   apply And.intro
-  case left.left => exact h₁ h₄
-  case left.right => exact h₅
-  case right.left =>
-    by_contra h₆
+  · exact h₁ h₄
+  · exact h₅
+  · by_contra h₆
     specialize h₂ policy h₄ h₆
     exact h₂.right h₅
-  case right.right => exact h₅
+  · exact h₅
 
 /--
   an alternate, proved-equivalent, definition of errorPolicies that's easier to prove things about
@@ -170,24 +169,21 @@ theorem alternate_errorPolicies_equiv_errorPolicies {policies : Policies} {reque
     intro pid p h₁ h₂
     exists p
     unfold errored hasError at h₂
-    split at h₂
-    case inl h₃ =>
-      unfold hasError
+    split at h₂ <;> rename_i h₃
+    · unfold hasError
       apply And.intro
-      case left => simp [h₃, h₁, List.mem_filter]
-      case right => simp at h₂; exact h₂
-    case inr => contradiction
+      · simp [h₃, h₁, List.mem_filter]
+      · simp at h₂; exact h₂
+    · contradiction
   case right =>
     intro p h₁
     exists p
     simp [List.mem_filter] at h₁
-    apply And.intro
-    case left => exact h₁.left
-    case right =>
-      unfold errored
-      split
-      case inl => rfl
-      case inr h₃ => simp [h₃] at h₁
+    apply And.intro h₁.left
+    unfold errored
+    split <;> rename_i h₃
+    · rfl
+    · simp [h₃] at h₁
 
 theorem errorPolicies_eq_for_sound_policy_slice (request : Request) (entities : Entities) (slice policies : Policies) :
   IsSoundPolicySlice request entities slice policies →
@@ -357,12 +353,10 @@ theorem mapOrErr_value_asEntityUID_on_uids_produces_set {list : List EntityUID} 
     repeat rw [List.subset_def] at *
     constructor <;> intro a h₃ <;>
     replace h₃ := List.mem_map_of_mem (Value.prim ∘ Prim.entityUID) h₃
-    case left =>
-      specialize h₁ h₃
+    · specialize h₁ h₃
       simp at h₁
       exact h₁
-    case right =>
-      specialize h₂ h₃
+    · specialize h₂ h₃
       simp at h₂
       exact h₂
   case h_2 err h =>
@@ -459,8 +453,8 @@ theorem produces_boolean_means_not_non_boolean {e : Expr} {request : Request} {e
   unfold producesNonBool at h₂
   generalize (evaluate e request entities) = res at h₁ h₂
   split at h₁
-  case h_1 => simp at h₂
-  case h_2 => split at h₂ <;> simp at h₁
+  · simp at h₂
+  · split at h₂ <;> simp at h₁
 
 theorem principal_scope_does_not_throw {policy : Policy} {request : Request} {entities : Entities} {err : Error} :
   ¬ (evaluate policy.principalScope.toExpr request entities = .error err)
--- a/cedar-lean/Cedar/Thm/Authorization/Authorizer.lean
+++ b/cedar-lean/Cedar/Thm/Authorization/Authorizer.lean
@@ -270,16 +270,6 @@ theorem satisfied_implies_resource_scope {policy : Policy} {request : Request} {
   exact resource_eval_ok_means_resource_in_uid h₁ h₂
 
 /--
-  A generic lemma that relates List.mapM to List.map. Not in Std AFAICT.
--/
-theorem if_f_produces_pure_then_mapM_f_is_pure_map {α β} [Monad m] [LawfulMonad m] {f : α → β} {list : List α} :
-  list.mapM ((fun a => pure (f a)) : α → m β) = pure (list.map f)
-:= by
-  induction list
-  case nil => simp
-  case cons x xs h => simp [h]
-
-/--
   A generic lemma about composing List.mapM with List.map. Not in Std AFAICT.
 -/
 theorem mapM_over_map {α β γ} [Monad m] [LawfulMonad m] {f : α → β} {g : β → m γ} {list : List α} :
@@ -293,8 +283,8 @@ theorem mapM_evaluate_uids_produces_uids {list : List EntityUID} {request : Requ
   List.mapM (evaluate · request entities) (list.map fun uid => Expr.lit (.entityUID uid)) = .ok (list.map (Value.prim ∘ Prim.entityUID))
 := by
   rw [mapM_over_map]
-  simp [evaluate]
-  apply if_f_produces_pure_then_mapM_f_is_pure_map
+  unfold evaluate
+  exact List.mapM_pure
 
 theorem asEntityUID_of_uid {uid : EntityUID} :
   Value.asEntityUID (.prim (.entityUID uid)) = .ok uid
@@ -310,80 +300,56 @@ theorem mapM_asEntityUID_of_uid {uids : List EntityUID} :
   case nil =>
     rw [List.map_nil, List.mapM_nil]
     simp [pure, Except.pure]
-  case cons x xs h_ind =>
+  case cons x xs ih =>
     rw [List.map_cons, List.mapM_cons]
-    simp [pure, Except.pure, asEntityUID_of_uid, h_ind]
+    simp [pure, Except.pure, asEntityUID_of_uid, ih]
 
 /--
-  A generic lemma about the behavior of List.mapM' in the Except monad
+  Std has `Option.isSome_iff_exists`, but not this analogue for `Except`
 -/
-theorem mapM'_ok_iff_f_ok_on_all_elements {f : α → Except ε β} {list : List α} :
-  Except.isOk (list.mapM' f) ↔ ∀ x ∈ list, Except.isOk (f x)
+theorem Except.isOk_iff_exists {x : Except ε α} :
+  Except.isOk x ↔ ∃ a, x = .ok a
 := by
-  simp [Except.isOk, Except.toBool]
-  constructor
-  case mp =>
-    induction list
-    case nil =>
-      intro _ x h₂
-      simp at h₂
-    case cons y ys h_ind =>
-      intro h₁ x h₂
-      unfold List.mapM' at h₁
-      cases h₄ : (f y) <;> simp [h₄] at h₁
-      case ok b =>
-        rcases (List.mem_cons.mp h₂) with h₅ | h₅
-        case inl => rw [← h₅] at h₄; simp [h₄]
-        case inr =>
-          apply h_ind; clear h_ind
-          case a =>
-            split at h₁ <;> split <;> simp
-            case h_1.h_2 h₅ _ _ h₆ => simp [h₆] at h₅
-            case h_2.h_2 => simp at h₁
-          case a => exact h₅
-  case mpr =>
-    induction list
-    case nil => simp [List.mapM', pure, Except.pure]
-    case cons x xs h_ind =>
-      intro h₂
-      split <;> simp
-      case h_2 err h₃ =>
-        cases h₄ : (f x) <;> simp [h₄] at h₂
-        case ok b =>
-          specialize h_ind h₂
-          split at h_ind <;> simp at h_ind
-          case h_1 err h₆ =>
-            simp [h₄, h₆, List.mapM', pure, Except.pure] at h₃
+  cases x <;> simp [Except.isOk, Except.toBool]
 
-theorem if_mapM'_doesn't_fail_on_list_then_doesn't_fail_on_set [LT α] [DecidableLT α] [StrictLT α] {f : α → Except ε β} {list : List α} :
-  Except.isOk (list.mapM' f) →
-  Except.isOk ((Set.elts (Set.make list)).mapM' f)
+theorem if_mapM_doesn't_fail_on_list_then_doesn't_fail_on_set [LT α] [DecidableLT α] [StrictLT α] {f : α → Except ε β} {as : List α} :
+  Except.isOk (as.mapM f) →
+  Except.isOk ((Set.elts (Set.make as)).mapM f)
 := by
-  simp [mapM'_ok_iff_f_ok_on_all_elements]
-  intro h₁ y h₂
-  apply h₁ y; clear h₁
-  rw [Set.make_mem]
-  rw [← Set.in_list_iff_in_set]
-  exact h₂
+  intro h₁
+  replace ⟨bs, h₁⟩ := Except.isOk_iff_exists.mp h₁
+  replace h₁ := List.mapM_ok_implies_all_ok h₁
+  cases as <;> simp at h₁
+  case nil => simp [Set.elts_make_nil, pure, Except.pure, Except.isOk, Except.toBool]
+  case cons ahd atl =>
+    replace ⟨⟨b, _, h₁⟩, h₂⟩ := h₁
+    apply Except.isOk_iff_exists.mpr
+    apply List.all_ok_implies_mapM_ok
+    intro a h₃
+    rw [Set.in_list_iff_in_set] at h₃
+    rw [← Set.make_mem] at h₃
+    rcases List.mem_cons.mp h₃ with h₃ | h₃
+    case a.inl => subst h₃ ; exists b
+    case a.inr =>
+      replace ⟨b, _, h₄⟩ := h₂ a h₃
+      exists b
 
-theorem mapM'_asEntityUID_on_set_uids_produces_ok {uids : List EntityUID} :
-  Except.isOk (List.mapM' Value.asEntityUID (Set.elts (Set.make (uids.map (Value.prim ∘ Prim.entityUID)))))
+theorem mapM_asEntityUID_on_set_uids_produces_ok {uids : List EntityUID} :
+  Except.isOk (List.mapM Value.asEntityUID (Set.elts (Set.make (uids.map (Value.prim ∘ Prim.entityUID)))))
 := by
-  apply if_mapM'_doesn't_fail_on_list_then_doesn't_fail_on_set
+  apply if_mapM_doesn't_fail_on_list_then_doesn't_fail_on_set
   unfold Except.isOk Except.toBool
   split <;> simp
-  case a.h_2 err h =>
-    rw [List.mapM'_eq_mapM] at h
-    simp [mapM_asEntityUID_of_uid] at h
+  case a.h_2 e h => simp [mapM_asEntityUID_of_uid] at h
 
 theorem mapOrErr_value_asEntityUID_on_uids_produces_set {list : List EntityUID} {err : Error} :
   Set.mapOrErr Value.asEntityUID (Set.make (list.map (Value.prim ∘ Prim.entityUID))) err = .ok (Set.make list)
 := by
   unfold Set.mapOrErr
-  rw [←List.mapM'_eq_mapM]
   split <;> simp
   case h_1 list' h =>
     -- in this case, mapping Value.asEntityUID over the set returns .ok
+    rw [← List.mapM'_eq_mapM] at h
     replace h := mapM'_asEntityUID_eq_entities h
     have ⟨h₁, h₂⟩ := Set.elts_make_is_id_then_equiv h; clear h
     rw [Set.make_make_eqv]
@@ -401,7 +367,7 @@ theorem mapOrErr_value_asEntityUID_on_uids_produces_set {list : List EntityUID} 
       exact h₂
   case h_2 err h =>
     -- in this case, mapping Value.asEntityUID over the set returns .error
-    have h₁ := @mapM'_asEntityUID_on_set_uids_produces_ok list
+    have h₁ := @mapM_asEntityUID_on_set_uids_produces_ok list
     simp [h, Except.isOk, Except.toBool] at h₁
 
 theorem action_in_set_of_euids_produces_boolean {list : List EntityUID} {request : Request} {entities : Entities} :
@@ -425,7 +391,8 @@ theorem principal_scope_produces_boolean {policy : Policy} {request : Request} {
   cases policy.principalScope.1 <;>
   simp [evaluate, Var.eqEntityUID, Var.inEntityUID, Var.isEntityType, apply₁, apply₂]
   case isMem ety uid =>
-    simp [Result.as, Lean.Internal.coeM, Coe.coe, Value.asBool, pure, Except.pure, CoeT.coe, CoeHTCT.coe, CoeHTC.coe, CoeOTC.coe, CoeTC.coe]
+    simp only [Result.as, Coe.coe, Value.asBool, pure, Except.pure, Except.bind_ok,
+      beq_eq_false_iff_ne, ne_eq, ite_not]
     generalize (inₑ request.principal uid entities) = b₁
     generalize (ety == request.principal.ty) = b₂
     split
@@ -454,7 +421,8 @@ theorem action_scope_produces_boolean {policy : Policy} {request : Request} {ent
     simp [evaluate, Var.eqEntityUID, Var.inEntityUID, Var.isEntityType, apply₁, apply₂]
     cases scope <;> simp [evaluate, apply₁, apply₂, Result.as]
     case isMem ety uid =>
-      simp [Result.as, Lean.Internal.coeM, Coe.coe, Value.asBool, pure, Except.pure, CoeT.coe, CoeHTCT.coe, CoeHTC.coe, CoeOTC.coe, CoeTC.coe]
+      simp only [Coe.coe, Value.asBool, pure, Except.pure, Except.bind_ok, beq_eq_false_iff_ne,
+        ne_eq, ite_not]
       generalize (inₑ request.action uid entities) = b₁
       generalize (ety == request.action.ty) = b₂
       split
@@ -471,7 +439,8 @@ theorem resource_scope_produces_boolean {policy : Policy} {request : Request} {e
   cases policy.resourceScope.1 <;>
   simp [evaluate, Var.eqEntityUID, Var.inEntityUID, Var.isEntityType, apply₁, apply₂]
   case isMem ety uid =>
-    simp [Result.as, Lean.Internal.coeM, Coe.coe, Value.asBool, pure, Except.pure, CoeT.coe, CoeHTCT.coe, CoeHTC.coe, CoeOTC.coe, CoeTC.coe]
+    simp only [Result.as, Coe.coe, Value.asBool, pure, Except.pure, Except.bind_ok,
+      beq_eq_false_iff_ne, ne_eq, ite_not]
     generalize (inₑ request.resource uid entities) = b₁
     generalize (ety == request.resource.ty) = b₂
     split
--- a/cedar-lean/Cedar/Thm/Authorization/Evaluator.lean
+++ b/cedar-lean/Cedar/Thm/Authorization/Evaluator.lean
@@ -57,16 +57,15 @@ theorem and_true_implies_right_true {e₁ e₂ : Expr} {request : Request} {enti
   simp [evaluate, h₂, Result.as, Coe.coe, Value.asBool] at h₁
   generalize h₃ : (evaluate e₂ request entities) = r₂
   simp [h₃] at h₁
-  cases r₂ <;> simp [Lean.Internal.coeM] at h₁
+  cases r₂ <;> simp only [Except.bind_err] at h₁
   case ok v₂ =>
-    cases v₂ <;> try simp at h₁
+    cases v₂ <;> try simp only [Except.bind_err] at h₁
     case _ p₂ =>
-      cases p₂ <;> simp at h₁
+      cases p₂ <;> simp only [Except.bind_ok, Except.bind_err] at h₁
       case _ b =>
         cases b
         case false =>
-          simp only [pure, Except.pure, CoeT.coe, CoeHTCT.coe, CoeHTC.coe, CoeOTC.coe,
-            CoeTC.coe, Coe.coe, Except.ok.injEq, Value.prim.injEq, Prim.bool.injEq] at h₁
+          simp only [pure, Except.pure, Except.ok.injEq, Value.prim.injEq, Prim.bool.injEq] at h₁
         case true => rfl
 
 /- some shorthand to make things easier to read and write -/
@@ -118,14 +117,14 @@ theorem ways_and_can_error {e₁ e₂ : Expr} {request : Request} {entities : En
           cases h_e₂ : (evaluate e₂ request entities) with
           | ok val =>
             cases val <;>
-            simp [h_e₂, evaluate, Result.as, Coe.coe, Value.asBool, Lean.Internal.coeM, pure, Except.pure] at h₁ <;>
+            simp [h_e₂, evaluate, Result.as, Value.asBool, pure, Except.pure] at h₁ <;>
             simp [h₁]
             case prim prim =>
               cases prim <;>
               simp [h_e₂] at h₁ <;>
               simp [h₁]
           | error e =>
-            simp [h_e₂, Lean.Internal.coeM] at h₁
+            simp only [h_e₂, ↓reduceIte, Except.bind_err, Except.error.injEq] at h₁
             simp [h₁]
         | false => simp [h_e₁] at h₁
   case error e =>
@@ -143,10 +142,10 @@ theorem and_produces_bool_or_error {e₁ e₂ : Expr} {request : Request} {entit
 := by
   cases h : evaluate (Expr.and e₁ e₂) request entities <;> simp
   case ok val =>
-    cases val <;> simp only [evaluate, Result.as, Coe.coe, Value.asBool, Bool.not_eq_true',
-      Lean.Internal.coeM, pure, Except.pure] at h <;>
-    generalize (evaluate e₁ request entities) = r₁ at h <;>
-    generalize (evaluate e₂ request entities) = r₂ at h
+    cases val
+    <;> simp only [evaluate, Result.as, Coe.coe, Value.asBool, Bool.not_eq_true', pure, Except.pure] at h
+    <;> generalize (evaluate e₁ request entities) = r₁ at h
+    <;> generalize (evaluate e₂ request entities) = r₂ at h
     case prim prim =>
       cases prim <;> simp
       case int | string | entityUID =>
--- a/cedar-lean/Cedar/Thm/Data/List.lean
+++ b/cedar-lean/Cedar/Thm/Data/List.lean
@@ -41,7 +41,7 @@ infix:50 " ≡ " => Equiv
 
 theorem Equiv.refl {a : List α} :
   a ≡ a
-:= by unfold List.Equiv; simp
+:= by unfold List.Equiv; simp only [Subset.refl, and_self]
 
 theorem Equiv.symm {a b : List α} :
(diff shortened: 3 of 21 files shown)


cedar-lean/Cedar/Thm/Authorization/Authorizer.lean
```diff
@@ -269,16 +269,6 @@ theorem satisfied_implies_resource_scope {policy : Policy} {request : Request} {
         (and_true_implies_right_true h₁))
   exact resource_eval_ok_means_resource_in_uid h₁ h₂
 
-/--
-  A generic lemma that relates List.mapM to List.map. Not in Std AFAICT.
--/
-theorem if_f_produces_pure_then_mapM_f_is_pure_map {α β} [Monad m] [LawfulMonad m] {f : α → β} {list : List α} :
-  list.mapM ((fun a => pure (f a)) : α → m β) = pure (list.map f)
```
cdisselkoen, line 276: Moved to Thm/Data/List.lean and renamed `mapM_pure`

cedar-lean/Cedar/Thm/Authorization/Authorizer.lean
```diff
@@ -310,80 +300,56 @@ theorem mapM_asEntityUID_of_uid {uids : List EntityUID} :
...
     rw [List.map_cons, List.mapM_cons]
-    simp [pure, Except.pure, asEntityUID_of_uid, h_ind]
+    simp [pure, Except.pure, asEntityUID_of_uid, ih]
 
 /--
-  A generic lemma about the behavior of List.mapM' in the Except monad
+  Std has `Option.isSome_iff_exists`, but not this analogue for `Except`
 -/
-theorem mapM'_ok_iff_f_ok_on_all_elements {f : α → Except ε β} {list : List α} :
-  Except.isOk (list.mapM' f) ↔ ∀ x ∈ list, Except.isOk (f x)
```
cdisselkoen, line 321: No longer needed, as #297's `List.mapM'_ok_implies_all_ok` is stronger

cedar-lean/Cedar/Thm/Authorization/Authorizer.lean
```diff
@@ -310,80 +300,56 @@ theorem mapM_asEntityUID_of_uid {uids : List EntityUID} :
...
-        case ok b =>
-          specialize h_ind h₂
-          split at h_ind <;> simp at h_ind
-          case h_1 err h₆ =>
-            simp [h₄, h₆, List.mapM', pure, Except.pure] at h₃
-
-theorem if_mapM'_doesn't_fail_on_list_then_doesn't_fail_on_set [LT α] [DecidableLT α] [StrictLT α] {f : α → Except ε β} {list : List α} :
-  Except.isOk (list.mapM' f) →
-  Except.isOk ((Set.elts (Set.make list)).mapM' f)
-:= by
```
cdisselkoen, line 361: Reformulated in terms of `mapM` below

cedar-lean/Cedar/Thm/Authorization/Authorizer.lean
```diff
@@ -310,80 +300,56 @@ theorem mapM_asEntityUID_of_uid {uids : List EntityUID} :
...
-  simp [mapM'_ok_iff_f_ok_on_all_elements]
-  intro h₁ y h₂
-  apply h₁ y; clear h₁
-  rw [Set.make_mem]
-  rw [← Set.in_list_iff_in_set]
-  exact h₂
-
-theorem mapM'_asEntityUID_on_set_uids_produces_ok {uids : List EntityUID} :
-  Except.isOk (List.mapM' Value.asEntityUID (Set.elts (Set.make (uids.map (Value.prim ∘ Prim.entityUID)))))
```
cdisselkoen, line 370: Reformulated in terms of `mapM` below

cedar-lean/Cedar/Thm/Data/List.lean
```diff
@@ -114,70 +114,9 @@ theorem filterMap_equiv (f : α → Option β) (xs ys : List α) :
   apply And.intro <;>
   intro b a h₃ h₄ <;>
   exists a <;>
-  simp [h₄]
-  exact h₁ h₃
-  exact h₂ h₃
-
-theorem filterMap_empty_iff_f_returns_none {f : α → Option β} {xs : List α} :
-  xs.filterMap f = [] ↔ ∀ x ∈ xs, f x = none
```
cdisselkoen, line 122: Moved to a dedicated section for `filterMap`, and renamed `filterMap_empty_iff_all_none`.  @emina I don't think the internal development depends on this one; if I am wrong, I can undo the rename.

cedar-lean/Cedar/Thm/Data/List.lean
```diff
@@ -114,70 +114,9 @@ theorem filterMap_equiv (f : α → Option β) (xs ys : List α) :
...
-        apply h₁ a
-        exact List.mem_cons_of_mem x h₂
-      case h_2 b h₂ =>
-        exfalso
-        specialize h₁ x
-        simp at h₁
-        simp [h₁] at h₂
-
-theorem filterMap_nonempty_iff_exists_f_returns_some {f : α → Option β} {xs : List α} :
-  xs.filterMap f ≠ [] ↔ ∃ x ∈ xs, (f x).isSome
```
cdisselkoen, line 155: Moved to a dedicated section for `filterMap`, and renamed `filterMap_nonempty_iff_exists`.  @emina I don't think the internal development depends on this one; if I am wrong, I can undo the rename.

cedar-lean/Cedar/Thm/Data/List.lean
```diff
@@ -114,70 +114,9 @@ theorem filterMap_equiv (f : α → Option β) (xs ys : List α) :
...
-  case mpr =>
-    intro h₁ h₂
-    rw [filterMap_empty_iff_f_returns_none] at h₂
-    replace ⟨x, h₁, h₃⟩ := h₁
-    specialize h₂ x h₁
-    simp [h₂, Option.isSome] at h₃
-
-theorem f_implies_g_then_subset {f g : α → Option β} {xs : List α} :
-  (∀ a b, f a = some b → g a = some b) →
-  xs.filterMap f ⊆ xs.filterMap g
```
cdisselkoen, line 173: Moved to a dedicated section for `filterMap`

cedar-lean/Cedar/Thm/Data/List.lean
```diff
@@ -745,60 +730,149 @@ theorem any_of_mem {f : α → Bool} {x : α} {xs : List α}
...
+  : List.map (λ x : { a : α // p a } => f x.val) (List.pmap Subtype.mk as h)
+    =
+    List.map f as
 := by
-  simp [mapM₁, attach, mapM_pmap_subtype]
+  induction as <;> simp [*]
 
-theorem mapM_implies_nil {f : α → Except β γ} {as : List α}
-  (h₁ : List.mapM f as = Except.ok []) :
-  as = []
```
cdisselkoen, line 779: Not removed, just moved

cedar-lean/Cedar/Thm/Data/List.lean
```diff
@@ -745,60 +730,149 @@ theorem any_of_mem {f : α → Bool} {x : α} {xs : List α}
...
+    List.all as f
 := by
-  rw [←List.mapM'_eq_mapM]
   induction as <;> simp [*]
 
-theorem mapM₁_eq_mapM [Monad m] [LawfulMonad m]
-  (f : α → m β)
-  (as : List α) :
-  List.mapM₁ as (fun x : { x // x ∈ as } => f x.val) =
-  List.mapM f as
```
cdisselkoen, line 773: Not removed, just moved

cedar-lean/Cedar/Thm/Data/List.lean
```diff
@@ -841,73 +915,323 @@ theorem forall₂_implies_all_right {α β} {R : α → β → Prop} {xs : List
       exists x
       simp only [mem_cons, ih, or_true, and_self]
 
-/-! ### mapM' -/
+/-! ### mapM, mapM', and mapM₁ -/
 
-theorem mapM'_ok_implies_forall₂ {α β γ} {f : α → Except γ β} {xs : List α} {ys : List β} :
-  List.mapM' f xs = .ok ys →
```
cdisselkoen, line 847: I needed the converse as well, so I made this an iff and renamed it `mapM'_ok_iff_forall2`.  This lemma was introduced in #297.  I reformulated the proofs in #291 (and here) to rely on this lemma instead of the other lemmas in #291.

cedar-lean/Cedar/Thm/Data/List.lean
```diff
@@ -841,73 +915,323 @@ theorem forall₂_implies_all_right {α β} {R : α → β → Prop} {xs : List
...
   List.mapM' f xs = .ok ys →
   ∀ y ∈ ys, ∃ x ∈ xs, f x = .ok y
 := by
   intro h
-  exact forall₂_implies_all_right (mapM'_ok_implies_forall₂ h)
+  exact forall₂_implies_all_right (mapM'_ok_iff_forall₂.mp h)
 
-theorem mapM'_some_implies_forall₂ {α β} {f : α → Option β} {xs : List α} {ys : List β} :
-  List.mapM' f xs = .some ys →
-  List.Forall₂ (λ x y => f x = .some y) xs ys
```
cdisselkoen, line 884: I needed the converse as well, so I made this an iff and renamed it `mapM'_some_iff_forall2`. This lemma was introduced in https://github.com/cedar-policy/cedar-spec/pull/297. I reformulated the proofs in https://github.com/cedar-policy/cedar-spec/pull/291 (and here) to rely on this lemma instead of the other lemmas in https://github.com/cedar-policy/cedar-spec/pull/291.

cedar-lean/Cedar/Thm/Data/List.lean
```diff
@@ -1158,40 +1486,7 @@ theorem mem_of_sortedBy_unique {α β} [LT β] [StrictLT β] [DecidableLT β] [D
...
--/
-
-theorem map_pmap_subtype
-  {p : α → Prop}
-  (f : α → β)
-  (as : List α)
-  (h : ∀ a, a ∈ as → p a)
-  : List.map (fun x : { a : α // p a } => f x.val) (List.pmap Subtype.mk as h)
-    =
-    List.map f as
```
cdisselkoen, line 1174: Moved to the `map` section (instead of a dedicated section for "copied from Mathlib")

cedar-lean/Cedar/Thm/Data/List.lean
```diff
@@ -1158,40 +1486,7 @@ theorem mem_of_sortedBy_unique {α β} [LT β] [StrictLT β] [DecidableLT β] [D
...
-  induction as <;> simp [*]
-
-theorem all_pmap_subtype
-  {p : α → Prop}
-  (f : α → Bool)
-  (as : List α)
-  (h : ∀ a, a ∈ as → p a)
-  : List.all (List.pmap Subtype.mk as h) (fun x : { a : α // p a } => f x.val)
-    =
-    List.all as f
```
cdisselkoen, line 1185: Moved to the `all` section (instead of a dedicated section for "copied from Mathlib")

cedar-lean/Cedar/Thm/Data/List.lean
```diff
@@ -1158,40 +1486,7 @@ theorem mem_of_sortedBy_unique {α β} [LT β] [StrictLT β] [DecidableLT β] [D
...
-:= by
-  induction as <;> simp [*]
-
-theorem map_congr {f g : α → β} : ∀ {l : List α},
-  (∀ x ∈ l, f x = g x) → map f l = map g l
```
cdisselkoen, line 1190: Moved to the `map` section (instead of a dedicated section for "copied from Mathlib")

cedar-lean/Cedar/Thm/Data/Map.lean
```diff
@@ -47,7 +84,85 @@ theorem wf_iff_sorted {α β} [LT α] [DecidableLT α] [StrictLT α] {m : Map α
...
+  `make` on equivalent lists of (k,v) pairs produces equal maps
+
+  (Organizationally belongs in the `make` section, but is needed before that)
+-/
+theorem make_eqv [LT α] [DecidableLT α] {xs ys : List (α × β)}:
+  xs ≡ ys →
+  Map.make xs = Map.make ys
+:= by
+  intro h; unfold make; simp
+  sorry
```
cdisselkoen, line 118: The one `sorry` in this PR.  I will continue working on this, but help/advice is also appreciated.

cedar-lean/Cedar/Thm/Data/Map.lean
```diff
@@ -85,59 +224,128 @@ theorem make_mem_list_mem [LT α] [StrictLT α] [DecidableLT α] {xs : List (α
...
 := by
-  simp [Map.WellFormed] at h₁
-  simp [Map.mapOnValues, Map.toList, Map.kvs, Map.make] at *
-  rw [h₁] ; simp only ; rw [eq_comm]
-  have h₂ : Prod.map id f = (fun (x : α × β) => (x.fst, f x.snd)) := by unfold Prod.map ; simp only [id_eq]
-  simp only [← h₂, ← List.canonicalize_of_map_fst, List.canonicalize_idempotent]
-
-theorem find?_mem_toList {α β} [LT α] [DecidableLT α] [DecidableEq α] {m : Map α β} {k : α} {v : β}
-  (h₁ : m.find? k = .some v) :
-  (k, v) ∈ m.toList
```
cdisselkoen, line 132: Not removed and not broken, just moved farther up, as I needed it farther up

cedar-lean/Cedar/Thm/Data/Map.lean
```diff
@@ -164,31 +372,279 @@ theorem mapOnValues_contains {α β γ} [LT α] [DecidableLT α] [DecidableEq α
...
-/-! ### sizeOf -/
+theorem findOrErr_ok_iff_find?_some [LT α] [DecidableLT α] [DecidableEq α] {m : Map α β} {k : α} {v : β} {e : Error} :
+  m.findOrErr k e = .ok v ↔ m.find? k = some v
+:= by
+  unfold findOrErr
+  cases m.find? k <;> simp
 
-theorem sizeOf_lt_of_value [SizeOf α] [SizeOf β] {m : Map α β} {k : α} {v : β}
-  (h : (k, v) ∈ m.1) :
-  sizeOf v < sizeOf m
```
cdisselkoen, line 172: Moved the entire `sizeOf` section to the top instead of the bottom, as I needed some `sizeOf` facts elsewhere in this file.

cedar-lean/Cedar/Thm/Data/Map.lean
```diff
@@ -164,31 +372,303 @@ theorem mapOnValues_contains {α β γ} [LT α] [DecidableLT α] [DecidableEq α
...
+  intro wf h₁
+  have h₂ := mapMOnValues_preserves_keys h₁
+  exact (List.map_eq_implies_sortedBy h₂).mp wf
+
+/--
+  Alternate proof of `mapMOnValues_wf`, that relies on
+  `List.mapM_some_eq_filterMap` instead of `mapMOnValues_preserves_keys`. Which do
+  we prefer?
+-/
+theorem mapMOnValues_wf_alt_proof [LT α] [DecidableLT α] [StrictLT α] {f : β → Option γ} {m₁ : Map α β} {m₂ : Map α γ} :
```
cdisselkoen, line 475: Thread for discussion of the two proofs of `mapMOnValues_wf`

cedar-lean/Cedar/Thm/Data/Map.lean
```diff
@@ -164,31 +372,303 @@ theorem mapOnValues_contains {α β γ} [LT α] [DecidableLT α] [DecidableEq α
...
+  simp only at *
+  subst k'
+  exists v'
+
+/--
+  alternate proof of `mapMOnValues_some_implies_all_some`, which instead of
+  relying on `mapMOnValues_some_implies_forall₂`, relies on
+  `List.mapM_some_implies_all_some`.  Which do we prefer?
+-/
+theorem mapMOnValues_some_implies_all_some_alt_proof [LT α] [DecidableLT α] {f : β → Option γ} {m₁ : Map α β} {m₂ : Map α γ} :
```
cdisselkoen, line 567: Thread for discussion of the two proofs of `mapMOnValues_some_implies_all_some`

cedar-lean/Cedar/Thm/Authorization/Authorizer.lean
khieta: Actually I think this is the stronger version since it's a bijection rather than an implication. Still good to remove if not needed though.

cedar-lean/Cedar/Thm/Authorization/Evaluator.lean
```diff
@@ -118,14 +117,14 @@ theorem ways_and_can_error {e₁ e₂ : Expr} {request : Request} {entities : En
...
-            simp [h_e₂, evaluate, Result.as, Coe.coe, Value.asBool, Lean.Internal.coeM, pure, Except.pure] at h₁ <;>
+            simp [h_e₂, evaluate, Result.as, Value.asBool, pure, Except.pure] at h₁ <;>
             simp [h₁]
             case prim prim =>
               cases prim <;>
               simp [h_e₂] at h₁ <;>
               simp [h₁]
           | error e =>
-            simp [h_e₂, Lean.Internal.coeM] at h₁
+            simp only [h_e₂, ↓reduceIte, Except.bind_err, Except.error.injEq] at h₁
```
khieta, line 127: Lean noob question: is the downward arrow some special syntax?

cedar-lean/Cedar/Thm/Data/List.lean
```diff
@@ -1203,4 +1498,65 @@ theorem filterMap_congr {f g : α → Option β} : ∀ {l : List α},
...
+        intro a h₂
+        apply h₁ a
+        exact List.mem_cons_of_mem hd h₂
+      case h_2 b h₂ =>
+        exfalso
+        specialize h₁ hd
+        simp only [mem_cons, true_or, forall_const] at h₁
+        simp [h₁] at h₂
+
+theorem filterMap_nonempty_iff_exists {f : α → Option β} {xs : List α} :
```
khieta, line 1560: Nit: I feel like `*_exists_some` would be a little more pretty & symmetric with `*_all_none`

cedar-lean/Cedar/Thm/Data/Map.lean
cdisselkoen: I believe I have a proof that this lemma actually is _not valid_.  I discovered this by noticing that the `Set` equivalent (`Set.make_make_eqv`) relies on `List.equiv_implies_canonical_eq`, and working out that `List.equiv_implies_canonical_eq` _does not hold_ for `Prod.fst` (see comments on `List.equiv_implies_canonical_eq` added in this PR).

Specifically, consider
```
  xs = [(1, false), (1, true)]
  ys = [(1, true), (1, false)]
```
Even though `xs ≡ ys`, after `Map.make`, in one map, `1` will map to `false`, and in the other, `1` will map to `true`.

cdisselkoen: Now the question is, how many other lemmas that currently rely on `Map.make_eqv` can we save.

cedar-lean/Cedar/Thm/Data/Map.lean
```diff
@@ -85,59 +224,128 @@ theorem make_mem_list_mem [LT α] [StrictLT α] [DecidableLT α] {xs : List (α
...
+
 theorem make_nil_is_empty {α β} [LT α] [DecidableLT α] :
   (Map.make [] : Map α β) = Map.empty
 := by simp [make, empty, List.canonicalize_nil]
 
+/--
+  Note that the converse of this is not true:
+  counterexample `xs = [(1, false)]`, `ys = []`, `ab = (1, false)`.
+-/
+theorem make_cons [LT α] [DecidableLT α] {xs ys : List (α × β)} {ab : α × β} :
```
khieta, line 256: Maybe change to `make_cons_pair` to distinguish from `make_cons` in Set.lean?

cedar-lean/Cedar/Thm/Data/Map.lean
```diff
@@ -85,59 +224,128 @@ theorem make_mem_list_mem [LT α] [StrictLT α] [DecidableLT α] {xs : List (α
...
+  rename_i h₂
+  have h₃ := List.find?_some h₂
+  simp only [beq_iff_eq] at h₃ ; subst h₃
+  exact List.mem_of_find?_eq_some h₂
+
+/--
+  The `mpr` direction of this does not need the `wf` precondition and, in fact,
+  is available separately as `find?_mem_toList` above
+-/
+theorem in_list_some_find? [DecidableEq α] [LT α] [DecidableLT α] [StrictLT α] {k : α} {v : β} {m : Map α β}
```
khieta, line 280: Nit: `in_list_iff_find?_some`

cedar-lean/Cedar/Thm/Data/Map.lean
```diff
@@ -164,31 +372,303 @@ theorem mapOnValues_contains {α β γ} [LT α] [DecidableLT α] [DecidableEq α
...
+    replace ⟨(a', g), h₃, h₄⟩ := h₃ (a, b) h₂
+    simp only [Option.pure_def, Option.bind_eq_bind, Option.bind_eq_some, Option.some.injEq,
+      Prod.mk.injEq, exists_eq_right_right] at h₄
+    replace ⟨h₄, h₄'⟩ := h₄
+    subst a'
+    exists g
+
+theorem mapMOnValues_some_implies_all_from_some [LT α] [DecidableLT α] {f : β → Option γ} {m₁ : Map α β} {m₂ : Map α γ} :
+  m₁.mapMOnValues f = some m₂ →
+  ∀ kv ∈ m₂.kvs, ∃ v, (kv.fst, v) ∈ m₁.kvs ∧ f v = kv.snd
```
khieta, line 590: Pretty similar in form to `mapMOnValues_some_implies_all_some`. Do you really need both? Is there a way to combine them into one statement?

cedar-lean/Cedar/Thm/Data/Map.lean
```diff
@@ -164,31 +372,303 @@ theorem mapOnValues_contains {α β γ} [LT α] [DecidableLT α] [DecidableEq α
...
+    simp only
+    replace h₃ := List.mapM_some_implies_all_from_some h₃
+    replace ⟨(a', b), h₃, h₄⟩ := h₃ (a, g) h₂
+    simp only [Option.pure_def, Option.bind_eq_bind, Option.bind_eq_some, Option.some.injEq,
+      Prod.mk.injEq, exists_eq_right_right] at h₄
+    replace ⟨h₄, h₄'⟩ := h₄
+    subst a'
+    exists b
+
+theorem mapMOnValues_none_iff_exists {α : Type 0} [LT α] [DecidableLT α] {f : β → Option γ} {m : Map α β} :
```
khieta, line 626: Nit: `*_iff_exists_none`

cedar-lean/Cedar/Thm/Data/List.lean
```diff
@@ -841,73 +941,323 @@ theorem forall₂_implies_all_right {α β} {R : α → β → Prop} {xs : List
...
+  exact forall₂_implies_all_right (mapM'_some_iff_forall₂.mp h)
+
+theorem mapM_some_implies_all_from_some {α β} {f : α → Option β} {xs : List α} {ys : List β} :
+  List.mapM f xs = .some ys →
+  ∀ y ∈ ys, ∃ x ∈ xs, f x = .some y
+:= by
+  rw [← List.mapM'_eq_mapM]
+  exact mapM'_some_implies_all_from_some
+
+theorem mapM'_none_iff_exists {α β} {f : α → Option β} {xs : List α} :
```
khieta, line 1224: Nit: `*_iff_exists_none`

cedar-lean/Cedar/Thm/Data/List.lean
```diff
@@ -841,73 +941,323 @@ theorem forall₂_implies_all_right {α β} {R : α → β → Prop} {xs : List
...
+      simp only [mapM'_cons, Option.pure_def, Option.bind_eq_bind, Option.bind_eq_none]
+      intro yhd h₃ ytl h₄
+      rcases h₁ with h₁ | h₁
+      case inl => subst h₁ ; simp [h₂] at h₃
+      case inr =>
+        replace h₄ := mapM'_some_implies_all_some h₄
+        replace ⟨y, _, h₅⟩ := h₄ x h₁
+        simp [h₂] at h₅
+
+theorem mapM_none_iff_exists {α β} {f : α → Option β} {xs : List α} :
```
khieta, line 1256: Nit: `*_iff_exists_none`

cedar-lean/Cedar/Thm/Data/List.lean
```diff
@@ -841,73 +941,323 @@ theorem forall₂_implies_all_right {α β} {R : α → β → Prop} {xs : List
...
+-/
 theorem mapM'_some_implies_all_from_some {α β} {f : α → Option β} {xs : List α} {ys : List β} :
   List.mapM' f xs = .some ys →
   ∀ y ∈ ys, ∃ x ∈ xs, f x = .some y
 := by
   intro h
-  exact forall₂_implies_all_right (mapM'_some_implies_forall₂ h)
+  exact forall₂_implies_all_right (mapM'_some_iff_forall₂.mp h)
+
+theorem mapM_some_implies_all_from_some {α β} {f : α → Option β} {xs : List α} {ys : List β} :
```
khieta, line 1217: Why the "from" in the lemma name?

khieta (commented):
Left some small (marginally useful) comments. Best of luck filling in the `sorry`! I don't have capacity to hack on the proof this week, but could play with it next week if there's anything left to do.

cedar-lean/Cedar/Thm/Authorization/Authorizer.lean
cdisselkoen: I guess they're incomparable.  This version is stronger in that it's a bijection, but weaker in that it only promises `Except.isOk` while the other lemma promises things about the actual values.

cedar-lean/Cedar/Thm/Authorization/Evaluator.lean
cdisselkoen: Not sure myself.  `simp?` suggested this

cdisselkoen: When I mouse over it tells me it tells Lean to apply this before entering the binders instead of after

cedar-lean/Cedar/Thm/Data/List.lean
cdisselkoen: Updated.

cedar-lean/Cedar/Thm/Data/Map.lean
cdisselkoen: This one is `Map.make_cons` and that one is `Set.make_cons`, is that enough to distinguish?  They're basically the same theorem but for `Map.make` and `Set.make` respectively

cedar-lean/Cedar/Thm/Data/Map.lean
cdisselkoen: Updated

cedar-lean/Cedar/Thm/Data/Map.lean
cdisselkoen: Updated

cedar-lean/Cedar/Thm/Data/List.lean
cdisselkoen: Updated

cedar-lean/Cedar/Thm/Data/List.lean
cdisselkoen: Updated

cedar-lean/Cedar/Thm/Data/Map.lean
cdisselkoen: I think `eq_iff_kvs_equiv` should still be true, it just needs a different proof that doesn't rely on `make_eqv`.  At a minimum, I don't think the counterexample above falsifies `eq_iff_kvs_equiv`.

cdisselkoen: Found a proof of `eq_iff_kvs_equiv` that uses `List.sortedBy_equiv_implies_eq` instead.

cedar-lean/Cedar/Thm/Authorization/Evaluator.lean
```diff
@@ -57,16 +57,15 @@ theorem and_true_implies_right_true {e₁ e₂ : Expr} {request : Request} {enti
   simp [evaluate, h₂, Result.as, Coe.coe, Value.asBool] at h₁
   generalize h₃ : (evaluate e₂ request entities) = r₂
   simp [h₃] at h₁
-  cases r₂ <;> simp [Lean.Internal.coeM] at h₁
```
emina, line 60: It is so nice to see these go away!

cedar-lean/Cedar/Thm/Data/List.lean
```diff
@@ -103,8 +103,8 @@ theorem map_equiv (f : α → β) (xs ys : List α) :
   exists p <;>
   rw [List.subset_def] at a b <;>
   simp only [and_true]
-  exact a h
-  exact b h
+  case left  => exact a h
```
emina, line 106: A very minor suggestion that can be safely ignored:  For these very short proofs about cases,  I prefer to use the `.` notation to separate them out instead of explicitly spelling out the case.  In VSCode, you type `\.` to  generate the bullet point anonymous case marker.  It looks cool :)

```
simp only [and_true]
· exact a h  
· exact b h  
```

cedar-lean/Cedar/Thm/Data/List.lean
emina: If we're removing or renaming existing lemmas, let's leave the old name and signature in place, marked as `@[deprecated]`.  See [here](https://github.com/leanprover/std4/blob/f58165d3d6e0b048d89e56944e98d9054b223d9b/Std/Data/BitVec/Lemmas.lean#L13-L14) for an example.

This will make it easier to fix the internal code.  Without that, Lean tells us that the old name is missing, but it can be hard to figure out / remember what's supposed to replace the old name in a 1,000+ LOC file :) 

Once the internal code is patched, I can put up a new PR that removes all the deprecated lemmas.

Alternatively, if we don't want to deal with `@[deprecated]` tags, let's adopt some commenting convention that's easy to grep for / remove, which can help figure out the mapping of old names to new names.

cedar-lean/Cedar/Thm/Data/List.lean
emina: > While the `*_implies_all_some` lemmas state that every element of the input list has a corresponding element in the output list, the `*_implies_all_from_some` lemmas state that every element of the output list has a corresponding element in the input list.

Yes, that was the rationale:  input / output distinction.

cedar-lean/Cedar/Thm/Data/List.lean
```diff
@@ -1056,8 +1412,7 @@ theorem find?_pair_map {α β γ} [BEq α] (f : β → γ) (xs : List (α × β)
   List.find? (λ x => x.fst == k) (List.map (λ x => (x.fst, f x.snd)) xs)
 := by
   induction xs
-  case nil =>
-    simp only [find?_nil, Option.map_none', map_nil]
+  case nil => simp
```
emina, line 1415: These "bare simps" make me nervous.  Can we revert to `simp only`? :). Note that we use `simp only` elsewhere in this proof to close goals.

cedar-lean/Cedar/Thm/Data/List.lean
```diff
@@ -1066,27 +1421,26 @@ theorem find?_pair_map {α β γ} [BEq α] (f : β → γ) (xs : List (α × β)
...
-  List.find? (fun x => x.fst == k) (List.map (Prod.map id f) xs) = .some fx  →
-  ∃ x, xs.find? (fun x => x.fst == k) = .some x ∧ fx = Prod.map id f x
+  List.find? (λ x => x.fst == k) (List.map (Prod.map id f) xs) = .some fx  →
+  ∃ x, xs.find? (λ x => x.fst == k) = .some x ∧ fx = Prod.map id f x
 := by
   intro h
   induction xs
-  case nil =>
-    simp only [map_nil, find?_nil] at h
+  case nil => simp at h
```
emina, line 1443: I'd prefer to leave `simp only` in place, with the goal of eventually replacing (nearly) all `simp`s with `simp only` (as we encounter them).  They are uglier, admittedly, but really good for proof stability.  I'm guessing they improve efficiency too, since search is limited to the `only` lemmas.

cedar-lean/Cedar/Thm/Data/List.lean
```diff
@@ -1100,7 +1454,7 @@ theorem find?_fst_map_implies_find? {α β γ} [BEq α] {f : β → γ} {xs : Li
       replace ⟨x, ih⟩ := ih h
       exists x
       simp only [Prod.map, id_eq] at heq
-      simp only [find?_cons, heq, ih, and_self]
+      simp [find?_cons, heq, ih]
```
emina, line 1457: Same as above:  let's leave `simp only` in place.

cedar-lean/Cedar/Thm/Data/Map.lean
emina: I don't have strong feelings on this one.  The `mapMOnValues_wf` proof is shorter than the `alt` version because it pushes the complexity into `mapMOnValues_preserves_keys`.  If this is the only use case for `mapMOnValues_preserves_keys`, I'd be tempted to go for the `alt` proof and delete the special-purpose lemma.

cedar-lean/Cedar/Thm/Data/Map.lean
emina: Same comment as above :). No strong feelings but my preference is to have fewer specialized lemmas that aren't used anywhere else.  (Obviously, there is a limit to this rule-of-thumb, and I do have super long proofs that are broken into many single-use lemmas---though in that case, I make those `private`.)

emina (changes requested):
LGTM.  Main request for changes: 

* Let's leave `simp only`s in place (and not replace them with `simp`s).
* If a theorem is renamed, either leave the old version as `@[deprecated]` in place, with a pointer to the updated one.  Or, if we don't want to deal with that, let's adopt some greppable commenting style that says which new names are to replace which old names---this really helps with updating internal code.  Otherwise, it's hard to remember / hunt down the new name to use when the old one goes away and breaks the proof.

cedar-lean/Cedar/Thm/Data/List.lean
cdisselkoen: Very reasonable.  Did this for this theorem and the one I commented on above; let me know if there are other instances where we need a deprecated alias.

cedar-lean/Cedar/Thm/Data/Map.lean
cdisselkoen: This is the only use case for `mapMOnValues_preserves_keys`, but my hesitation is that since we've gone through the trouble to prove it now, we might want to keep it in case it is useful in the future.  Or maybe it's just super unlikely to be useful in the future?

cedar-lean/Cedar/Thm/Data/Map.lean
cdisselkoen: Wait is there a `private` keyword in Lean?

In this case, unlike in the above case, neither of the lemmas these proofs rely on exist solely for this proof's benefit.  Schematically, we have
```
A ---- B
|      |
C ---- D
```
This is A, we need proofs of A,B,C,D, and we can choose whether to prove A "through" B or "through" C.  

(In the figure, the top row is Map lemmas, the bottom row is List lemmas, the left side is mapM[OnValues] lemmas, the right side is `forall2` lemmas.)

cedar-lean/Cedar/Thm/Authorization/Evaluator.lean
cdisselkoen: Pretty sure Lean 4.6 or 4.7 newly enabled omitting the `Lean.Internal.coeM` and friends

cedar-lean/Cedar/Thm/Data/List.lean
cdisselkoen: Looks good to me!  Made a pass doing this.  When the case names are informative (`nil`, `cons`, `some`, `none`, etc) I like having them there, but when they are not (looking at you, `split` and `rcases`) I like this notation when the proofs in each case are short.

cedar-lean/Cedar/Thm/Data/List.lean
cdisselkoen: Plenty of "bare simps" existing before this PR (including in this file), but I'll do a pass

cdisselkoen: Noting that we have said previously that `simp` to close a goal (as here) should be perfectly stable

cdisselkoen: Pushed a pass removing some bare simps

cdisselkoen: There's a lot of other `simp` that don't have `simp only`, including in this file.  Can we not block this PR on this issue, and just make an issue to follow up?

cedar-lean/Cedar/Thm/Data/Map.lean
emina: Not sure.  I haven't needed it in the large internal coebase :). But I don't see harm in keeping it.

cedar-lean/Cedar/Thm/Data/Map.lean
emina: Yeah, the `private` keyword is great:  it limits the visibility of the theorem / definition to the defining module.  

Wish it worked for macros!

cedar-lean/Cedar/Thm/Data/List.lean
cdisselkoen: Fixed, and see comment thread above

cedar-lean/Cedar/Thm/Authorization/Evaluator.lean
emina: Yes, it was the update for the win!

cedar-lean/Cedar/Thm/Data/List.lean
emina: Definitely no need to block this PR on old bare `simp`s.  I should have been clearer:  my goal is to not introduce new ones and get rid of old ones eventually :)

emina (approved):

khieta (commented):
Agreed that removing `case left =>` and `case right =>` when possible makes things prettier.

Approving now so you'll meet the approval requirement once @emina signs off.

khieta (approved):
Oops, forgot to check "approve".

Issue #96, opened 2023-08-04
`cedar-policy-generators` should generate (and handle) schemas using common types


cdisselkoen:
### Category

DRT generator(s)

### Describe the feature you'd like to request

Today, the generators do not generate schemas using common types, and `Schema::from_schemafrag()` and friends panic on encountering common types.

### Describe the solution you'd like

The generators should generate schemas using common types (maybe when some setting is enabled), and `Schema::from_schemafrag()` should produce a valid `Schema` when given a schema that includes common types.

### Describe alternatives you've considered

At least not panicking would be an improvement.

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

cdisselkoen:
Partially addressed in #98; see notes there

cdisselkoen:
To be clear, #98 addressed the issues in `Schema::from_schemafrag()` not handling common types.  It does not address the problem that our schema generators do not generate schemas using common types -- that remains to be done.

shaobo-he-aws:
Related to https://github.com/cedar-policy/cedar/pull/766. We will have more sophisticated common type references and hence need DRT to help with the implementation.

john-h-kastner-aws:
Do we feel #306 resolves this by adding a standalone PBT target for common type resolution, or is there additional value in using common types in existing validator targets as well?

PR #306, opened 2024-05-02
Add a target to PBT common type resolution

shaobo-he-aws:
*Issue #, if available:*

Partially addressed #96

*Description of changes:*

The target replaces child-types of `SchemaType`s or themselves with common type IDs in a `SchemaFragment` and tests if the two `SchemaFragment` resolve to the same `ValidatorSchema`.

### DIFF
--- a/cedar-drt/fuzz/src/schemas.rs
+++ b/cedar-drt/fuzz/src/schemas.rs
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 
 use cedar_policy_validator::{ActionType, ApplySpec, NamespaceDefinition, SchemaFragment};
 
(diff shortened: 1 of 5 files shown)


john-h-kastner-aws (approved):

cedar-drt/fuzz/src/schemas.rs
```diff
@@ -115,23 +115,37 @@ fn either_empty(spec: &ApplySpec) -> bool {
...
             .map(|(name, ty)| (name, HashMap::from_iter(ty.attributes()))),
     );
-    let attr_tys2 = HashMap::from_iter(
+    let entity_attr_tys2 = HashMap::from_iter(
         schema2
             .entity_types()
             .map(|(name, ty)| (name, HashMap::from_iter(ty.attributes()))),
     );
-    attr_tys1 == attr_tys2
+    let context_ty1: HashSet<cedar_policy_validator::types::Type> = HashSet::from_iter(
```
john-h-kastner-aws, line 136: Does it need any other changes to use common types in context?

shaobo-he-aws: Yes 😂 Good catch!

shaobo-he-aws: Done.

andrewmwells-amazon (approved):

PR #308, opened 2024-05-08
Partial Evaluation DRT

aaronjeline:
*Issue #, if available:*
N/A
*Description of changes:*
Adds DRT for partial evaluation 

It does find a discrepancy

### DIFF
--- a/cedar-drt/fuzz/fuzz_targets/partial-eval.rs
+++ b/cedar-drt/fuzz/fuzz_targets/partial-eval.rs
@@ -15,34 +15,23 @@
  */
 
 #![no_main]
+use cedar_drt::utils::expr_to_est;
 use cedar_drt::*;
 use cedar_drt_inner::*;
-use cedar_policy_core::ast;
-use cedar_policy_core::ast::Policy;
-use cedar_policy_core::ast::PolicyID;
-use cedar_policy_core::ast::PolicySet;
-use cedar_policy_core::ast::Value;
-use cedar_policy_core::authorizer::AuthorizationError;
-use cedar_policy_core::authorizer::{Authorizer, PartialResponse, Response};
+use cedar_policy_core::ast::Expr;
 use cedar_policy_core::entities::Entities;
+use cedar_policy_generators::schema::arbitrary_schematype_with_bounded_depth;
 use cedar_policy_generators::{
-    abac::{ABACPolicy, ABACRequest},
-    err::Error,
-    hierarchy::HierarchyGenerator,
-    schema::Schema,
+    abac::ABACRequest, err::Error, hierarchy::HierarchyGenerator, schema::Schema,
     settings::ABACSettings,
 };
 use libfuzzer_sys::arbitrary::{self, Arbitrary, Unstructured};
 use log::debug;
+use log::info;
 use serde::Serialize;
-use smol_str::SmolStr;
-use std::collections::HashMap;
-use std::collections::HashSet;
-use std::convert::TryFrom;
-use std::sync::Arc;
 
 /// Input expected by this fuzz target:
-/// An ABAC hierarchy, policy, and 8 associated requests
+/// An ABAC hierarchy, expression, and 8 associated requests
 #[derive(Debug, Clone, Serialize)]
 struct FuzzTargetInput {
     /// generated schema
@@ -51,9 +40,10 @@ struct FuzzTargetInput {
     /// generated entity slice
     #[serde(skip)]
     pub entities: Entities,
-    /// generated policy
-    pub policy: ABACPolicy,
-    /// the requests to try for this hierarchy and policy. We try 8 requests per
+    /// generated expression
+    #[serde(serialize_with = "expr_to_est")]
+    pub expression: Expr,
+    /// the requests to try for this hierarchy and expression. We try 8 requests per
     /// policy/hierarchy
     #[serde(skip)]
     pub requests: [ABACRequest; 8],
@@ -78,8 +68,16 @@ impl<'a> Arbitrary<'a> for FuzzTargetInput {
     fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
         let schema = Schema::arbitrary(SETTINGS.clone(), u)?;
         let hierarchy = schema.arbitrary_hierarchy(u)?;
-        let policy = schema.arbitrary_policy(&hierarchy, u)?;
+        let toplevel_type = arbitrary_schematype_with_bounded_depth(
+            &SETTINGS,
+            schema.entity_types(),
+            SETTINGS.max_depth,
+            u,
+        )?;
 
+        let expr_gen = schema.exprgenerator(Some(&hierarchy));
+        let expression =
+            expr_gen.generate_expr_for_schematype(&toplevel_type, SETTINGS.max_depth, u)?;
         let requests = [
             schema.arbitrary_request(&hierarchy, u)?,
             schema.arbitrary_request(&hierarchy, u)?,
@@ -95,7 +93,7 @@ impl<'a> Arbitrary<'a> for FuzzTargetInput {
         Ok(Self {
             schema,
             entities,
-            policy,
+            expression,
             requests,
         })
     }
@@ -116,121 +114,18 @@ impl<'a> Arbitrary<'a> for FuzzTargetInput {
         ])
     }
 }
-
-/// Check if two `Response`s are equivalent.
-/// They are equivalent iff:
-///   1) They decision is the same
-///   2) If one decision has any errors, the other decision
-///      must have at least one error.
-///      (The error contents and number of errors do not need to be equal)
-fn responses_equiv(a1: Response, a2: Response) -> bool {
-    (a1.decision == a2.decision)
-        && (a1.diagnostics.errors.is_empty() == a2.diagnostics.errors.is_empty())
-}
-
-/// Substitute all policies in the set according to the mapping
-fn substitute_set(p: &PolicySet, mapping: &HashMap<SmolStr, Value>) -> PolicySet {
-    let mut new_set = PolicySet::default();
-    for policy in p.policies() {
-        new_set
-            .add(substitute_policy(policy, mapping))
-            .expect("Failed to add substituted policy to set");
-    }
-
-    new_set
-}
-
-/// Substitute the condition clause of this policy according to the mapping
-fn substitute_policy(p: &Policy, mapping: &HashMap<SmolStr, Value>) -> Policy {
-    let condition = p.condition().substitute(mapping);
-    Policy::from_when_clause_annos(
-        p.effect(),
-        Arc::new(condition),
-        p.id().clone(),
-        None,
-        p.annotations_arc().clone(),
-    )
-}
-
-fn get_id(e: &AuthorizationError) -> &PolicyID {
-    match e {
-        AuthorizationError::PolicyEvaluationError { id, .. } => id,
-    }
-}
-
-/// Check if a partial response is consistent with a concrete response
-fn partial_response_correctness(partial: &PartialResponse, concrete: &Response) {
-    let actually_errored: HashSet<_> = concrete.diagnostics.errors.iter().map(get_id).collect();
-    let partial_errored: HashSet<_> = partial.definitely_errored().collect();
-    // Ensure the errors in the concrete response are a superset of errors in the partial
-    // response
-    assert!(actually_errored.is_superset(&partial_errored));
-    let determining = &concrete.diagnostics.reason;
-    // Ensure that `may_be_determining` produces an over approximation of policies in the
-    // reasons set
-    let over_approx = partial
-        .may_be_determining()
-        .map(|p| p.id().clone())
-        .collect::<HashSet<_>>();
-    assert!(over_approx.is_superset(determining));
-    // Ensure that `must_be_determining` produces an under approximation of policies in the
-    // reasons set
-    let under_approx = partial
-        .must_be_determining()
-        .map(|p| p.id().clone())
-        .collect::<HashSet<_>>();
-    assert!(determining.is_superset(&under_approx));
-}
-
 // The main fuzz target. This is for type-directed fuzzing of ABAC
-// hierarchy/policy/requests
+// hierarchy/expression/requests
 fuzz_target!(|input: FuzzTargetInput| {
     initialize_log();
-    let mut policyset = ast::PolicySet::new();
-    policyset.add_static(input.policy.into()).unwrap();
+    let def_impl = LeanDefinitionalEngine::new();
     debug!("Schema: {}\n", input.schema.schemafile_string());
-    debug!("Policies: {policyset}\n");
-    let mapping = input
-        .schema
-        .unknown_pool
-        .mapping()
-        .map(|(k, v)| (SmolStr::from(k), v))
-        .collect();
+    debug!("Expr: {:?}\n", input.expression);
 
-    if policyset.policies().all(|p| !p.condition().is_unknown()) {
-        return; // Don't waste time testing a policy w/ no unknowns
-    }
+    let request = input.requests.first().unwrap().clone().into();
+    debug!("{request}");
+    let ((), total_dur) =
+        time_function(|| run_pe_test(&def_impl, request, &input.expression, &input.entities, true));
 
-    for q in input.requests.into_iter().map(ast::Request::from) {
-        let auth = Authorizer::new();
-        let ans = auth.is_authorized_core(q.clone(), &policyset, &input.entities);
-
-        if let Some(_decision) = ans.decision() {
-            // Concrete evaluation should also succeed w/out any substitutions
-            let concrete_res = auth.is_authorized(q, &policyset, &input.entities);
-            assert!(responses_equiv(ans.into(), concrete_res));
-        } else {
-            // If we are not able to reach a concrete decision:
-            // First we ensure that if we substitute the residuals and evaluate, we get a concrete
-            // answer consistent with the partial response
-            let concrete_response = ans
-                .reauthorize(&mapping, &auth, q.clone(), &input.entities)
-                .unwrap();
-            assert!(
-                concrete_response.decision().is_some(),
-                "Substituting all `unknown`s should give a decision"
-            );
-            let concrete: Response = concrete_response.into();
-            partial_response_correctness(&ans, &concrete);
-
-            // Then we check if we substitute the original policies and evaluate, we get a concrete
-            // answer consistent with the partial answer
-            let substituted_start = substitute_set(&policyset, &mapping);
-            let final_answer = auth.is_authorized(q.clone(), &substituted_start, &input.entities);
-            partial_response_correctness(&ans, &final_answer);
-
-            // Then check that those two concrete responses are equivalent
-            assert!(responses_equiv(concrete, final_answer));
-        }
-    }
+    info!("{}{}", TOTAL_MSG, total_dur.as_nanos());
 });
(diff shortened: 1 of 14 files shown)


cedar-drt/fuzz/fuzz_targets/partial-eval.rs
```diff
@@ -51,8 +40,9 @@ struct FuzzTargetInput {
     /// generated entity slice
     #[serde(skip)]
     pub entities: Entities,
-    /// generated policy
-    pub policy: ABACPolicy,
+    /// generated expression
+    #[serde(serialize_with = "expr_to_est")]
+    pub expression: Expr,
     /// the requests to try for this hierarchy and policy. We try 8 requests per
     /// policy/hierarchy
```
cdisselkoen, line 47: maybe comment needs updating, if we aren't DRT'ing policies in this target

cedar-drt/fuzz/fuzz_targets/partial-eval.rs
```diff
@@ -15,31 +15,20 @@
...
+use log::info;
 use serde::Serialize;
-use smol_str::SmolStr;
-use std::collections::HashMap;
-use std::collections::HashSet;
-use std::convert::TryFrom;
-use std::sync::Arc;
 
 /// Input expected by this fuzz target:
 /// An ABAC hierarchy, policy, and 8 associated requests
```
cdisselkoen, line 34: maybe comment needs updating. Not sure if it's valuable to list the fields in this comment, since you can just look at the public fields & their docstrings

cedar-drt/fuzz/fuzz_targets/partial-eval.rs
```diff
@@ -116,121 +114,18 @@ impl<'a> Arbitrary<'a> for FuzzTargetInput {
...
-                .reauthorize(&mapping, &auth, q.clone(), &input.entities)
-                .unwrap();
-            assert!(
-                concrete_response.decision().is_some(),
-                "Substituting all `unknown`s should give a decision"
-            );
-            let concrete: Response = concrete_response.into();
-            partial_response_correctness(&ans, &concrete);
+    let request = input.requests.first().unwrap().clone().into();
+    debug!("{}", request);
```
cdisselkoen, line 126:
```suggestion
    debug!("{request}");
```

cedar-drt/fuzz/src/lib.rs
```diff
@@ -43,6 +43,64 @@ use std::collections::HashSet;
...
+    let eval = Evaluator::new(request.clone(), entities, &exts);
+    use cedar_policy_core::ast::PartialValue;
+    use cedar_testing::cedar_test_impl::ExprOrValue;
+    use log::debug;
+    let expected = match eval.partial_interpret(expr, &std::collections::HashMap::default()) {
+        Ok(PartialValue::Value(v)) => Some(ExprOrValue::value(v)),
+        Ok(PartialValue::Residual(r)) => Some(ExprOrValue::Expr(r)),
+        Err(_) => None,
+    };
+    debug!("Expected: {:?}", expected);
```
cdisselkoen, line 71:
```suggestion
    debug!("Expected: {expected:?}");
```

cedar-drt/fuzz/src/lib.rs
```diff
@@ -43,6 +43,64 @@ use std::collections::HashSet;
...
+        expected.clone(),
+    );
+    match definitional_res {
+        TestResult::Failure(err) => {
+            // TODO(#175): Ignore cases where the definitional code returned an error due to
+            // an unknown extension function.
+            //      if err.contains("jsonToExtFun: unknown extension function") {
+            //         return;
+            //    }
+            // No other errors are expected
```
cdisselkoen, line 87: I'm confused whether this is still a TODO or not, since the relevant code seems to be commented out

cedar-drt/src/definitional_request_types.rs
```diff
@@ -34,6 +35,14 @@ pub struct EvaluationRequest<'a> {
     pub expected: Option<&'a ast::Expr>,
 }
 
+#[derive(Debug, Serialize)]
+pub struct PartialEvlauationRequest<'a> {
```
cdisselkoen, line 39:
```suggestion
pub struct PartialEvaluationRequest<'a> {
```

cedar-drt/src/lean_impl.rs
```diff
@@ -227,6 +230,45 @@ impl LeanDefinitionalEngine {
         }
     }
 
+    fn deserialize_partial_evaluation_response(response_string: String) -> TestResult<bool> {
+        use log::debug;
+        debug!("Response: `{response_string}`");
+        let resp: PartialEvaluationResponse =
+            serde_json::from_str(&response_string).expect("could not deserialize json");
+        debug!("resp: `{:?}`", resp);
```
cdisselkoen, line 238:
```suggestion
        debug!("resp: `{resp:?}`");
```

cedar-lean/Cedar/Partial/Evaluator.lean
```diff
@@ -129,12 +129,16 @@ def evaluateVar (v : Var) (req : Partial.Request) : Result Partial.Value :=
     | some kvs => .ok (.value (Map.make kvs))
     | none     => .ok (.residual (Partial.Expr.record (req.context.kvs.map fun (k, v) => (k, v.asPartialExpr))))
 
+def call : ExtFun → List Spec.Value → Result Value
+  | .unknown, [.prim (.string name)] => .ok (.residual (.unknown name))
+  | fn, args =>  do
+    let val ←  Spec.call fn args
```
cdisselkoen, line 135:
```suggestion
    let val ← Spec.call fn args
```

cedar-lean/Cedar/Partial/Expr.lean
```diff
@@ -15,8 +15,8 @@
 -/
 
 import Cedar.Data
+import Cedar.Thm.Data.List
```
cdisselkoen, line 18: @emina doesn't want files in `Cedar/Partial` to import files in `Cedar/Thm`

cedar-lean/Cedar/Partial/Expr.lean
```diff
@@ -48,6 +48,27 @@ inductive Expr where
 
 deriving instance Repr, Inhabited for Expr
 
+
+-- Do we want some kind of roundtrip proof here?
```
cdisselkoen, line 52: I think that would be nice

cedar-lean/Cedar/Thm/Validation/Typechecker/Call.lean
```diff
@@ -373,6 +373,46 @@ theorem type_of_call_ipAddr_recognizer_is_sound {xfn : ExtFun} {xs : List Expr}
       apply bool_is_instance_of_anyBool
     }
 
+-- It's impossible to call `unknown` in a well typed way
+theorem type_of_unknown_call_untyped :
+  ∀ (f : ExtFun) (xs : List Expr) (c : Capabilities) (env : Environment),
```
cdisselkoen, line 378: Our style would usually be
```suggestion
theorem type_of_unknown_call_untyped (f : ExtFun) (xs : List Expr) (c : Capabilities) (env : Environment) :
```

I'm not sure offhand why/whether we should prefer one style over the other

cedar-lean/Cedar/Thm/Validation/Typechecker/Call.lean
```diff
@@ -373,6 +373,46 @@ theorem type_of_call_ipAddr_recognizer_is_sound {xfn : ExtFun} {xs : List Expr}
...
+  f = .unknown ∨ f = .error →
+  ∃ e, typeOf (.call f xs) c env = err e :=
+  by
+  intros f xs c env h
+  unfold typeOf
+  simp
```
cdisselkoen, line 384: Our recommendation in one of our READMEs is to avoid bare `simp` except to close a goal, for proof stability reasons (hedging against lemmas being added/removed from `simp` in future versions of Lean).  You can write `simp?` and then click to accept the suggestion of an equivalent `simp only` command.

cedar-lean/Cedar/Thm/Validation/Typechecker/Call.lean
```diff
@@ -373,6 +373,46 @@ theorem type_of_call_ipAddr_recognizer_is_sound {xfn : ExtFun} {xs : List Expr}
...
+  case error e' => {
+    -- Either there's a type error in the arguments
+    exists e'
+  }
+  case ok _ => {
+    -- Or the call itself has no typing rule
+    cases h
+    <;> rename_i hf
+    <;> rw [hf]
+    <;> simp
```
cdisselkoen, line 395: another `simp` to replace with `simp only` per our style guide

cedar-lean/Cedar/Partial/Expr.lean
```diff
@@ -48,6 +48,27 @@ inductive Expr where
...
+def Expr.asSpec (e : Expr) : Option Spec.Expr :=
+  match e with
+  | .lit p => some (.lit p)
+  | .var v => some (.var v)
+  | .ite cond cons alt =>  .ite <$> cond.asSpec <*> cons.asSpec <*> alt.asSpec
```
cdisselkoen, line 57:
```suggestion
  | .ite cond cons alt => .ite <$> cond.asSpec <*> cons.asSpec <*> alt.asSpec
```

cedar-lean/DiffTest/Parser.lean
```diff
@@ -246,6 +261,26 @@ partial def exprToValue : Expr → ParseResult Value
 def jsonToValue (json : Lean.Json) : ParseResult Value :=
   jsonToExpr json >>= exprToValue
 
+
+inductive ValueOrExpr where
+  | value : Value → ValueOrExpr
+  | expr : Cedar.Partial.Expr  → ValueOrExpr
```
cdisselkoen, line 267:
```suggestion
  | expr : Cedar.Partial.Expr → ValueOrExpr
```

cdisselkoen, line 267: What's the difference between this and
```
inductive ValueOrExpr where
  | value (v : Value)
  | expr (x : Cedar.Partial.Expr)
```

cedar-lean/DiffTest/Parser.lean
```diff
@@ -246,6 +261,26 @@ partial def exprToValue : Expr → ParseResult Value
...
+
+def jsonToValueOrExpr (json : Lean.Json) : ParseResult ValueOrExpr := do
+  match json.getObjVal? "Value" with
+    | .ok v => ValueOrExpr.value <$> (jsonToValue v)
+    | .error _ => match json.getObjVal? "Expr" with
+      | .ok e => ValueOrExpr.expr <$> (jsonToPartialExpr e)
+      | .error _ => .error "Expected either `expr` or `value`"
```
cdisselkoen, line 274: did we expect `expr`/`value`, or `Expr`/`Value`?

cedar-drt/fuzz/fuzz_targets/partial-eval.rs
```diff
@@ -78,8 +68,16 @@ impl<'a> Arbitrary<'a> for FuzzTargetInput {
...
+        let toplevel_type = arbitrary_schematype_with_bounded_depth(
+            &SETTINGS,
+            schema.entity_types(),
+            SETTINGS.max_depth,
+            u,
+        )?;
+
+        let expr_gen = schema.exprgenerator(Some(&hierarchy));
+        let expression =
+            expr_gen.generate_expr_for_schematype(&toplevel_type, SETTINGS.max_depth, u)?;
```
cdisselkoen, line 80: will this generate expressions that include unknowns?

cedar-drt/fuzz/fuzz_targets/partial-eval.rs
```diff
@@ -95,7 +93,7 @@ impl<'a> Arbitrary<'a> for FuzzTargetInput {
         Ok(Self {
             schema,
             entities,
-            policy,
+            expression,
             requests,
```
cdisselkoen, line 97: will `entities` and `requests` include entity/context attribute values that contain unknowns?

cdisselkoen (commented):
Looks good to me overall

cedar-lean/Cedar/Thm/Validation/Typechecker/Call.lean
cdisselkoen: or with `csimp` if we accept #311

cedar-lean/DiffTest/Parser.lean
```diff
@@ -165,57 +169,68 @@ partial def jsonToExpr (json : Lean.Json) : ParseResult Expr := do
...
+  | "Unknown" => do
+    let name ← getJsonField body "name" >>= jsonToString
+    .ok (.unknown name)
+  | tag => .error s!"jsonToPartialExpr: unknown tag {tag}"
+
+def jsonToExpr (json : Lean.Json) : ParseResult Expr := do
+  let pe ← jsonToPartialExpr json
+  match pe.asSpec with
+  | some e => .ok e
+  | none => .error s!"expression contained unknown"
```
khieta, line 231:
```suggestion
  | none => .error s!"jsonToExpr: expression contained unknown"
```
When debugging in the past, I've found it useful to include the function name in the error message.

cedar-lean/DiffTest/Parser.lean
khieta: I think the difference is just whether the fields are named.

cedar-policy-generators/src/expr.rs
```diff
@@ -268,11 +268,7 @@ impl<'a> ExprGenerator<'a> {
         if self.should_generate_unknown(max_depth, u)? {
             let v = self.generate_value_for_type(target_type, max_depth, u)?;
             let name = self.unknown_pool.alloc(target_type.clone(), v);
-            let unknown_type: Option<ast::Type> = target_type.clone().try_into().ok();
-            match unknown_type {
-                Some(ty) => Ok(ast::Expr::unknown(ast::Unknown::new_with_type(name, ty))),
-                None => Ok(ast::Expr::unknown(ast::Unknown::new_untyped(name))),
-            }
+            Ok(ast::Expr::unknown(ast::Unknown::new_untyped(name)))
```
khieta, line 271: What's the reason for this change?

cedar-lean/DiffTest/Parser.lean
khieta, line 274:
```suggestion
      | .error _ => .error "jsonToValueOrExpr: expected either `Expr` or `Value`"
```

cedar-lean/Cedar/Spec/ExtFun.lean
```diff
@@ -37,6 +37,8 @@ inductive ExtFun where
   | isLoopback
   | isMulticast
   | isInRange
+  | unknown             ----- Partial Evaluation Functions ----
+  | error
```
khieta, line 41: What is the `error` extension for?

If possible, I think it's be better to keep the partial evaluation functionality contained to `Cedar/Partial` rather than let it leak into `Cedar/Spec`.

khieta (commented):
Some high-level comments
* I imagine we'll also want to be DRT'ing policies, since the logic to compute the "definitely erroring policies" and such is a little confusing. Might make sense to add an additional target in this PR, like how we have `abac-type-directed` and `eval_type_directed`.
* Let's change the name to `partial-eval-type-directed` for consistency with the other targets.
* Can you add support for PE to `Cli/Main.lean` too? Should be easy since you've already set up the DRT functionality.

aaronjeline:
> Some high-level comments
> 
> * I imagine we'll also want to be DRT'ing policies, since the logic to compute the "definitely erroring policies" and such is a little confusing. Might make sense to add an additional target in this PR, like how we have `abac-type-directed` and `eval_type_directed`.

Absolutely, just waiting for partial auth code to be merged 


> * Let's change the name to `partial-eval-type-directed` for consistency with the other targets.

Will do 
> * Can you add support for PE to `Cli/Main.lean` too? Should be easy since you've already set up the DRT functionality.

Sure

cedar-lean/Cedar/Partial/Expr.lean
aaronjeline: Ok, if I need to for theorems where should I move those theorems? (And also curious what's the organizational reason?)

cedar-drt/fuzz/src/lib.rs
aaronjeline: Mistake, commented out for debugging.

cedar-lean/DiffTest/Parser.lean
emina: Yes that’s the only difference IIRC:  you get the specified names in the error messages and other help widgets.

cedar-lean/Cedar/Spec/ExtFun.lean
emina: Can we think of some other solution here?  This error function is not part of the language spec itself, and it would be a pity to make the spec messier for a specific feature like PE.

aaronjeline: It will be part of the spec once we release partial evaluation no? As in it will be a valid extension function?

emina: Thinking about it more, I think this function is quite problematic.  Specifically, what is its type? If we can't assign a type to it, then PE will produce ill-typed expressions even when given well-typed input. 

In other words, PE should preserve correct types, just like it preserves correct evaluation semantics (for non-erroring executions).

aaronjeline: If we don't like this solution then we have to be ok with not evaluating branches to errors, which is maybe fine.

cedar-drt/fuzz/fuzz_targets/partial-eval.rs
aaronjeline: yes

emina (approved):

cdisselkoen (approved):
Nothing in my comments needs to be blocking

cedar-policy-generators/src/expr.rs
cdisselkoen: Bumping this question from @khieta as well

cdisselkoen (approved):

cedar-drt/fuzz/fuzz_targets/partial-eval.rs
aaronjeline: It should, as it's using the `exprgenerator` from the schema, which has a single settings blob that says to generator unknowns.

cedar-policy-generators/src/expr.rs
aaronjeline: Reverted, good catch @khieta

cedar-lean/Cedar/Thm/Partial/Expr.lean
```diff
@@ -0,0 +1,23 @@
...
+ limitations under the License.
+-/
+
+import Cedar.Thm.Partial.Expr.Lemmas
+namespace Cedar.Partial.Thm.Expr
+
+theorem roundtrip (e : Spec.Expr) (pe : Partial.Expr) :
+  e.asPartialExpr = pe ↔ pe.asSpec = some e := by
+  apply Iff.intro <;> intros h
+  all_goals { (first | apply roundtrip₂ | apply roundtrip₁) <;> assumption }
```
cdisselkoen, line 23: maybe something like this is clearer?
```suggestion
  · apply roundtrip₂ ; assumption 
  · apply roundtrip₁ ; assumption
```

cdisselkoen: Or, instead of `assumption`, use `h` directly.
```suggestion
  · exact roundtrip₂ _ _ h 
  · exact roundtrip₁ _ _ h
```
or something similar

cedar-lean/DiffTest/Parser.lean
cdisselkoen: did you reject this suggestion? it was resolved but it doesn't appear in the latest diff

cdisselkoen (approved):

PR #313, opened 2024-05-09
More datastructure lemmas.

emina:
This PR removes deprecated List lemmas and adds more useful lemmas about Lists and Sets.

### DIFF
--- a/cedar-lean/Cedar/Thm/Data/Set.lean
+++ b/cedar-lean/Cedar/Thm/Data/Set.lean
@@ -38,6 +38,22 @@ theorem if_wellformed_then_exists_make [LT α] [DecidableLT α] (s : Set α) :
   intro h₁
   exists s.elts
 
+theorem wf_iff_sorted {α} [LT α] [DecidableLT α] [StrictLT α]  (s : Set α) :
+  s.WellFormed ↔ s.elts.Sorted
+:= by
+  simp only [WellFormed, make, toList, elts]
+  have hid : (fun x => x) = @id α := by
+    apply funext ; simp only [id_eq, implies_true]
+  rw [hid]
+  constructor <;>
+  intro h
+  case mp =>
+    rw [h]
+    simp only [List.Sorted, List.canonicalize_sortedBy id s.1]
+  case mpr =>
+    rw [← (List.sortedBy_implies_canonicalize_eq h),
+      List.canonicalize_idempotent, List.sortedBy_implies_canonicalize_eq h]
+
 /-! ### contains and mem -/
 
 theorem contains_prop_bool_equiv [DecidableEq α] {v : α} {s : Set α} :
@@ -91,6 +107,10 @@ theorem in_set_means_list_non_empty {α : Type u} (v : α) (s : Set α) :
   contradiction
 
 /-! ### empty -/
+
+theorem empty_eq_mk_empty {α} :
+  (Set.empty : Set α) = Set.mk []
+:= by simp only [empty]
 
 theorem empty_no_elts {α : Type u} (v : α) :
   ¬ v ∈ Set.empty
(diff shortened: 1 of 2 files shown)


cedar-lean/Cedar/Thm/Data/Set.lean
```diff
@@ -321,6 +353,74 @@ theorem inter_wf {α} [LT α] [StrictLT α] [DecidableLT α] [DecidableEq α] {s
...
+  simp only [intersects, List.any_eq_true, in_list_iff_in_set, contains_prop_bool_equiv]
+  constructor
+  case mp =>
+    intro h
+    replace ⟨x, h⟩ := h
+    by_contra hc
+    simp only [ne_eq, Bool.not_eq_false] at hc
+    simp only [empty_iff_not_exists, not_exists] at hc
+    specialize hc x
+    have _ := mem_inter_iff.mpr h
```
cdisselkoen, line 400:
```suggestion
    have := mem_inter_iff.mpr h
```

cedar-lean/Cedar/Thm/Data/Set.lean
```diff
@@ -38,6 +38,22 @@ theorem if_wellformed_then_exists_make [LT α] [DecidableLT α] (s : Set α) :
...
 
+theorem wf_iff_sorted {α} [LT α] [DecidableLT α] [StrictLT α]  (s : Set α) :
+  s.WellFormed ↔ s.elts.Sorted
+:= by
+  simp only [WellFormed, make, toList, elts]
+  have hid : (fun x => x) = @id α := by
+    apply funext ; simp only [id_eq, implies_true]
+  rw [hid]
+  constructor <;>
+  intro h
```
cdisselkoen, line 49: In other places (like line 926 above) you have this on one line.  Which is preferable?

cedar-lean/Cedar/Thm/Data/Set.lean
```diff
@@ -179,6 +199,14 @@ theorem make_wf [LT α] [DecidableLT α] [StrictLT α] (xs : List α) :
 := by
   simp only [WellFormed, make, toList, elts, List.canonicalize_idempotent]
 
+theorem make_sorted {α} [LT α] [DecidableLT α] [StrictLT α] {xs : List α} :
+  xs.Sorted → Set.make xs = Set.mk xs
+:= by
+  intro h
+  have hid : (fun x => x) = @id α := by
+    apply funext ; simp only [id_eq, implies_true]
```
cdisselkoen, line 207: Is this not true by definitional equality?  Does the `simp only` below fail if you don't include `hid` as a lemma?

cdisselkoen (approved):
Looks good

cedar-lean/Cedar/Thm/Data/Set.lean
emina: Ah, I use `have _ := ` to make it clear that the introduced hypothesis will never be used by name.  Note that `have := ` introduces a usable name into the proof context (`this`).

cedar-lean/Cedar/Thm/Data/Set.lean
emina: No thought out preference though empirically, it appears that I favor putting things on line :D

cedar-lean/Cedar/Thm/Data/Set.lean
cdisselkoen: I see

cedar-lean/Cedar/Thm/Data/Set.lean
emina: Surprisingly, this is necessary, and `simp only` does indeed fail without it.

cdisselkoen: Weird

khieta (approved):

PR #314, opened 2024-05-10
Overly permissive LUB for qualified attr types to match Rust (3.2.x)

john-h-kastner-aws:
Applies #263 to 3.2.x release branch

The least upper bound between a required and optional qualified attribute type is defined as an optional qualified type to match the Rust implementation.  This is sound, but is too permissive for our SMT encoding. We apply the change the 3.2.x release branch nonetheless so that differential testing does not report this discrepancy.

### DIFF
--- a/cedar-lean/Cedar/Thm/Validation/Typechecker/LUB.lean
+++ b/cedar-lean/Cedar/Thm/Validation/Typechecker/LUB.lean
@@ -110,7 +110,7 @@ theorem lubRecord_find_implies_find_left {a : Attr} {qty : QualifiedType} {rty r
   (h₂ : Map.find? (Map.mk rty) a = some qty) :
   ∃ qty₁,
     Map.find? (Map.mk rty₁) a = some qty₁ ∧
-    Qualified.isRequired qty₁ = Qualified.isRequired qty
+    (Qualified.isRequired qty → Qualified.isRequired qty₁)
 := by
   have ⟨qty₁, qty₂, h₃, h₄, h₅⟩ := lubRecord_find_implies_find h₁ h₂
   exists qty₁ ; simp [h₃]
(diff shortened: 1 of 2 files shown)


khieta:
Does this discrepancy still exist in the 3.1 branch?

john-h-kastner-aws:
> Does this discrepancy still exist in the 3.1 branch?

Yes, the discrepancy should exist there. Not clear why we've never seen it on that version, but nightly testing of 3.2 found it tonight.

john-h-kastner-aws:
The changes also applies cleanly to 3.1 and 3.0, so we could backport it further, but 3.0 isn't tested nightly anymore and we'll probably drop 3.1 after release 3.2

khieta:
No need to backport to 3.0 (or earlier) since we started nightly difftesting against the Lean in 3.1. I'm also surprised we haven't observed the behavior difference in 3.1.

khieta (approved):

emina:
> Applies #263 to 3.2.x release branch
> 
> The least upper bound between a required and optional qualified attribute type is defined as an optional qualified type to match the Rust implementation. This is sound, but is too permissive for our SMT encoding. We apply the change the 3.2.x release branch nonetheless so that differential testing does not report this discrepancy.

I think we need to fix the Rust code here.  The strict validator is a precursor to the analysis, and needs to reject policies that cannot be analyzed.  This implementation will accept such policies.

Is the plan to fix the Rust code eventually, and this is a temporary patch?

andrewmwells-amazon:
> I think we need to fix the Rust code here. The strict validator is a precursor to the analysis, and needs to reject policies that cannot be analyzed. This implementation will accept such policies.
> 
> Is the plan to fix the Rust code eventually, and this is a temporary patch?

Updating from slack thread:
The fixed Rust version will be released in 4.0 (https://github.com/cedar-policy/cedar/pull/769)

PR #317, opened 2024-05-10
split Cedar.Thm.Data.List into submodules

cdisselkoen:
Another PR suggesting a reorganization that folks can feel free to shoot down if it's undesired or not idiomatic or etc.

Cedar/Thm/Data/List.lean is very long (1600ish lines).  This PR breaks it into submodules, roughly along the borders of its existing section headers.  This brings a few advantages:
- Easier to find what you're looking for when it's not a 1600-line file
- Potential performance improvement, as some List lemmas can now be processed in parallel during `lake build`
- Definitely a performance improvement when working in the List file itself, as restarting the file doesn't have to restart all of the List lemmas from the top


emina:
Idiomatic Lean code prefers one long file for "grab bag lemmas" over many small files.  For example, the List lemmas module in the standard library is all in a single file with 2,800+ LOC: https://github.com/leanprover-community/batteries/blob/main/Batteries/Data/List/Lemmas.lean 

I found this unusual when I started but have come to appreciate it.  When you don't know / remember the name of a lemma, searching for it across multiple files is a lot harder (for me at least) than quickly scrolling through one big file.  

While we did our best to partition the list lemmas according to their functionality (as did the standard library), many of these lemmas deal with more than one function / definition (e.g., `mem` lemmas).  And since they can be located in only one file, if I guess wrong where a lemma about a function may be located, it's going to take me a lot longer to find it than scrolling through the mega file.  (Speaking from experience: I'm currently regretting organizing these "grab bag lemmas" into several smaller modules in the internal development early on, and will merge some of these into fewer bigger files.)

So, I'd vote to keep the One Big File for List, except for the `canonical` and `insertCanonical` lemmas.  Those are special-purpose and moving them all together into a separate `Canonical` module would be consistent with how the standard library is organized.  For example, they put specialized permutation lemmas into the `Count` module: https://github.com/leanprover-community/batteries/blob/main/Batteries/Data/List/Perm.lean

cdisselkoen:
That sounds fine to me

emina:
Here is the Std-like split (and a few new lemmas): https://github.com/cedar-policy/cedar-spec/pull/321

cdisselkoen:
Closing this in favor of #321

Issue #268, opened 2024-03-29
Improve performance of `formatter` fuzz target


john-h-kastner-aws:
### Category

DRT target(s)

### Describe the feature you'd like to request

This DRT target is regularly generating "slow units" in our nightly testing. Unlike with some other targets, these are easily reproducible from the recorded test files.

on the current [HEAD](https://github.com/cedar-policy/cedar-spec/tree/be4bae36dca1ed59fdd05558bba4b99794d42c17):

```bash
base64 --decode > slow_unit <<EOF
/wEm//8BAABH////////////////////////////////////////////////////////////////
////////////////////////////////////////////////////////////////MjIy3jAyMjIv
MgAyMjIx2sUy////////////////////////////////////jo6Ojo6Ojo6Ojo6Ojo6Ojo6Ojo6O
jo6Ojo6Ojo6Ojo6Ojo6Ojo6Ojo6Ojo6Ojo6Ojo6Ojo6Ojo6Ojo6Ojo6Ojo6Ojo6Ojo6Ojo6Ojo6O
jo6Ojo6Ojo6Ojo6Ojo6Ojo6Ojo6Ojo6Ojo6Ojo6Ojo6Ojo6Ojo6Ojo6Ojo6Ojo6Ojo6Ojo7/////
//////////////////////////////////////////////+Ojo6Ojo6OjuHh4f////8Q1tbW1tbW
1tbW1tbW1tbW1tbW1tbW1tbW1tbW1tDW1gAAAAAAHDMAFAAr/////////w8AfzMzOA==
EOF
cargo fuzz run -s none formatter slow_unit
```

should output something like

```
fuzz/target/x86_64-unknown-linux-gnu/release/formatter: Running 1 inputs 1 time(s) each.
Running: slow_unit
Executed slow_unit in 16102 ms
```

I believe there are two issues here:

1. The Cedar policy generated by this test case is large, and takes some significant time to format. Formatting [this policy file](https://gist.github.com/john-h-kastner-aws/f126dfe0f9cf325f6120294b1ea1b851#file-slow_format-cedar) takes ~1.5 seconds while just checking that it parses only takes ~0.1 seconds (measured from the CLI). This still leaves most of the time unaccounted for. 
[...]

PR #318, opened 2024-05-13
Assert that comments are not reordered by formatter

john-h-kastner-aws:
The formatter fuzz target only  checked that all comments in the original policy were present in the output without checking that they appeared in the same order.  This allowed cedar-policy/cedar#787 to exist. The fuzz target with the new, stronger, assertion passed local fuzzing after applying the fixes in  cedar-policy/cedar#861.

As a pleasant side effect, it's actually more efficient to check for the comments in the order they were inserted than in an arbitrary order. The check is now a single linear scan rather than a linear scan for each comment, likely resolving the slow units noted in #268. The slow input in that issue is improved from 16 second to about 4 seconds.

### DIFF
--- a/cedar-drt/fuzz/fuzz_targets/formatter.rs
+++ b/cedar-drt/fuzz/fuzz_targets/formatter.rs
@@ -27,6 +27,7 @@ use cedar_policy_generators::{
 use libfuzzer_sys::arbitrary::{self, Arbitrary, Unstructured};
 use log::debug;
 use serde::Serialize;
+use similar_asserts::SimpleDiff;
 use smol_str::SmolStr;
 use std::collections::HashMap;
 use uuid::Uuid;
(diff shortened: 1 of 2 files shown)


cdisselkoen (approved):

shaobo-he-aws (approved):

cedar-drt/fuzz/fuzz_targets/formatter.rs
```diff
@@ -104,12 +105,31 @@ fn round_trip(p: &StaticPolicy) -> Result<StaticPolicy, parser::err::ParseErrors
...
-        &policies_str_to_pretty(&attach_comment(&p.to_string(), &mut uuids), &config)
-            .expect("pretty-printing should not fail");
+        &policies_str_to_pretty(&commented, &config).expect("pretty-printing should not fail");
     // check if pretty-printing drops any comment
+    let mut formatted_policy_tail: &str = formatted_policy_str.as_str();
     for u in &uuids {
-        assert!(formatted_policy_str.contains(u), "missing comment: {}\n", u);
+        match formatted_policy_tail.split_once(u) {
+            Some((_, after_uuid)) => formatted_policy_tail = after_uuid,
+            None => {
```
shaobo-he-aws, line 116: Nit: remove it?

john-h-kastner-aws: It's printed only once we know the test case is an error, so no reason to remove it, and it makes debugging failures much easier

PR #320, opened 2024-05-14
First proof about the partial authorizer

cdisselkoen:
While #291 was the first proof about the partial _evaluator_, this is the first proof about the partial _authorizer_.  It states that partial authorization and concrete authorization produce the same decision when applied to concrete inputs.  Like #291, this theorem is limited in that it says nothing about the behavior of partial authorization on general inputs.  It also is limited in that it only talks about the authorization _decision_ and not the other outputs (determining policies etc).  All of these limits will be addressed in future theorems in future PRs.


aaronjeline (approved):

andrewmwells-amazon (approved):

emina (approved):
LGTM!

PR #324, opened 2024-05-15
`PartialOnConcrete` proof for the rest of the `Partial.Response`

cdisselkoen:
Follows on to #320 by extending the proof to not cover just the partial decision, but also the other parts of the `Partial.Response`: the `overapproximateDeterminingPolicies`, `underapproximateDeterminingPolicies`, and `errors`.  Wraps everything up in a top-level theorem `partial_authz_eqv_authz_on_concrete`.

### DIFF
--- a/cedar-lean/Cedar/Thm/Partial/Authorization.lean
+++ b/cedar-lean/Cedar/Thm/Partial/Authorization.lean
@@ -132,14 +132,14 @@ theorem underapproximate_determining_eqv_determining_on_concrete {policies : Pol
     <;> simpa [h₃] using h₁
 
 /--
-  On concrete inputs, partial authorization's `errors` are the same policies as
-  concrete authorization's `erroringPolicies`
+  On concrete inputs, partial authorization's `errorPolicies` are the same
+  policies as concrete authorization's `erroringPolicies`
 -/
-theorem partial_authz_errors_eqv_erroringPolicies_on_concrete {policies : Policies} {req : Spec.Request} {entities : Spec.Entities} {presp : Partial.Response} {resp : Spec.Response}
+theorem partial_authz_errorPolicies_eqv_erroringPolicies_on_concrete {policies : Policies} {req : Spec.Request} {entities : Spec.Entities} {presp : Partial.Response} {resp : Spec.Response}
   (wf : req.WellFormed) :
   Spec.isAuthorized req entities policies = resp →
   Partial.isAuthorized req entities policies = presp →
-  Set.make (presp.errors.map Prod.fst) = resp.erroringPolicies
+  presp.errorPolicies = resp.erroringPolicies
 := by
   intro h₁ h₂
   subst h₁ h₂
@@ -147,7 +147,7 @@ theorem partial_authz_errors_eqv_erroringPolicies_on_concrete {policies : Polici
   cases (Spec.satisfiedPolicies .forbid policies req entities).isEmpty <;>
   cases (Spec.satisfiedPolicies .permit policies req entities).isEmpty <;>
   simp only [and_true, and_false, ite_true, ite_false] <;>
-  exact PartialOnConcrete.errors_eq_errorPolicies wf
+  exact PartialOnConcrete.errorPolicies_eq_errorPolicies wf
 
 /--
   Partial-authorizing with concrete inputs gives the same concrete outputs as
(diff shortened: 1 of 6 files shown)


cedar-lean/Cedar/Thm/Partial/Authorization.lean
```diff
@@ -67,4 +69,103 @@ theorem partial_authz_on_concrete_gives_concrete {policies : Policies} {req : Sp
...
+
+/--
+  On concrete inputs, partial authorization's `errors` are the same policies as
+  concrete authorization's `erroringPolicies`
+-/
+theorem partial_authz_errors_eqv_erroringPolicies_on_concrete {policies : Policies} {req : Spec.Request} {entities : Spec.Entities} {presp : Partial.Response} {resp : Spec.Response}
+  (wf : req.WellFormed) :
+  Spec.isAuthorized req entities policies = resp →
+  Partial.isAuthorized req entities policies = presp →
+  Set.make (presp.errors.map Prod.fst) = resp.erroringPolicies
```
khieta, line 142: Nit: I feel like it would be more consistent if the partial response also used the name `erroringPolicies` and returned just the policy ids.

cedar-lean/Cedar/Thm/Partial/Authorization/PartialOnConcrete.lean
```diff
@@ -200,4 +200,53 @@ theorem knownForbids_eq_forbids {policies : Policies} {req : Spec.Request} {enti
   unfold Partial.Response.knownForbids Partial.Response.forbids
   apply mustBeSatisfied_eq_mayBeSatisfied (eff := .forbid) wf
 
+/--
+  on concrete inputs, `Partial.Response.errors` and `Spec.errorPolicies` are equal
+-/
+theorem errors_eq_errorPolicies {policies : Policies} {req : Spec.Request} {entities : Spec.Entities}
+  (wf : req.WellFormed) :
+  Set.make ((Partial.isAuthorized req entities policies).errors.map Prod.fst) =
+  Spec.errorPolicies policies req entities
```
khieta, line 209: Nit: Another naming inconsistency: `errors` vs `errorPolicies`

khieta (approved):
Looks fine to me. Left some comments about naming just because I'm picky 🙂 You can choose to ignore.

cedar-lean/Cedar/Thm/Partial/Authorization/PartialResponse.lean
```diff
@@ -0,0 +1,82 @@
+/-
+ Copyright 2022-2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
andrewmwells-amazon, line 2:
```suggestion
 Copyright Cedar Contributors
```

andrewmwells-amazon (approved):

cedar-lean/Cedar/Thm/Partial/Authorization.lean
cdisselkoen: For Spec, we made the conscious choice that we only wanted the spec to specify which policies error, not the kind of error.  I suppose we should make the same choice here?  Just noting that this is a consequential choice in terms of how tightly we are specifying the behavior of PE.

cedar-lean/Cedar/Thm/Partial/Authorization/PartialOnConcrete.lean
cdisselkoen: Currently, `errors` returns actual errors while `errorPolicies` returns only policy IDs.  We could change this, but this has more ramifications, see thread above

cedar-lean/Cedar/Thm/Partial/Authorization.lean
khieta: I feel like we probably want to make the same choice for both. Usually @emina has the strongest opinions about (over-)specifying error behavior.

emina: We should not over-specify error behavior :).  

The current behavior of the Spec code is at the right level of detail:  it says which policies error but does not overspecify the details.  If we mandate the kinds of errors that must returned, we will force all implementations to conform to that error spec---unless we weaken what it means to conform to the spec, which I don't think we should do without very good justification.

cdisselkoen: I wish this was caught on #299 which introduced it :) I'll change it here though.

cedar-lean/Cedar/Thm/Partial/Authorization/PartialOnConcrete.lean
cdisselkoen: Spec is also inconsistent with `errorPolicies` vs `erroringPolicies` (perhaps they wanted to intentionally avoid the naming clash between the standalone function and the field of `Response`)?  I think changing Spec is out of scope for this PR

cedar-lean/Cedar/Thm/Partial/Authorization.lean
cdisselkoen: Fixed and pushed

emina (approved):

PR #325, opened 2024-05-15
Define the `subst` and `reEvaluateWithSubst` operations

cdisselkoen:
Like #281 and #299, this has (mostly) just definitions and not many proofs about them (yet) -- those will come in a separate PR for ease of reviewing.

This defines the `subst` and `reEvaluateWithSubst` operations that are core to the PE soundness proof.  (which roughly will say that, if PE returns residuals and we `reEvaluateWithSubst` substituting for all the unknowns, we get the same, concrete, result as if we had substituted in the first place and used ordinary evaluation.)

This PR comes with three proofs about `subst`: that `Partial.Expr.subst` on a concrete `Expr` gives the same `Expr` back, and that both `Partial.Value.subst` and `Partial.Expr.subst` do likewise for a concrete `Value`.  Proofs about `reEvaluateWithSubst` are left for a future PR.


cedar-lean/Cedar/Spec/Request.lean
```diff
@@ -14,6 +14,7 @@
  limitations under the License.
 -/
 
+import Cedar.Data.Map
```
khieta, line 17: Nit: why the new import here?

cedar-lean/Cedar/Partial/Response.lean
```diff
@@ -195,4 +203,30 @@ def Response.underapproximateDeterminingPolicies (resp : Partial.Response) : Set
...
+  | { residuals, req, entities } => do
+  let req' ← req.subst subsmap
+  some {
+    residuals := residuals.filterMap λ residual => match residual with
+      | .error id e => some (.error id e)
+      | .residual id effect cond => match Partial.evaluate (cond.subst subsmap) req' (entities.subst subsmap) with
+        | .ok (.value (.prim (.bool false))) => none
+        | .ok (.value v) => some (.residual id effect v.asPartialExpr)
+        | .ok (.residual r) => some (.residual id effect r)
+        | .error e => some (.error id e)
```
khieta, line 227: Might be clearer to move this out to a separate function & explicitly document that any residuals that evaluate to false are dropped.

khieta (approved):

cedar-lean/Cedar/Spec/Request.lean
cdisselkoen: This file directly/explicitly uses `Map`, and I find it better style to explicitly import things that we explicitly use, rather than relying on Lean's transitive imports

cedar-lean/Cedar/Partial/Response.lean
cdisselkoen: I can make a note, but I also note that this is the standard behavior of `Partial.isAuthorized`, and indeed documented as an invariant on `Partial.Response`.

cdisselkoen: Which part exactly do you suggest moving out into a separate function?

cdisselkoen: Aha, makes sense

cdisselkoen: Factored this out in the latest commit

khieta (approved):

cedar-lean/Cedar/Partial/Response.lean
```diff
@@ -196,4 +204,46 @@ def Response.underapproximateDeterminingPolicies (resp : Partial.Response) : Set
     -- when the decision is Unknown, nothing is guaranteed to be determining.
     Set.empty
 
+/--
+  Re-evaluate with the given substitution for unknowns, giving a new
+  `Residual`, or `none` if the residual is now `false`.
+
+  Assumes that `req` and `entities` have already been substituted.
```
andrewmwells-amazon, line 211: Instead of having this in a comment, can the type enforce this? (It looks like that would be easy from just this PR, but I don't have the rest of the work paged in like you do, so maybe this is completely off).

andrewmwells-amazon (approved):

cedar-lean/Cedar/Partial/Response.lean
cdisselkoen: In what way?  Having a wrapper around `Partial.Entities` that just indicates "we've performed a `subst` on this"?

An alternative is to take a `Spec.Entities` here, which would enforce that there can't be unknowns.  This points at an important(ish) `subst` design decision:
- "Partial" substitution:
    - `subst` on `Partial.Value` returns `Partial.Value`, `subst` on `Partial.Entities` returns `Partial.Entities`, etc
    - Allows "incomplete" substitutions where you only substitute some of the unknowns
    - Substitution map can be `Unknown` to `Partial.Value`, meaning you could substitute in an expression containing `unknown`s
    - `Response.reEvaluateWithSubst` has to return another `Partial.Response`
- "Full" substitution":
    - `subst` on `Partial.Value` returns `Spec.Value`, `subst` on `Partial.Entities` returns `Spec.Entities`, etc
    - Does not allow "incomplete" substitutions where you only substitute some of the unknowns
    - Substitution map has to be `Unknown` to `Spec.Value`, so you can only substitute in concrete values, not expressions containing `unknown`s
    - `Response.reEvaluateWithSubst` can return `Option Spec.Response`

This PR implements "partial" substitution, which also matches the Rust behavior better -- in Rust, `PartialResponse.reauthorize()` returns `PartialResponse` not `Response`.  But, we could also do "full" substitution, either instead or in addition.  What do you think?

aaronjeline: I think we want partial for maximum flexibility

emina: What is the use case for partial substitutions?

cdisselkoen: The other point is that partial substitutions are more general, and we can probably implement full substitutions completely in terms of partial (plus additional preconditions like "the substitution map has a mapping for all the unknowns that appear" and "the substitution map's values contain no unknowns themselves").

cdisselkoen: I have probably >1000 lines of proof about the definition of `subst` in this PR already, if that makes a difference.

emina: > we can probably implement full substitutions completely in terms of partial 

We are going to need some form of full substitution definition (either on its own or as extra constraints on partial) to state soundness, right?

I don't object to continuing with the current definitions.  My main worry would be that they will make the soundness proof harder than it would be otherwise.  The internal codebase has to do something similar to this, and it's using full substitutions.  That said, the datastructures and notions of equivalence for that proof are more complicated.  So maybe a more complex notion of substitution isn't a concern here.  That's the main thing I'd watch out for.

PR #326, opened 2024-05-15
Add fuzz targets using arbitrary strings

john-h-kastner-aws:
Some of our fuzz target work by generating a structured policy, printing it to a string, and then testing properties of that string. This approach is limited because it will not test policy strings that are not generated by our policy printing function. We can avoid this by doing unstructured fuzzing.

To make this targets find interesting inputs in a tractable amount of time, you will need to seed the corpus with some valid cedar/schema files. For instance:
`mkdir -p ./fuzz/corpus/formatter && find ../cedar -name "*.cedar" -exec cp {} ./fuzz/corpus/formatter/ \;`

The `formatter-bytes` target would have found cedar-policy/cedar#862 and the `schema-roundtrip-bytes` found   cedar-policy/cedar#875, both only when seeded with a small corpus of valid files.


shaobo-he-aws:
It looks great to me. Can we update the READMEs with the instruction to add corpus?

john-h-kastner-aws:
I'll think more about where the initial corpus should come from. This style of fuzzing is unlikely to generate new features it did see in the original corpus, so it's important that the corpus covers many interesting classes of input to begin with.

khieta (approved):
Lgtm. Is there a way for us to pre-seed the corpus so that it works out of the box? Or provide a script that does the appropriate seeding?

I'm thinking we could use the [handwritten integration tests](https://github.com/cedar-policy/cedar-integration-tests/tree/main/tests) and policies from [tinytodo](https://github.com/cedar-policy/cedar-examples/tree/main/tinytodo) and/or other example applications.

aaronjeline (approved):
Looks good to me as long as Kesha's comments are addressed

shaobo-he-aws (approved):

PR #331, opened 2024-05-23
Revise schema equivalence checks

shaobo-he-aws:
*Issue #, if available:*

*Description of changes:*

We now consider trivial schemas to be equivalent.

### DIFF
--- a/cedar-drt/fuzz/src/schemas.rs
+++ b/cedar-drt/fuzz/src/schemas.rs
@@ -37,7 +37,12 @@ use cedar_policy_validator::{ActionType, ApplySpec, NamespaceDefinition, SchemaF
 /// used. Whether or not there are applicable resources is useless.
 ///
 pub fn equivalence_check(lhs: SchemaFragment, rhs: SchemaFragment) -> Result<(), String> {
-    if lhs.0.len() == rhs.0.len() {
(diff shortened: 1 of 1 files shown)


john-h-kastner-aws (approved):

cedar-drt/fuzz/src/schemas.rs
```diff
@@ -53,6 +58,18 @@ pub fn equivalence_check(lhs: SchemaFragment, rhs: SchemaFragment) -> Result<(),
     }
 }
 
+// A schema fragment is trivial if it is empty or it's the empty namespace with
+// all declarations being empty
+fn is_trivial(schema: &SchemaFragment) -> bool {
+    if schema.0.len() == 0 {
+        true
+    } else if let Some(def) = schema.0.get(&None) {
+        def.entity_types.is_empty() && def.common_types.is_empty() && def.common_types.is_empty()
```
khieta, line 67: `def.common_types.is_empty()` is repeated twice -- is one of them support to be `action_types`?

shaobo-he-aws: Good catch. Fixing.

shaobo-he-aws: Fixed.

khieta (approved):

PR #332, opened 2024-05-23
Revise model and proofs for the IPAddr extension.

emina:
This PR revises the model and relevant proofs for the IP Address extension type and functions.

The semantics of the operations is still the same, but the representation and code are more uniform.  

In particular:

* We use the `BitVec` type to uniformly model all parts of an IP Address.  The existing model uses a mix of data types (UInts and Fins), which led to non-uniform proof obligations (for example, in `Cedar.Data.Thm.LT`).

* We introduce a new `CIDR` structure that is parametric in address width.  The `IPNet` data wraps the new structure, and IP operations are defined as parametric functions over their underlying `CIDR` values.  One pleasant consequence of this design is that we no longer need to convert 32-bit addresses to 128 bits to operate on them.

### DIFF
--- a/cedar-lean/Cedar/Spec/Ext/IPAddr.lean
+++ b/cedar-lean/Cedar/Spec/Ext/IPAddr.lean
@@ -20,133 +20,119 @@ namespace Cedar.Spec.Ext
 
 namespace IPAddr
 
------ UInt128 -----
-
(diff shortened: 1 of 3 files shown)


cdisselkoen (approved):
LGTM

khieta (approved):

PR #333, opened 2024-05-24
split Thm/Partial/Evaluation/Basic into Thm/Partial/Evaluation/Props and Thm/Partial/Evaluation/WellFormed

cdisselkoen:
This file had two pretty different things going on, so it makes sense to make two more focused files instead, particularly as the Props section keeps growing (as it is anticipated to as more of the PE soundness proof gets checked in).


khieta (approved):
Seems fine.

PR #334, opened 2024-05-24
Proof that partial evaluation returns well-formed values

cdisselkoen:
_This PR depends on #333, which must be merged first._

The next major lemma for the PE soundness proof: a proof that partial evaluation always returns well-formed values, given well-formed inputs.

I believe this lemma does not hold for the definition of `WellFormed` currently on `main`, so this PR makes that definition stronger.  In particular, this PR makes the following changes to the definition of `WellFormed`:
- `Spec.Value` now requires not only that sets are `WellFormed` sets and records are `WellFormed` maps, but also that any values contained in sets/records are `WellFormed` themselves
- `Spec.Request` now requires not only that the `context` is a `WellFormed` map, but also that any values contained in the `context` are `WellFormed` themselves.  `WellFormed` for `Partial.Request` was already defined this way; this PR brings `Spec.Request` in line with that.
- `Partial.EntityData.WellFormed` (and thus `Partial.Entities.WellFormed`) now requires not only that `ancestors` is a `WellFormed` set and that `attrs` is a `WellFormed` map, but also that any values contained in `attrs` are `WellFormed` themselves.

The existing proof that partial authorization on concrete inputs is equivalent to concrete authorization, only requires that the request context is a `WellFormed` map, and does not require the new stronger definition (nor does it require that `entities` are `WellFormed` at all), so I updated the precondition for it to keep the assumptions minimal in that theorem.

### DIFF
--- a/cedar-lean/Cedar/Thm/Partial/Evaluation/WellFormed.lean
+++ b/cedar-lean/Cedar/Thm/Partial/Evaluation/WellFormed.lean
@@ -28,51 +28,53 @@ import Cedar.Thm.Data.Set
 
 namespace Cedar.Spec
 
-/--
-  We define `WellFormed` for `Value` in the obvious way
--/
-def Value.WellFormed (v : Value) : Prop :=
-  match v with
-  | .set s => s.WellFormed
-  | .record r => r.WellFormed
+open Cedar.Data
+
+/-- All `Prim`s are structurally WellFormed. -/
+def Prim.WellFormed : Spec.Prim → Prop
   | _ => true
 
-/--
-  `Request`s are `WellFormed` if the context is `WellFormed`
--/
-def Request.WellFormed (req : Request) : Prop :=
-  req.context.WellFormed
+/-- All `Ext`s are structurally WellFormed. -/
+def Ext.WellFormed : Ext → Prop
+  | .decimal _ => true
+  | .ipaddr _  => true
+
+def Value.WellFormed : Spec.Value → Prop
+  | .prim p => p.WellFormed
+  | .set s => s.WellFormed ∧ ∀ elt ∈ s, elt.WellFormed
+  | .record r => r.WellFormed ∧ ∀ kv ∈ r.kvs, kv.snd.WellFormed
+  | .ext x => x.WellFormed
+decreasing_by
+  all_goals simp_wf
+  case _ h₁ => -- set
+    have := Set.sizeOf_lt_of_mem h₁
+    omega
+  case _ h₁ => -- record
+    have h₂ := Map.sizeOf_lt_of_value h₁
+    apply Nat.lt_trans h₂
+    have h₃ := Map.sizeOf_lt_of_kvs r
+    simp [Map.kvs] at *
+    omega
+
+def Request.WellFormed (req : Spec.Request) : Prop :=
+  req.context.WellFormed ∧ ∀ val ∈ req.context.values, val.WellFormed
 
 end Cedar.Spec
 
 namespace Cedar.Partial
 
-/--
-  We define `WellFormed` for `Partial.Value` using `Spec.Value.WellFormed`
--/
-def Value.WellFormed (pval : Partial.Value) : Prop :=
-  match pval with
+def Value.WellFormed : Partial.Value → Prop
   | .value v => v.WellFormed
   | .residual _ => true
 
-/--
-  `Partial.Request`s are `AllWellFormed` if the context is `WellFormed` and
-  all the context's constituent `Partial.RestrictedValue`s are also `WellFormed`.
-  (principal, action, and resource are always well-formed)
--/
-def Request.AllWellFormed (preq : Partial.Request) : Prop :=
-  preq.context.WellFormed ∧ ∀ rpval ∈ preq.context.values, rpval.WellFormed
+def Request.WellFormed (preq : Partial.Request) : Prop :=
+  preq.context.WellFormed ∧ ∀ pval ∈ preq.context.values, pval.WellFormed
 
-/--
-  We define `WellFormed` for `Partial.EntityData` in the obvious way
--/
 def EntityData.WellFormed (edata : Partial.EntityData) : Prop :=
-  edata.attrs.WellFormed ∧ edata.ancestors.WellFormed
+  edata.attrs.WellFormed ∧
+  edata.ancestors.WellFormed ∧
+  ∀ pval ∈ edata.attrs.values, pval.WellFormed
 
-/--
-  `Partial.Entities` are `AllWellFormed` if they are `WellFormed` and all the
-  constituent `Partial.EntityData` are also `WellFormed`
--/
 def Entities.AllWellFormed (entities : Partial.Entities) : Prop :=
   entities.WellFormed ∧ ∀ edata ∈ entities.values, edata.WellFormed
 
--- a/cedar-lean/Cedar/Thm/Partial/Evaluation/WellFormed.lean
+++ b/cedar-lean/Cedar/Thm/Partial/Evaluation/WellFormed.lean
@@ -1,0 +1,79 @@
+/-
+ Copyright Cedar Contributors
+
+ Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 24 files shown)


cedar-lean/Cedar/Thm/Partial/Evaluation/WellFormed.lean
```diff
@@ -28,51 +28,53 @@ import Cedar.Thm.Data.Set
...
-  `Request`s are `WellFormed` if the context is `WellFormed`
--/
-def Request.WellFormed (req : Request) : Prop :=
-  req.context.WellFormed
+/-- All `Ext`s are structurally WellFormed. -/
+def Ext.WellFormed : Ext → Prop
+  | .decimal _ => true
+  | .ipaddr _  => true
+
+def Value.WellFormed : Spec.Value → Prop
```
emina, line 42: Not a blocker to merging the PR, but you may find it easier to do proofs over an inductive predicate that defines well-formedness rather than an explicit recursive function. Something like [this](https://github.com/cedar-policy/cedar-spec/blob/57f3da8af7c42c16529934f601f72fbf7dba6eb3/cedar-lean/Cedar/Thm/Validation/Typechecker/Types.lean#L49).

The inductive predicate approach doesn't require explicit termination arguments in the definition, and you can do induction directly on an assumption that says an argument is well-formed.

emina (approved):
LGTM

cedar-lean/Cedar/Thm/Partial/Evaluation/WellFormed.lean
cdisselkoen: I'm trying this, but finding it less convenient for proofs about well-formedness (could be user error or inexperience).  For instance, to discharge an obligation `Partial.Value.WellFormed (Partial.Value.value (Spec.Value.record m))`, previously I could use `simp [Partial.Value.WellFormed, Spec.Value.WellFormed]` to convert this to a pair of obligations `m.WellFormed` and `∀ v, ...`.  With the `inductive` definition of `Partial.Value.WellFormed` and `Spec.Value.WellFormed`, instead of one `simp` I seem to now need two `apply`s to accomplish the same thing: `apply Partial.Value.WellFormed.wf_val (.record m)` followed by `apply Spec.Value.WellFormed.wf_record m`.

emina: > I seem to now need two `apply`s to accomplish the same thing: `apply Partial.Value.WellFormed.wf_val (.record m)` followed by `apply Spec.Value.WellFormed.wf_record m`.

I like to use one `exact` for these; something like `exact Partial.Value.WellFormed.wf_val (.record m) (Spec.Value.WellFormed.wf_record m)`.

cdisselkoen: Has to be one `apply` in my case (as I want to leave one obligation as a goal) and it's a lot uglier than the original `simp`, but maybe it's more precise / less fragile in some sense.

cdisselkoen: I'm more concerned about the second problem.  One thing that is fundamentally different about the `inductive` definition is that we could have multiple arms of `Partial.Value.WellFormed` that apply to the same arm of `Partial.Value`.  That generality is powerful, but not needed here, and one negative consequence of it is that when Lean sees `Partial.Value.WellFormed (Partial.Value.value v)` it doesn't immediately know which arm of `Partial.Value.WellFormed` to use, as it does with the `def`.  That prevents the `simp` from working.  I don't know how to tell Lean to use a specific arm of `Partial.Value.WellFormed` inside the consequent of a hypothesis like this.

emina: > Later in the same proof, I have a hypothesis `Partial.Value.value v ∈ Map.values request.context → Partial.Value.WellFormed (Partial.Value.value v)`. Previously, I could use `simp [Partial.Value.WellFormed] at h` to simplify this to `Partial.Value.value v ∈ Map.values request.context → Spec.Value.WellFormed v`. It's unclear how I can accomplish the same transformation with `Partial.Value.WellFormed` defined as an `inductive`. By which I mean, probably there is a way but I don't know the right tactic?

I haven't looked at that particular proof in detail, but for hypotheses that look like this, I get rid of the implication first by applying `h` to the antecedent.  When left with `h : Partial.Value.WellFormed (Partial.Value.value v)`.  Then, if you do `cases h`, Lean will deconstruct `h` similarly to how it deconstructs an inductive data type.

Alternatively, if you prefer to keep the implication, you can a prove lemma that says `Partial.Value.WellFormed (Partial.Value.value v) → Spec.Value.WellFormed v` and use those with `rw`.  These proofs showing what's implied by an inductive predicate are quick to do with `cases`.

In general, inductive predicates are very nice to work with if you need to perform induction on a value/expr/etc that's WellFormed.  See, e.g., [here](https://github.com/leanprover-community/batteries/blob/60d622c124cebcecc000853cdae93f4251f4beb5/Batteries/Data/RBMap/WF.lean#L507-L511).  Another benefit is that they are idiomatic :)  For example, the standard library uses them consistently for predicate definitions over inductive types (e.g., [RBNode.WF](https://github.com/leanprover-community/batteries/blob/60d622c124cebcecc000853cdae93f4251f4beb5/Batteries/Data/RBMap/Basic.lean#L602-L611), [List.Forall_2](https://leanprover-community.github.io/mathlib4_docs/Batteries/Data/List/Basic.html#List.Forall%E2%82%82), [List.perm](https://github.com/leanprover-community/batteries/blob/60d622c124cebcecc000853cdae93f4251f4beb5/Batteries/Data/List/Basic.lean#L1349-L1357), etc).

Again, not a merge blocker by any means.  We can change this later if we want to.

cdisselkoen: In this case I need to keep the implication because my goal is `Spec.Value.WellFormed v` and I want to apply this hypothesis.

Noted about using `rw` and a lemma.  And noted about not being a merge blocker.  I'll work on this for a subsequent PR.

cedar-lean/Cedar/Thm/Partial/Evaluation/WellFormed.lean
```diff
@@ -28,51 +28,53 @@ import Cedar.Thm.Data.Set
...
+  edata.attrs.WellFormed ∧
+  edata.ancestors.WellFormed ∧
+  ∀ pval ∈ edata.attrs.values, pval.WellFormed
 
-/--
-  `Partial.Entities` are `AllWellFormed` if they are `WellFormed` and all the
-  constituent `Partial.EntityData` are also `WellFormed`
--/
 def Entities.AllWellFormed (entities : Partial.Entities) : Prop :=
   entities.WellFormed ∧ ∀ edata ∈ entities.values, edata.WellFormed
```
khieta, line 79: Nit: it would feel more consistent if this predicate was called `WellFormed` instead of `AllWellFormed`. I guess this would require changes the definition of `PartialEntities` -- would that break everything?

khieta (approved):
Lgtm.

Maybe you could consider putting the predicate `∀ pval, Partial.evaluate _ _ _ = .ok pval → pval.WellFormed` with the rest of the predicates in `Props.lean`. It's not critical since this predicate is easy to inline, I just like to have all the definitions in one place 🙂

cedar-lean/Cedar/Thm/Partial/Evaluation/WellFormed.lean
cdisselkoen: Yeah, since `Partial.Entities` is just an alias for a particular `Map` type, `Partial.Entities.WellFormed` is already defined and means just that the `Map` is `Map.WellFormed`.  If we made it a struct and not an alias we could reclaim the `WellFormed` name, and probably we should

cdisselkoen: Did the replacement, no more `AllWellFormed`

khieta (approved):

PR #335, opened 2024-05-24
Some style improvements/fixups in the Lean

cdisselkoen:
A bunch of minor changes collapsed into one PR (happy to split out any of these bullets into separate PRs if desired):

- In a bunch of theorems, make arguments optional if they are implied by the theorem's hypotheses
    - This change not applied in `Thm/Data` because of the breaking changes for callers; we can do that in a separate PR
- Conversely, in a couple theorems, make arguments required if they are not implied by the theorem's hypotheses
    - This change also not applied in `Thm/Data`
- don't open `Decision` or `Effect` in `Authorization.lean`  (it's easy and clear to just write `.deny` and `.forbid` instead of `deny` and `forbid`)
- use proper subscripts in hypothesis names, not `h0` etc
- In one place, avoid duplicating a lemma, instead calling out to the copy in `Thm/Data/List`
- linebreak a few long lines
- make a couple proofs prettier
- when invoking lemmas in `Thm/Partial/Subst`, call them with just `Subst.blah` instead of `Partial.Subst.blah`.  Substitution is a concept unique to partial evaluation, so it's clear we're dealing with partial evaluation.  By contrast, when invoking lemmas in `Thm/Partial/Evaluation`, I will continue to call them with `Partial.Evaluation.blah` instead of `Evaluation.blah` (even when Lean would accept the latter) to make it clear whether we're invoking a theorem about partial evaluation or concrete evaluation.


khieta (approved):
Seems fine to me.

emina (approved):
LGTM

PR #343, opened 2024-05-29
Add target for json-to-cedar policy conversion

john-h-kastner-aws:
Given a JSON formatted policy, check that, if we can parse it to an AST then we can round-trip that AST through the human-readable format and get back the same policy. Finds cedar-policy/cedar#925

There aren't actually any `*.cedar.json` files in the repo currently. I opened https://github.com/cedar-policy/cedar/pull/926 to add one example, but that's a very small corpus. We'll need to add some more or seed this target another way. Possibly by implementing cedar-policy/cedar#461 and converting existing `*.cedar` files to JSON.

### DIFF
--- a/cedar-drt/initialize_corpus.sh
+++ b/cedar-drt/initialize_corpus.sh
@@ -33,7 +33,7 @@ print_usage() {
 }
 
 if [ $# -ne 1 ]; then
-    echo "Expected exactly one arguemtn, but $# arguments provided."
+    echo "Expected exactly one argument, but $# arguments provided."
     print_usage
 fi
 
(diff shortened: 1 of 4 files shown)


shaobo-he-aws (approved):

cedar-drt/initialize_corpus.sh
```diff
@@ -47,6 +47,11 @@ initialize_corpus() {
       cp "$f" "$corpus_dir/$(uuidgen)-$(basename "$f")"
     done ;;
 
+  convert-policy-json-to-cedar)
```
khieta, line 50: Not the fault of this PR, but I noticed a typo "arguemtn" in this file on L36 - can you fix in this PR?

khieta (approved):

khieta (approved):

PR #350, opened 2024-06-06
Update to Lean 4.8.0 and Batteries.

emina:
This PR updates the code and proofs to use Lean 4.8.0 and the `Batteries` (formerly `Std`) library.


andrewmwells-amazon (approved):

emina:
> Build issues can probably be fixed by adding (or replacing) `Batteries:static` everywhere there's a `Std:static` in the build scripts.

Yes, that fixed some issues (and a similar change to lean_impl.rs) but the build is still failing.  It looks like the static `Batteries` libraries is either not being built or not being found.

andrewmwells-amazon (approved):

cdisselkoen (approved):

PR #359, opened 2024-06-13
Move DRT Lean library build command into bash script

andrewmwells-amazon:
*Issue #, if available:*

*Description of changes:*
Move DRT Lean library build command into bash script

I think this will be the cleanest way to fix https://github.com/cedar-policy/cedar/actions/runs/9506669677/job/26204425743#step:11:9

We'd like to keep the same CI .yaml files for each repo (so the `main` branch) so we can make changes in one place. In this case, the build steps have to change because Lean renamed `Std` -> `Batteries`, so each branch will have this script which hopefully won't change again rather than changing the yaml files.


Swolebrain (approved):

shaobo-he-aws (approved):

shaobo-he-aws:
It seems we need to add execution permission to this file.

shaobo-he-aws (approved):

khieta (approved):

aaronjeline (approved):

PR #361, opened 2024-06-14
Adds modelling of condition clauses

aaronjeline:
*Issue #, if available:*
N/A
*Description of changes:*
The lean model doesn't capture the details of `when`/`unless` clauses (Including order of evaluation/short-circuiting). 
This is a miss as it's something implementations will need to get correct. 
This PR adds `when`/`unless` clauses and the proper algorithm for transforming them into expressions.
The JSON format we currently use to test the Rust impl does not support `when`/`unless` clauses (and can't be adapted too, as it's the serde serialization of a struct that has already lost that information). So we just interpret the expression past from Rust as a single `when` clause. When we switch to using ESTs for this testing, we will get the full structure.

### DIFF
--- a/cedar-lean/Cedar/Spec/Policy.lean
+++ b/cedar-lean/Cedar/Spec/Policy.lean
@@ -55,8 +55,8 @@ inductive ActionScope where
 abbrev PolicyID := String
 
 inductive ConditionKind where
-  | When
-  | Unless
+  | when
+  | unless
 
 structure Condition where
   kind : ConditionKind
@@ -112,8 +112,8 @@ def ActionScope.toExpr : ActionScope → Expr
 
 def Condition.toExpr (c : Condition) : Expr :=
   match c.kind with
-  | .When => c.body
-  | .Unless => Expr.unaryApp .not c.body
+  | .when => c.body
+  | .unless => Expr.unaryApp .not c.body
 
 -- Conditions are evaluated top to bottom, and short circuit
 def conditionsToExpr (cs : Conditions) : Expr :=
--- a/cedar-lean/Cedar/Spec/Policy.lean
+++ b/cedar-lean/Cedar/Spec/Policy.lean
@@ -54,13 +54,24 @@ inductive ActionScope where
 
 abbrev PolicyID := String
 
+inductive ConditionKind where
+  | When
+  | Unless
+
+structure Condition where
+  kind : ConditionKind
+  body : Expr
+
+abbrev Conditions := List Condition
+
 structure Policy where
   id : PolicyID
   effect : Effect
   principalScope : PrincipalScope
   actionScope : ActionScope
   resourceScope : ResourceScope
-  condition : Expr
+  condition : Conditions
+
 
 abbrev Policies := List Policy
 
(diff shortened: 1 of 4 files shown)


cedar-lean/Cedar/Spec/Policy.lean
```diff
@@ -99,14 +110,23 @@ def ActionScope.toExpr : ActionScope → Expr
     let exprs := es.map (fun e => .lit (.entityUID e))
     .binaryApp (.mem) (.var .action) (.set exprs)
 
+def Condition.toExpr (c : Condition) : Expr :=
+  match c.kind with
+  | .when => c.body
+  | .unless => Expr.unaryApp .not c.body
+
+-- Conditions are evaluated top to bottom, and short circuit
+def conditionsToExpr (cs : Conditions) : Expr :=
```
emina, line 119: We can write this using typeclass inference to avoid some wrapping.  Writing `Conditions.toExpr` means we can also write `cs.toExpr` in the client code when Lean knows that `cs` is of type `Conditions` (see below):

```suggestion
def Conditions.toExpr (cs : Conditions) : Expr :=
  cs.foldr (λ c expr => .and c.toExpr expr) true
```

cedar-lean/Cedar/Spec/Policy.lean
```diff
@@ -99,14 +110,23 @@ def ActionScope.toExpr : ActionScope → Expr
...
+
 def Policy.toExpr (p : Policy) : Expr :=
   .and
     p.principalScope.toExpr
     (.and
       p.actionScope.toExpr
       (.and
         p.resourceScope.toExpr
-        p.condition))
+        (conditionsToExpr p.condition)))
```
emina, line 129: Using the `dot` notation with the above change to save some typing :):
```suggestion
        p.condition.toExpr))
```

cedar-lean/Cedar/Thm/Authorization/Authorizer.lean
```diff
@@ -541,7 +541,7 @@ theorem error_implies_scope_satisfied {policy : Policy} {request : Request} {ent
       exfalso
       clear h₂
       unfold producesNonBool at h₅
-      have h₄ := and_produces_bool_or_error policy.resourceScope.toExpr policy.condition request entities
+      have h₄ := and_produces_bool_or_error policy.resourceScope.toExpr (conditionsToExpr policy.condition) request entities
```
emina, line 544:
```suggestion
      have h₄ := and_produces_bool_or_error policy.resourceScope.toExpr policy.condition.toExpr request entities
```

cedar-lean/Cedar/Thm/Authorization/Authorizer.lean
```diff
@@ -554,7 +554,7 @@ theorem error_implies_scope_satisfied {policy : Policy} {request : Request} {ent
     -- in this case, evaluating (action ∧ resource ∧ condition) produced a non-boolean
     exfalso
     unfold producesNonBool at h₄
-    generalize (Expr.and policy.resourceScope.toExpr policy.condition) = resource_and_condition at h₂ h₄
+    generalize (Expr.and policy.resourceScope.toExpr (conditionsToExpr policy.condition)) = resource_and_condition at h₂ h₄
```
emina, line 557:
```suggestion
    generalize (Expr.and policy.resourceScope.toExpr policy.condition.toExpr) = resource_and_condition at h₂ h₄
```

emina (approved):

emina (approved):

cdisselkoen (approved):

PR #362, opened 2024-06-14
change `Subsmap` to map to `Partial.Expr`

cdisselkoen:
Makes the model more general, and works with the upcoming PE soundness proof.


aaronjeline (approved):
Seems very reasonable

emina (approved):

PR #365, opened 2024-06-17
Add more datastructure lemmas

cdisselkoen:
A number of these are analogues of existing lemmas, that we were just missing.


emina (approved):
LGTM.

khieta (approved):
LGTM

PR #366, opened 2024-06-18
Getting rid of unspecified

aaronjeline:
*Issue #, if available:*
N/A
*Description of changes:*
Implementation of [RFC 55](https://github.com/cedar-policy/rfcs/blob/main/text/0055-remove-unspecified.md)

### DIFF
--- a/cedar-drt/create_corpus.sh
+++ b/cedar-drt/create_corpus.sh
@@ -14,14 +14,21 @@
 # limitations under the License.
 
 FUZZ_TARGET="${FUZZ_TARGET:-abac}"
-TIMEOUT="${TIMEOUT:-15m}" # 15m = 900s
+TIMEOUT_MINS="${TIMEOUT:-15}" # 15m = 900s
 JOBS="${JOBS:-4}"
-DUMP_DIR="${DUMP_DIR:-corpus_tests}"
+DUMP_DIR="${DUMP_DIR:-corpus-tests}"
 
 # Assumes cedar-drt is already correctly built and configured
 rm -rf $DUMP_DIR
 source set_env_vars.sh
 export FUZZ_TARGET
 export DUMP_DIR
-timeout $TIMEOUT cargo fuzz run -s none $FUZZ_TARGET -j $JOBS -- -rss_limit_mb=8196
+echo "Running fuzzer for $TIMEOUT_MINS minutes ..."
+TIMEOUT=$(( $TIMEOUT_MINS * 60 ))
+cargo fuzz run -s none $FUZZ_TARGET -j $JOBS -- -rss_limit_mb=8196 -max_total_time=$TIMEOUT
+echo "Fuzzer done!"
+# Minimize corpus to remove redundant seeds
+echo "Minimizing corpus..."
+cargo fuzz cmin "$FUZZ_TARGET" -s none
+echo "Minimization done"
 ./synthesize_tests.sh
--- a/cedar-drt/fuzz/src/schemas.rs
+++ b/cedar-drt/fuzz/src/schemas.rs
@@ -104,7 +104,6 @@ fn action_type_equivalence(name: &str, lhs: ActionType, rhs: ActionType) -> Resu
             (None, None) => Ok(()),
             (Some(lhs), Some(rhs)) => {
                 // If either of them has at least one empty appliesTo list, the other must have the same attribute.
-                // Otherwise both of them must apply to unspecified entities or non-empty entity lists, which must be equal.
                 if rhs == lhs {
                     Ok(())
                 } else {
--- a/cedar-drt/fuzz/src/schemas.rs
+++ b/cedar-drt/fuzz/src/schemas.rs
@@ -16,7 +16,7 @@
 
 use std::collections::{HashMap, HashSet};
 
-use cedar_policy_validator::{ActionType, ApplySpec, NamespaceDefinition, SchemaFragment};
+use cedar_policy_validator::{ActionType, NamespaceDefinition, SchemaFragment};
 
 /// Check if two schema fragments are equivalent, modulo empty apply specs.
 /// We do this because there are schemas that are representable in the JSON that are not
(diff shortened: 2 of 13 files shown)


shaobo-he-aws (approved):
It feels so good to see them removed 😄

shaobo-he-aws (approved):

cedar-drt/create_corpus.sh
john-h-kastner-aws, line 1: What are these changes?

cedar-drt/fuzz/src/schemas.rs
```diff
@@ -105,7 +105,7 @@ fn action_type_equivalence(name: &str, lhs: ActionType, rhs: ActionType) -> Resu
             (Some(lhs), Some(rhs)) => {
                 // If either of them has at least one empty appliesTo list, the other must have the same attribute.
                 // Otherwise both of them must apply to unspecified entities or non-empty entity lists, which must be equal.
-                if (either_empty(&lhs) && either_empty(&rhs)) || rhs == lhs {
+                if rhs == lhs {
```
khieta, line 108: I imagine this comment needs to be updated if this guard is removed. Likely some other comments in this file need updating too.

Also: `either_empty` is checking for empty lists, which are used for action groups -- why do these checks need to be removed for RFC55?

cedar-drt/create_corpus.sh
aaronjeline: Using libfuzzers timeout mechanism and a tweak kesha asked me to add

cedar-drt/synthesize_tests.sh
```diff
@@ -22,8 +22,6 @@ source set_env_vars.sh
 
 set -u
 
-# Minimize corpus to remove redundant seeds
-cargo fuzz cmin "$FUZZ_TARGET" -s none
```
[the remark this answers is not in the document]
aaronjeline: Kesha asked me to include while I was writing this PR

cedar-policy-generators/src/expr.rs
```diff
@@ -566,7 +566,7 @@ impl<'a> ExprGenerator<'a> {
                                             // This does not use an explicit namespace because entity types
                                             // implicitly use the schema namespace if an explicit one is not
                                             // provided.
-                                            name: entity_name.clone().into(),
+                                            name: ast::Name::from(entity_name.clone()).into(),
```
khieta, line 569: `ast::Name::from(entity_name.clone()).into()` is used a couple places in this PR. Without having thought about it too much -- is there a way to do this without cloning? What is it converting _into_?

cedar-policy-generators/src/schema.rs
```diff
@@ -820,49 +816,34 @@ impl Schema {
...
-                            if principal_and_resource_types_exist {
-                                if u.ratio(1, 8)? {
-                                    picked_resource_types = None;
-                                }
-                                if u.ratio(1, 8)? {
-                                    picked_principal_types = None;
-                                }
-                            } else {
-                                principal_and_resource_types_exist = true;
-                            }
```
khieta, line 859: Remove comment since you're removing this code. But also: why remove this code?

cedar-policy-generators/src/expr.rs
shaobo-he-aws: Cloning a `Name` is actually very cheap.

cedar-policy-generators/src/schema.rs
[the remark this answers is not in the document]
vasumv: yes I think we still need this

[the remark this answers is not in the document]
aaronjeline: Fixed it so the generator will sometimes make lists of length 0

aaronjeline: `principal` and `resource` can't be `None` anymore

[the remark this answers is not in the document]
vasumv: Actually, I explicitly made it so that the generator will create lists of minimum 1, and then later on decide whether to empty the list. I think we shouldn't initially generate lists of 0, since that's what caused the validity issue in the first place (no resource / principal types in all the actions). 

Would it be possible keep the minimum 1, and the subsequent coin flip to empty `picked_resource_types` and `picked_principal_types`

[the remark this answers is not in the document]
aaronjeline: Yeah sure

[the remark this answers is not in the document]
vasumv: thanks! I think you should also still keep the `principal_and_resource_types_exist` check so that it only deletes if we have at least one valid action.

cedar-drt/create_corpus.sh
```diff
@@ -14,14 +14,21 @@
 # limitations under the License.
 
 FUZZ_TARGET="${FUZZ_TARGET:-abac}"
-TIMEOUT="${TIMEOUT:-15m}" # 15m = 900s
+TIMEOUT_MINS="${TIMEOUT:-15}" # 15m = 900s
```
andrewmwells-amazon, line 17:
```suggestion
TIMEOUT_MINS="${TIMEOUT:-15}"
```

cedar-drt/create_corpus.sh
```diff
@@ -14,14 +14,21 @@
...
 # Assumes cedar-drt is already correctly built and configured
 rm -rf $DUMP_DIR
 source set_env_vars.sh
 export FUZZ_TARGET
 export DUMP_DIR
-timeout $TIMEOUT cargo fuzz run -s none $FUZZ_TARGET -j $JOBS -- -rss_limit_mb=8196
-./synthesize_tests.sh
\ No newline at end of file
+echo "Running fuzzer for $TIMEOUT_MINS minutes ..."
+TIMEOUT=$(( $TIMEOUT_MINS * 60 ))
```
andrewmwells-amazon, line 27:
```suggestion
TIMEOUT_S=$(( $TIMEOUT_MINS * 60 ))
```

cedar-drt/create_corpus.sh
```diff
@@ -14,14 +14,21 @@
...
+cargo fuzz run -s none $FUZZ_TARGET -j $JOBS -- -rss_limit_mb=8196 -max_total_time=$TIMEOUT
```
andrewmwells-amazon, line 28:
```suggestion
cargo fuzz run -s none $FUZZ_TARGET -j $JOBS -- -rss_limit_mb=8196 -max_total_time=$TIMEOUT_S
```

andrewmwells-amazon (approved):

PR #368, opened 2024-06-18
Proof that subst preserves errors

cdisselkoen:
The next lemma required for PE soundness. We already have a lemma stating that "if partial evaluation returns a concrete value, then it returns the same concrete value after any substitution of unknowns [in the original expression/request/entity store]".  This is the counterpart for when partial evaluation returns an error.

Notably, the statement is not that partial evaluation must return _the same_ error after any substitution of unknowns; that doesn't hold (one of many counterexamples: `principal.foo + false` where the principal is initially unknown, but the substitution maps it to some entity which doesn't have a `foo` attribute. This gives a type error before substitution, but an attrDoesNotExist error after substitution.)  Instead, the statement is that partial evaluation must return _some_ error (not necessarily the same error) after any substitution of unknowns.

As of this writing, I have not yet normalized the `simp`s in this PR.


cedar-lean/lake-manifest.json
```diff
@@ -4,10 +4,10 @@
  [{"url": "https://github.com/leanprover-community/batteries",
    "type": "git",
    "subDir": null,
-   "rev": "551ff2d7dffd7af914cdbd01abbd449fe3e3d428",
+   "rev": "51e6e0d24db9341fb031288c298b7e6b56102253",
    "name": "batteries",
    "manifestFile": "lake-manifest.json",
-   "inputRev": "main",
+   "inputRev": "v4.8.0",
```
khieta, line 10: Needed for this PR?

cedar-lean/Cedar/Thm/Partial/Evaluation/AndOr.lean
```diff
@@ -152,4 +152,52 @@ theorem subst_preserves_evaluation_to_value {x₁ x₂ : Partial.Expr} {req req'
...
+  Inductive argument that if partial-evaluation of a `Partial.Expr.and` or
+  `Partial.Expr.or` returns an error, then it also returns an error (not
+  necessarily the same error) after any substitution of unknowns
+
+  The proof of `subst_preserves_evaluation_to_value` for this
+  request/entities/subsmap is passed in as an argument, because this file can't
+  import `Thm/Partial/Evaluation.lean` to access it.
+  Alternately, this entire inductive proof could live in its own set of files,
+  all of which could depend on `Thm/Partial/Evaluation.lean` and its theorems
+  like `subst_preserves_evaluation_to_value`.
```
khieta, line 165: The current design seems fine to me. Thanks for leaving a comment to avoid future confusion. Though I think it would be cleaner to have this "Alternatively..." text in an issue rather than in the code.

cedar-lean/Cedar/Thm/Partial/Evaluation.lean
```diff
@@ -444,3 +444,114 @@ theorem subst_preserves_evaluation_to_value {expr : Partial.Expr} {req req' : Pa
...
+      exact subst_preserves_errors wf_r wf_e h_req
+    }
+termination_by expr
+
+/--
+  Corollary (contrapositive) of the above:
+  If partial evaluation returns ok after any substitution of unknowns,
+  then it must return ok before that substitution
+-/
+theorem subst_preserves_errors_mt {expr : Partial.Expr} {req req' : Partial.Request} {entities : Partial.Entities} {subsmap : Subsmap}
```
khieta, line 544: Why `_mt`?

khieta (approved):
Overall structure and theorem statement seems fine to me. Approving as-is, though I imagine you'll still do some hacking on the proofs (e.g., normalizing `simp`s)

cedar-lean/Cedar/Thm/Partial/Evaluation/AndOr.lean
cdisselkoen: I'd like to only make the issue if we think we actually want to move to that design.

cedar-lean/Cedar/Thm/Partial/Evaluation.lean
cdisselkoen: From the Latin "modus tollens", which in formal logic is the rule for reducing contrapositives.  I named this in analogue to Lean's use of `mp` (from the Latin "modus ponens") to mean the forward direction of an implication.  Suggestions for a different name? I suppose `_contrapositive`?

cedar-lean/lake-manifest.json
cdisselkoen: Fixed

cedar-lean/Cedar/Thm/Partial/Evaluation/AndOr.lean
cdisselkoen: Having thought about it some more, and especially seeing as this pattern will come up a couple more times as I upstream the rest of the PE soundness proof, I think the module-per-toplevel-inductive-proof organization probably makes the most sense.  I'll make the issue.

cdisselkoen: #372

cdisselkoen: Also replaced the "Alternately" text with references to #372.

emina (approved):
Looks good to me.

PR #371, opened 2024-06-21
Comparison fix

aaronjeline:
*Issue #, if available:*
N/A
*Description of changes:*
Fixes a small mistake in the comparison functions between schemas.

### DIFF
--- a/cedar-drt/fuzz/src/schemas.rs
+++ b/cedar-drt/fuzz/src/schemas.rs
@@ -113,6 +113,12 @@ fn action_type_equivalence(name: &str, lhs: ActionType, rhs: ActionType) -> Resu
                     ))
                 }
             }
+            // An action w/ empty applies to list is equivalent to an action with _no_ applies to
+            // section at all.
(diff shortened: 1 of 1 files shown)


john-h-kastner-aws (approved):

cedar-drt/fuzz/src/schemas.rs
```diff
@@ -113,6 +113,12 @@ fn action_type_equivalence(name: &str, lhs: ActionType, rhs: ActionType) -> Resu
                     ))
                 }
             }
+            // An action w/ empty applies to list is equivalent to an action with _no_ applies to
+            // section at all.
+            // This is because neither action can be legally applied to any principal/resources.
+            (Some(applies_to), None) | (None, Some(applies_to)) if both_empty(&applies_to) => {
```
khieta, line 119: I think you want an `either_empty` function here, because if either list is empty then the action applies to nothing.

aaronjeline: Oh yeah you're right

khieta (approved):

PR #375, opened 2024-06-26
More datastructure lemmas.

emina:
More utility lemmas about lists and sets.

### DIFF
--- a/cedar-lean/Cedar/Thm/Data/List/Basic.lean
+++ b/cedar-lean/Cedar/Thm/Data/List/Basic.lean
@@ -77,6 +77,21 @@ theorem cons_equiv_cons (x : α) (xs ys : List α) :
     apply List.cons_subset_cons; assumption
   }
 
+theorem cons_equiv_implies_equiv (x : α) (xs ys : List α) :
+  x :: xs ≡ x :: ys → x ∉ xs → x ∉ ys → xs ≡ ys
+:= by
+  simp [List.Equiv, List.subset_def]
+  intro h₁ h₂ _ _
+  constructor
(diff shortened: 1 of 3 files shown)


cedar-lean/Cedar/Thm/Data/List/Basic.lean
```diff
@@ -77,6 +77,21 @@ theorem cons_equiv_cons (x : α) (xs ys : List α) :
     apply List.cons_subset_cons; assumption
   }
 
+theorem cons_equiv_implies_equiv (x : α) (xs ys : List α) :
+  x :: xs ≡ x :: ys → x ∉ xs → x ∉ ys → xs ≡ ys
+:= by
+  simp [List.Equiv, List.subset_def]
+  intro h₁ h₂ _ _
+  constructor
+  case' left  => have h₃ := h₁
```
cdisselkoen, line 86: what is `case'`?

cdisselkoen (approved):
Looks great

cedar-lean/Cedar/Thm/Data/List/Basic.lean
emina: It's a variant of `case` that lets you work on a goal without having to close it (as required by `case`).  It's convenient if you have a bunch of similar proof contexts with a few small differences.  You can work on each separately with `case'` to get them into a common shape, then discharge them all together with `all_goals`.

cdisselkoen: Wow, I needed this a long time ago :) Awesome!

aaronjeline (approved):

PR #377, opened 2024-06-28
backport generator improvement

john-h-kastner-aws:
Backport Vasu's improvements from #369 to 2.4.x. We still test that version, so there's some benefit there, but mainly I'm using this version internally.

### DIFF
--- a/cedar-policy-generators/src/schema.rs
+++ b/cedar-policy-generators/src/schema.rs
@@ -35,6 +35,7 @@ use cedar_policy_validator::{
     TypeOfAttribute, ValidatorSchema,
 };
 use smol_str::SmolStr;
+use smol_str::ToSmolStr;
 use std::collections::BTreeMap;
 use std::str::FromStr;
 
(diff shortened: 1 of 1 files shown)


cedar-policy-generators/src/schema.rs
```diff
@@ -637,82 +638,64 @@ impl Schema {
...
-                                        return Err(Error::NoValidPrincipalOrResourceTypes);
-                                    }
+                            let mut picked_resource_types =
+                                pick_entity_types(&mut resource_types, u)?;
+                            let mut picked_principal_types =
+                                pick_entity_types(&mut principal_types, u)?;
+                            // If we already have resource_types and principal_types, randomly make them empty
+                            if principal_and_resource_types_exist {
+                                if u.ratio(1, 8)? {
+                                    picked_resource_types = None;
```
vasumv, line 683: can we change this to `principal_resource_types.clear()`? @aaronjeline added this fix when unsupported was removed. I don't think it makes a difference but looks a bit cleaner.

vasumv (approved):

cedar-policy-generators/src/schema.rs
john-h-kastner-aws: I think this would change what exactly the generator is doing. After aarons change, this can't be `None` so it we make it empty instead. Prior to that change, `None` and empty are different.

vasumv: Ah right, depends on whether you want to test unsupported or not. I'm guessing we should do both, maybe we can add two more coin flips for empty then?

khieta (approved):

PR #385, opened 2024-07-09
Ensures `Cedar.Thm.Partial.Expr` is actually built

aaronjeline:
*Issue #, if available:*
N/A
*Description of changes:*
`Cedar.Thm.Partial.Expr` wasn't actually being built. This caused some drift where it no longer actually verified.
Both are fixed, and updates batteries to new rev while I was here.


khieta (approved):
Changes look fine, but heads up that `Cedar.Thm.Partial.Expr` is proposed to be deleted in #379. You should look at that PR if you haven't already.

aaronjeline:
No longer needed due to a refactoring

PR #396, opened 2024-07-17
Handle `None`/empty equivlance for action in schema

john-h-kastner-aws:
*Issue #, if available:*

Fixes a failure in `convert-schema-json-to-human` target. Not found by our standard nightly fuzzing run, but instead by an independent overnight run. Failed after ~7.5 hours.

*Description of changes:*


john-h-kastner-aws:
There are some other issues with this equivalence check @vasumv has turned up with modified generators, and I expect the exact changes necessary here will depend on cedar-policy/cedar#1060, so I'll close this for now so we can sort out exactly what the equivalence check needs to be later.

PR #398, opened 2024-07-22
Pin batteries to 4.8.0

aaronjeline:
*Issue #, if available:*
N/A
*Description of changes:*
Updating batteries to latest


cdisselkoen (approved):
Looks like it's anchoring us to 4.8.0 rather than `main`, which is not an update per se, but I assume it is what we want?

khieta (commented):
Based on https://github.com/leanprover-community/batteries/tags, it looks like the latest version is 4.9.1. Is there a reason we want 4.8.0?

cdisselkoen:
We're currently on Lean 4.8; I assume we want to be on the matching version of Batteries?

Separately, we could update to Lean 4.9

khieta (approved):
Gotcha. I'll approve then.

Issue #142, opened 2023-10-30
Formalize the validation of entity data and requests


cdisselkoen:
### Category

Lean formalization

### Describe the feature you'd like to request

Cedar contains code to (optionally) validate entity data and requests (against a schema) as they are constructed.  This functionality could also be added to the Cedar formal model.

### Describe alternatives you've considered

N/A

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

PR #399, opened 2024-07-29
Entity and Request Validation

bhaktishh:
*Issue #, if available:*
#142
[cedar pr](https://github.com/cedar-policy/cedar/pull/1097)
*Description of changes:*

This PR adds entity and request validation to the lean specification, along with the required infrastructure for two new DRT targets, `entity-validation` and `request-validation`.
CI will fail until the cedar PR is merged. 

@khieta @emina


cedar-drt/fuzz/Cargo.toml
```diff
@@ -179,3 +179,15 @@ name = "convert-policy-json-to-cedar"
 path = "fuzz_targets/convert-policy-json-to-cedar.rs"
 test = false
 doc = false
+
+[[bin]]
+name = "entity-validation"
```
khieta, line 184: Also add a description of the new targets to the table in `cedar-drt/README.md`

cedar-drt/fuzz/fuzz_targets/entity-validation.rs
```diff
@@ -0,0 +1,83 @@
...
+     enable_arbitrary_func_call: true,
+     enable_unknowns: false,
+     enable_action_in_constraints: true,
+     enable_unspecified_apply_spec: true,
+ };
+ 
+ impl<'a> Arbitrary<'a> for FuzzTargetInput {
+     fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
+         let schema: Schema = Schema::arbitrary(SETTINGS.clone(), u)?;
+         let hierarchy = schema.arbitrary_hierarchy(u)?;
```
khieta, line 57: I think this is supposed to generate hierarchies that are consistent with the schema -- do you have stats for how often it generates something that passes entity validation? If it always passes, then you may want to switch to an arbitrary generation scheme (see the "mode" setting [here](https://github.com/cedar-policy/cedar-spec/blob/b83d16100200ca5e2947d298b03ea5d9812e60a4/cedar-policy-generators/src/hierarchy.rs#L312)).

cedar-drt/fuzz/fuzz_targets/entity-validation.rs
```diff
@@ -0,0 +1,83 @@
...
+         Ok(Self { schema, hierarchy })
+     }
+ 
+     fn size_hint(depth: usize) -> (usize, Option<usize>) {
+         arbitrary::size_hint::and_all(&[
+             Schema::arbitrary_size_hint(depth),
```
khieta, line 63: I think you also need `HierarchyGenerator::size_hint(depth)` here? @aaronjeline or @vasumv probably know the answer.

cedar-drt/fuzz/fuzz_targets/entity-validation.rs
```diff
@@ -0,0 +1,83 @@
...
+         ])
+     }
+ }
+ 
+ // Non-type-directed fuzzing of (strict) validation.
```
khieta, line 68: Outdated comment

cedar-drt/fuzz/src/lib.rs
```diff
@@ -294,6 +294,97 @@ pub fn run_val_test(
     }
 }
 
+// API for request validation
```
khieta, line 297: Reminder to delete these unused comments

cedar-drt/fuzz/fuzz_targets/request-validation.rs
```diff
@@ -0,0 +1,112 @@
...
+        let requests = input
+        .requests
+        .into_iter()
+        .map(Into::into)
+        .collect::<Vec<_>>();
+        for request in requests.iter().cloned() {
+            debug!("Request: {request}");
+            let (_, total_dur) =
+                time_function(|| run_req_val_test(&def_impl, schema.clone(), request, Extensions::all_available()));
+            info!("{}{}", TOTAL_MSG, total_dur.as_nanos());
```
khieta, line 108: I'd just delete all the `time_function` and `info!` stuff for the first pass. We can add it back later if we decide we want it.

cedar-drt/fuzz/src/lib.rs
```diff
@@ -294,6 +294,97 @@ pub fn run_val_test(
...
+            Some(&schema),
+            extensions
+        ));
+    info!("{}{}", RUST_AUTH_MSG, rust_auth_dur.as_nanos());
+
+    let definitional_res = custom_impl.validate_request(&schema, &request);
+    match definitional_res {
+        TestResult::Failure(_) => {
+            ;
+        }
```
khieta, line 337: I imagine you also want to check that Rust validation fails here? Same comment in `run_ent_val_test` below.

cedar-drt/fuzz/src/lib.rs
```diff
@@ -294,6 +294,97 @@ pub fn run_val_test(
...
+    custom_impl: &impl CedarTestImplementation,
+    schema: ValidatorSchema,
+    entities: Entities,
+    extensions: Extensions<'_>
+) {
+    let (rust_res, rust_auth_dur) =
+    time_function(|| Entities::from_entities(
+        entities.iter().cloned(),
+        Some(&cedar_policy_validator::CoreSchema::new(&schema)),
+        TCComputation::ComputeNow, // todo
```
khieta, line 366: `AlreadyComputed` (or whatever the name is) should be fine here since the entities are coming from an `Entities` object

cedar-lean/Cedar/Data/Set.lean
```diff
@@ -112,6 +112,7 @@ def singleton {α} (a : α) : Set α :=
 def foldl {α β} (f : α → β → α) (init : α) (s : Set β) : α :=
   s.elts.foldl f init
 
+
```
khieta, line 115: Nit: revert these small spacing changes to keep the diff clean

cedar-lean/Cedar/Validation/RequestEntityValidator.lean
```diff
@@ -0,0 +1,157 @@
+import Cedar.Validation.Validator
+import Cedar.Validation.Typechecker
+namespace Cedar.Validation
+
+open Cedar.Spec
+open Cedar.Data
+
```
khieta, line 7: What's the relation of the definitions in this file to the `RequestAndEntitiesMatchEnvironment` predicate used in proofs?

khieta:
Also, you'll want to run `cargo fmt` (the Rust formatter) to get CI closer to passing (there will still be some failures until the `cedar` PR is merged)

cedar-lean/Cedar/Validation/RequestEntityValidator.lean
```diff
@@ -0,0 +1,157 @@
...
+abbrev RequestValidationResult := Except RequestValidationError Unit
+
+inductive EntityValidationError where
+| typeError (loc : String)
+
+abbrev EntityValidationResult := Except EntityValidationError Unit
+
+def instanceOfBoolType (b : Bool) (bty : BoolType) : Bool :=
+match b, bty with
```
emina, line 19: Nit: we indent the function body (`match`) and leave `match` cases at the same level of indentation as the `match`.

cedar-lean/Cedar/Validation/RequestEntityValidator.lean
```diff
@@ -0,0 +1,157 @@
...
+
+def instanceOfEntityType (e : EntityUID) (ety : EntityType ) : Bool := ety == e.ty
+
+def instanceOfExtType (ext : Ext) (extty: ExtType) : Bool :=
+match ext, extty with
+  | .decimal _, .decimal => true
+  | .ipaddr _, .ipAddr => true
+  | _, _ => false
+
+def requiredAttributesPresent (r : Map Attr Value) (rty : Map Attr (Qualified CedarType)) (k : Attr) :=
```
emina, line 33: This is testing the presence of a single attribute.  Let's rename it to `requiredAttributePresent`.

cedar-lean/Cedar/Validation/RequestEntityValidator.lean
```diff
@@ -0,0 +1,157 @@
...
+match rty.find? k with
+    | .some qty => if qty.isRequired then r.contains k else true
+    | _ => true
+
+def instanceOfType (v : Value) (ty : CedarType) : Bool := match v, ty with
```
emina, line 38: Formatting:  `match v, ty with` on a separate line, indented with respect to `def` (and `match` cases at the same level of indentation as `match`).

Alternatively, change this line to `def instanceOfType : Value → CedarType → Bool` and then indent the cases one level with respect to `def`.

cedar-lean/Cedar/Validation/RequestEntityValidator.lean
```diff
@@ -0,0 +1,157 @@
...
+| .prim (.bool b), .bool bty => instanceOfBoolType b bty
+| .prim (.int _), .int => true
+| .prim (.string _), .string => true
+| .prim (.entityUID e), .entity ety => instanceOfEntityType e ety
+| .set s, .set ty => (s.elts.attach.map (λ ⟨v, _⟩ => instanceOfType v ty)).all id
```
emina, line 43: Why do a `map` and then `all id`?  I think the following simpler expressions should work:

```
s.elts.attach.all (λ ⟨v, _⟩ => instanceOfType v ty)
```

cedar-lean/Cedar/Validation/RequestEntityValidator.lean
```diff
@@ -0,0 +1,157 @@
...
+    case _ h₁ =>
+      have := Set.sizeOf_lt_of_mem h₁
+      omega
+    case _ h₁ =>
+      cases r
+      simp only [Map.kvs] at h₁
+      simp only [Map.mk.sizeOf_spec]
+      omega
+
+def instanceOfRequestType (request : Request) (reqty : RequestType) : Bool :=
```
emina, line 64: Why use nested if-then-else here?  The following should be equivalent:

```
instanceOfEntityType request.principal reqty.principal &&
request.action == reqty.action && 
instanceOfEntityType request.resource reqty.resource &&
instanceOfType request.context (.record reqty.context)

```

cedar-lean/Cedar/Validation/RequestEntityValidator.lean
```diff
@@ -0,0 +1,157 @@
...
+/--
+For every entity in the store,
+1. The entity's type is defined in the type store.
+2. The entity's attributes match the attribute types indicated in the type store.
+3. The entity's ancestors' types are consistent with the ancestor information
+   in the type store.
+-/
+
+def instanceOfEntitySchema (entities : Entities) (ets : EntitySchema) : EntityValidationResult :=
+entities.toList.forM (λ (uid, data) => match ets.find? uid.ty with
```
emina, line 82: Indent the function body with respect to `def`.  Because this lambda is quite complicated, I suggest defining it separately using a `where` clause, for readability and proofs.  When proving the equivalence of this definition to the predicate version, it will be convenient to have a named function for the lambda's body and prove its properties separately:

```
def instanceOfEntitySchema (entities : Entities) (ets : EntitySchema) : EntityValidationResult :=
  entities.toList.forM λ (uid, data) => instanceOfEntityData uid data
where 
  instanceOfEntityData uid data => 
    match ets.find? uid.ty with 
    | ...
```

cedar-lean/Cedar/Validation/RequestEntityValidator.lean
```diff
@@ -0,0 +1,157 @@
...
+                   else .error (.typeError s!"entity ancestors inconsistent with type store information")
+                  else .error (.typeError "entity attributes do not match type store")
+  | _ => .error (.typeError "entity type not defined in type store"))
+
+/--
+For every action in the entity store, the action's ancestors are consistent
+with the ancestor information in the action store.
+-/
+def instanceOfActionSchema (entities : Entities) (as : ActionSchema) : EntityValidationResult :=
+as.toList.forM (λ (uid, data) => match entities.find? uid with
```
emina, line 94: Same comment as for `instanceOfEntitySchema`.

cedar-lean/Cedar/Validation/RequestEntityValidator.lean
```diff
@@ -0,0 +1,157 @@
...
+}
+
+/--
+Update the entity schema with the entities created for action schema entries.
+This involves the construction of the ancestor information for the associated types
+by inspecting the concrete hierarchy.
+-/
+def updateSchema (schema : Schema) (actionSchemaEntities : Entities) : Schema :=
+  let uniqueTys := Set.make (actionSchemaEntities.keys.map (·.ty)).elts
+  let newEntitySchemaEntries := uniqueTys.elts.map (λ ty =>
```
emina, line 122: The inlined lambda here is complex.  Consider moving it to a separate `where` clause.

emina (changes requested):
Looks good!

A few formatting nits throughout: indent the function's body with respect to `def`; keep `match` expressions on a separate line; and keep `match` cases at the same level of indentation as the `match`.

If the body of an anonymous lambda is complicated, it's a good idea to move it to a separate named function using a `where` clause.  This helps with readability, and also with proofs (because the named functions are opaque and can be unfolded on demand).

cedar-lean/Cedar/Validation/RequestEntityValidator.lean
bhaktishh: ah it was nested because i was returning a custom error for each one before; i changed it in the last pass

cedar-drt/fuzz/fuzz_targets/entity-validation.rs
bhaktishh: I think it is arbitrary? I don't have stats on how often it passes but i thought [SchemaBased](https://github.com/cedar-policy/cedar-spec/blob/b83d16100200ca5e2947d298b03ea5d9812e60a4/cedar-policy-generators/src/hierarchy.rs#L362) was the one that produced conforming hierarchies

cedar-drt/fuzz/fuzz_targets/entity-validation.rs
bhaktishh: The code for `validation-drt` doesn't use this, so I assumed it wouldn't be needed here

cedar-drt/fuzz/src/lib.rs
bhaktishh: I think this is not necessarily when validation fails, but more like when parsing fails or something along those lines. I wasn't sure what the desired behavior here would be and forgot to ask

cedar-drt/fuzz/fuzz_targets/request-validation.rs
bhaktishh: Are you sure? It's not that big of a deal and it just matches up with the rest of the targets better.

cedar-lean/Cedar/Validation/RequestEntityValidator.lean
bhaktishh: it's almost a reflection from prop to bool, except that the function for entity validation has a pre-processing step that adds entries in the action schema to the list of entities and the entity schema, to match both the rust code and the preconditions for proofs.

cedar-drt/fuzz/fuzz_targets/entity-validation.rs
khieta: Yup, you're right. Since `match_types` is false above you should be good.

cedar-drt/fuzz/src/lib.rs
khieta: My bad, you're right. Let's panic here (and in `run_ent_val_test`) so we can quickly identify if there's an issue with the parsing.

cedar-drt/fuzz/fuzz_targets/validation-drt.rs
```diff
@@ -59,6 +59,7 @@ impl<'a> Arbitrary<'a> for FuzzTargetInput {
     fn size_hint(depth: usize) -> (usize, Option<usize>) {
         arbitrary::size_hint::and_all(&[
             Schema::arbitrary_size_hint(depth),
+            HierarchyGenerator::size_hint(depth),
```
khieta, line 62: Add this to `entity-validation.rs` and `request-validation.rs` too

cedar-lean/Cedar/Validation/RequestEntityValidator.lean
khieta: As discussed in person: it would be awesome if we could reuse the `Prop` definitions here instead of copying them with the `Bool` type

cedar-drt/fuzz/src/lib.rs
```diff
@@ -294,6 +294,71 @@ pub fn run_val_test(
...
+    let (rust_res, rust_auth_dur) =
+        time_function(|| ast::Request::new_with_unknowns(
+            request.principal().clone(),
+            request.action().clone(),
+            request.resource().clone(),
+            request.context().cloned(),
+            Some(&schema),
+            extensions
+        ));
+    info!("{}{}", RUST_AUTH_MSG, rust_auth_dur.as_nanos());
```
khieta, line 312: Update `RUST_AUTH_MSG` to something specific to request validation

cedar-lean/Cedar/Validation/RequestEntityValidator.lean
bhaktishh: will make a new PR for proofs of this

cedar-lean/Cedar/Validation/RequestEntityValidator.lean
```diff
@@ -0,0 +1,162 @@
...
+
+def requestMatchesEnvironment (env : Environment) (request : Request) : Bool := instanceOfRequestType request env.reqty
+
+def validateRequest (schema : Schema) (request : Request) : RequestValidationResult :=
+  if ((schema.toEnvironments.any (requestMatchesEnvironment · request)))
+  then .ok ()
+  else .error (.typeError "request could not be validated in any environment")
+
+def entitiesMatchEnvironment (env : Environment) (entities : Entities) : EntityValidationResult :=
+instanceOfEntitySchema entities env.ets >>= λ _ => instanceOfActionSchema entities env.acts
```
khieta, line 111:
```suggestion
  instanceOfEntitySchema entities env.ets >>= λ _ => instanceOfActionSchema entities env.acts
```

khieta (approved):

cedar-lean/Cedar/Spec/Value.lean
```diff
@@ -42,6 +42,7 @@ structure Name where
 
 abbrev EntityType := Name
 
+
```
emina, line 45: Can we remove this whitespace to get rid of the spurious diff?

emina (approved):

khieta (approved):

emina (approved):

PR #405, opened 2024-07-31
Fix Lean CLI JSON samples and add test to CI

john-h-kastner-aws:
*Issue #, if available:*

JSON samples were broken when removing unspecified and we didn't notice because they aren't tested.

*Description of changes:*


khieta (approved):
Nice

PR #414, opened 2024-08-08
proofs of equivalence for propositional and boolean formulations for entity and request validation

bhaktishh:
*Description of changes:*
This PR adds proofs of equivalence for the propositional and boolean formulations of entity and request validation. It also updates the toplevel validation soundness theorem to use the computational validation functions as prerequisites rather than the predicates. 

@khieta @emina


cedar-lean/DiffTest/Main.lean
```diff
@@ -122,6 +122,9 @@ def runAndTime (f : Unit -> α) : BaseIO (Timed α) := do
     | .ok json => do
         let schema ← getJsonField json "schema" >>= jsonToSchema
         let entities ← getJsonField json "entities" >>= jsonToEntities
+        let actionEntities := (schema.acts.mapOnValues actionSchemaEntryToEntityData)
+        let entities := Cedar.Data.Map.make (entities.kvs ++ actionEntities.kvs)
+        let schema := updateSchema schema actionEntities
```
khieta, line 127: I'm still not entirely convinced about the `updateSchema` step. I feel like it shouldn't be necessary, since we don't do the equivalent in the Rust. I also worry that it might lead to different results -- IIUC, the Lean code will end up saying that an `Action` type can have an ancestor of type `Action`, while the schema actually says that `Action::"foo"` can only have ancestor `Action::"bar"`. Did you explore the option to validate action entities separately from non-action entities?

cedar-lean/Cedar/Thm/Validation/Validator.lean
```diff
@@ -36,7 +36,7 @@ def AllEvaluateToBool (policies : Policies) (request : Request) (entities : Enti
   ∀ policy ∈ policies, EvaluatesToBool policy.toExpr request entities
 
 def RequestAndEntitiesMatchSchema (schema : Schema) (request : Request) (entities : Entities) :Prop :=
-  ∀ env ∈ schema.toEnvironments,
+  ∃ env ∈ schema.toEnvironments,
```
khieta, line 39: Maybe add a doc comment on this predicate explaining what it does and why you want "exists" instead of "forall". (I imagine "forall" is more intuitive, which is why we used it originally.)

khieta (approved):
Looks great! Just left a couple small comments.

cedar-lean/Cedar/Thm/Validation/Validator.lean
bhaktishh: It was actually me that wrote that originally and I just understood it very wrong, and no one caught it in the last PR. I can add a doc comment about it though.

cedar-lean/DiffTest/Main.lean
bhaktishh: Let me @emina here because she was much more confident than me about this approach. I did explore that option but this was the easier change

emina: Updating the schema is necessary because the proofs assume that entities are consistent with the schema:

https://github.com/cedar-policy/cedar-spec/blob/8bfa7dc9c5652f79539d3a158e6fec479e3ba79d/cedar-lean/Cedar/Thm/Validation/Typechecker/Types.lean#L82

For every entity, there exists an entry in the schema, and the type of that entity is consistent with the schema.  Actions are entities so this constraint applies to them too, and Action types must be defined in the schema to satisfy the constraint.

Updating the schema to include the action types (and their ancestors) won't affect typechecking, because the relevant typechecking code first tests if something is an action, before checking conditions that apply to other entities, e.g.:

https://github.com/cedar-policy/cedar-spec/blob/8bfa7dc9c5652f79539d3a158e6fec479e3ba79d/cedar-lean/Cedar/Validation/Typechecker.lean#L137

It is possible to change the assumptions (to distinguish between actions and other entities in the InstanceOfEntitySchema predicate) so that updating the schema isn't necessary.  That involves changing the proofs to work with the new assumptions.  IIRC, Bhakti tried it at some point (?), and it seemed like it would be pretty involved.  We can revisit now that we have a working version to see how much work is involved.

emina: Whichever approach we adopt, it's worth considering how it will be extended (eventually) to support other enumerated entities.  Actions are a special case of an enumerated type, and the approach we take here will be relevant to other enums too.

emina: Actually, this does raise an interesting question.  What happens when we try to typecheck an action that's not concretely specified, e.g., `principal.action in resource.action`, where both `principal.action` and `resource.action` have some Action type.

The Lean validator will typecheck this case in the same way that it typechecks `in` for other entity types---namely, by falling back to the `descendentOf` check.  

What does Rust do?

If Rust does something different, then we might have a miss in differential testing.

khieta: Here's the relevant code in the Rust: https://github.com/cedar-policy/cedar/blob/2bd45edd0fdb746e25355f78195bd8b87c70d70b/cedar-policy-core/src/entities/conformance.rs#L52

Basically, it has totally separate handling for actions and non-actions. For action, it just checks that the provided entity exactly matches the action declaration in the schema.

emina: Right, we can't do that in Lean because of the assumption that actions are entities (see above).  I don't object to revisiting the assumptions and proofs.

How does the Rust typechecker handle cases like `principal.action in resource.action`?

khieta: In Rust, I'm not sure if you can have a general `Action` type in schemas, which would be required to have a field of type `Action`. So I'm not sure about the behavior of `principal.action in resource.action`. @john-h-kastner-aws do you know?

emina: Interesting, so actions in Rust aren't really entities.  They are sort of second-class values: only the `action` variable or a literal can have type action. 

It seems like it could be useful to allow actions in attributes, so I'm curious to know why we decided to disallow it.

john-h-kastner-aws: Relevant issue: https://github.com/cedar-policy/cedar/issues/304

Trying to declare the `Action` entity type gives an error like ```entity type `Action` declared```. Trying to use it without declaring gives something like `failed to resolve type: Action`.

emina: Ah very interesting!  

So it sounds like we should keep first-class action support in Lean.

The main question is whether to keep in the current form, where action types are in the entity schema and action data (hierarchy etc) is in the action schema.

Also let’s keep in mind enums and which solution would work better for them as well.

khieta: Ok, I think I understand things better now: the `updateSchema` function adds (imprecise) action information to the entity schema, while the full action information remains in the action schema. The full information is used when possible, and the less precise information is used as a fallback. I think this design makes sense & will likely work for enums as well.

I'm still concerned though that we've added code in the Lean that doesn't exist in the Rust, so I'd like to see some assurances about what the new code does. Is there a nice way to state and prove that the `updateSchema` function adds information to the entity part of the schema that is consistent with the action part?

emina: I think so.

For example, we could prove that `actionSchema` data is consistent with the result of `updateSchema` according to the constraints given in `InstanceOfEntitySchema`. 

We can also prove that the contents of a schema are always included in `updateSchema`, and finally, that nothing other than action type data is added by `updateSchema`.

From these, some auxiliary lemmas should follow easily.  For example, if an entity is  `InstanceOfEntitySchema` for the original schema, then it's also `InstanceOfEntitySchema` for the updated schema.  And same for `InstanceOfActionSchema`.

bhaktishh:
@khieta As we discussed, I think this one is ready to merge, as I'll separate out the proofs about `updateSchema` into another PR. I've refactored this PR to remove progress on `updateSchema`.

khieta (approved):
Splitting out into a separate PR sounds good. @emina can we get your sign off on this PR too?

emina (approved):

PR #419, opened 2024-08-19
Fix slow units in `convert-policy-json-to-cedar`

shaobo-he-aws:
*Issue #, if available:*

*Description of changes:*
Slow units occur when we deserialize ESTs directly from strings. This PR avoids them by deserializing the strings to JSON values and then to ESTs.

### DIFF
--- a/cedar-drt/fuzz/fuzz_targets/convert-policy-json-to-cedar.rs
+++ b/cedar-drt/fuzz/fuzz_targets/convert-policy-json-to-cedar.rs
@@ -30,7 +30,8 @@ enum ESTParseError {
 }
 
 fuzz_target!(|est_json_str: String| {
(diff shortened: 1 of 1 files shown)


khieta (approved):
It'd be great to have some data in the description of the PR for how much this matters. It seems weird that this makes a difference!

Also: let's make sure we use the faster code path in our public APIs too (I assume we provide something like `Policy::from_json_string()`?)

cdisselkoen (approved):

cdisselkoen:
+1 to that, if anything I would have expected that this would actually make things _slower_.  (Otherwise why doesn't serde_json implement `from_str` internally as string -> serde_json::Value -> T?)

john-h-kastner-aws (approved):
Any idea if this is a general change we should make elsewhere, or is it somehow specific to our EST structs?

PR #420, opened 2024-08-20
update for cedar#1136

cdisselkoen:
*Issue #, if available:*

cedar-policy/cedar#1136

### DIFF
--- a/cedar-drt/fuzz/src/schemas.rs
+++ b/cedar-drt/fuzz/src/schemas.rs
@@ -14,16 +14,18 @@
  * limitations under the License.
  */
 
-use cedar_policy_core::ast::{Id, InternalName, UnreservedId};
+use cedar_policy_core::ast::{Id, InternalName};
 use cedar_policy_validator::json_schema::{
-    ApplySpec, EntityType, RecordType, Type, TypeOfAttribute, TypeVariant,
+    ApplySpec, EntityAttributeType, EntityAttributeTypeInternal, EntityAttributes,
+    EntityAttributesInternal, EntityType, RecordAttributeType, RecordType, Type, TypeVariant,
 };
 use cedar_policy_validator::RawName;
 use itertools::Itertools;
-use std::collections::{HashMap, HashSet};
+use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
 
 use cedar_policy_validator::json_schema;
 use std::fmt::{Debug, Display};
+use std::hash::Hash;
 
 /// Check if two schema fragments are equivalent, modulo empty apply specs.
 /// We do this because there are schemas that are representable in the JSON that are not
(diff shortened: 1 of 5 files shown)


shaobo-he-aws (approved):

cedar-drt/fuzz/src/schemas.rs
```diff
@@ -83,171 +85,301 @@ fn remove_trivial_empty_namespace<N>(schema: &mut json_schema::Fragment<N>) {
...
+            let rhs_keys: BTreeSet<_> = rhs.keys().collect();
+            let missing_keys = lhs_keys.symmetric_difference(&rhs_keys).join(", ");
+            Err(format!("Missing keys: {missing_keys}"))
+        }
+    }
+}
+
+impl<N: Clone + PartialEq + Debug + Display + TypeName + Ord> Equiv for EntityType<N> {
+    fn equiv(lhs: &Self, rhs: &Self) -> Result<(), String> {
+        if !vector_equiv(&lhs.member_of_types, &rhs.member_of_types) {
```
khieta, line 175: Any reason `vector_equiv` isn't also an implementation of `Equiv`?

cedar-drt/fuzz/src/schemas.rs
```diff
@@ -331,53 +463,58 @@ fn action_type_equivalence<N: PartialEq + Debug + Display + Clone + TypeName + O
...
-    let context_ty2: HashSet<&cedar_policy_validator::types::Type> = HashSet::from_iter(
-        schema2
-            .action_entities()
-            .unwrap()
-            .iter()
-            .map(|e| schema1.get_action_id(e.uid()).unwrap().context_type()),
-    );
-    entity_attr_tys1 == entity_attr_tys2 && context_ty1 == context_ty2
+impl Equiv for cedar_policy_validator::ValidatorSchema {
+    fn equiv(lhs: &Self, rhs: &Self) -> Result<(), String> {
```
khieta, line 487: Seems a little weird that this function isn't using any other `equiv` functions (e.g., for `HashMap`s)

khieta (approved):
Seems fine from my cursory inspection (and the fact that CI passes).

cedar-drt/fuzz/src/schemas.rs
cdisselkoen: It actually does set equivalence, and I suppose it's unclear whether we _always_ want set equivalence whenever we're applying `Equiv` to vectors?

It would probably be reasonable to make it the canonical `Equiv` implementation though, and assume that we never have ordered lists in the schema format (or anything else we're `Equiv`ing)

cdisselkoen: Removed `vector_equiv` entirely by collecting into hashset/btreeset first. Makes more sense to do set equivalence that way instead of sorting anyway.

khieta (approved):

PR #426, opened 2024-08-27
Changes needed by #1150

shaobo-he-aws:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cedar-policy-generators/src/schema.rs
+++ b/cedar-policy-generators/src/schema.rs
@@ -31,6 +31,7 @@ use crate::{accum, gen, gen_inner, uniform};
 use arbitrary::{self, Arbitrary, Unstructured};
 use cedar_policy_core::ast::{self, Effect, PolicyID, UnreservedId};
 use cedar_policy_core::extensions::Extensions;
+use cedar_policy_validator::json_schema::CommonTypeId;
 use cedar_policy_validator::{
     json_schema, ActionBehavior, AllDefs, RawName, SchemaError, ValidatorNamespaceDef,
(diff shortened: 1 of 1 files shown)


cdisselkoen (approved):

john-h-kastner-aws (approved):

cedar-policy-generators/src/schema.rs
```diff
@@ -641,19 +643,24 @@ impl Bindings {
...
+    fn add_binding(&mut self, binding: (json_schema::Type<ast::InternalName>, CommonTypeId)) {
         let (ty, id) = binding;
+        let id: UnreservedId = id.into();
         // create a new id when the provided id has been used
         let id_smolstr: &str = id.as_ref();
         let new_id = if self.ids.contains(id_smolstr) {
             let mut new_id = id.to_string();
             while self.ids.contains(new_id.as_str()) {
                 new_id.push('_');
+                new_id.push('_');
```
john-h-kastner-aws, line 655: I'm looking at this again because of a DRT failure on this target in 3.4 and don't understand why you need to push `_` three times here.

shaobo-he-aws: IIRC, adding three `_`s makes sure that an id is valid. Counterexamples for adding one time and two times are `_cedar` and `cedar`.

john-h-kastner-aws: so probably not relevant to 3.4 because we reserved names in 4?

shaobo-he-aws: Correct.

PR #429, opened 2024-08-29
DRT for Entity Slicing

oflatt:
*Issue #, if available:*

*Description of changes:*

This PR adds a new fuzzing target to test the results of entity slicing. It checks that the authorization result is still correct after performing slicing. It was already able to catch a bug.

This should be merged after: https://github.com/cedar-policy/cedar/pull/1154


shaobo-he-aws:
Closed in favor of #475

PR #430, opened 2024-09-06
update doc-gen4 version

khieta:
*Issue #, if available:*

*Description of changes:*

I believe the reason for the recent deployment failures (e.g., https://github.com/cedar-policy/cedar-spec/actions/runs/10726230216/job/29745800347) is that `doc-gen4` was pinned to the wrong version. I've updated `lakefile.lean` to point to the commit corresponding to the v4.10.0 release. I also added a PR CI check so we can find out earlier if the doc build is failing.

~~Also, while playing with some proofs, I found that we could remove some `termination_by` clauses with the latest version of Lean. Not sure if it's best practice to leave them in or take them out if not needed.~~

### DIFF
--- a/cedar-lean/Cedar/Thm/Partial/Evaluation/Evaluate.lean
+++ b/cedar-lean/Cedar/Thm/Partial/Evaluation/Evaluate.lean
@@ -418,6 +418,7 @@ theorem subst_preserves_evaluation_to_value {expr : Expr} {req req' : Partial.Re
       have := List.sizeOf_lt_of_mem h₂
       exact subst_preserves_evaluation_to_value wf_r wf_e wf_s h_req hx
     exact Call.subst_preserves_evaluation_to_value ih h_req v h₁
+termination_by expr
 
 /--
   If partial evaluation returns an error, then it also returns an error (not
@@ -509,6 +510,7 @@ theorem subst_preserves_errors {expr : Expr} {req req' : Partial.Request} {entit
       have := List.sizeOf_lt_of_mem hx
       exact subst_preserves_errors wf_r wf_e wf_s h_req
     }
+termination_by expr
 
 /--
   Corollary (contrapositive) of the above:
--- a/cedar-lean/Cedar/Spec/Wildcard.lean
+++ b/cedar-lean/Cedar/Spec/Wildcard.lean
@@ -69,6 +69,10 @@ def wildcardMatchIdx (text : List Char) (pattern : Pattern) (i j : Nat)
     r := charMatch (text.get ⟨i, (by omega)⟩) (pattern.get ⟨j, (by omega)⟩) &&
          (← wildcardMatchIdx text pattern (i + 1) (j + 1) (by omega) (by omega))
   modifyGet λ cache => (r, cache.insert (i, j) r)
+termination_by
+  (text.length - i) + (pattern.length - j)
+decreasing_by
+  all_goals { simp_wf ; omega }
 
 def wildcardMatch (text : String) (pattern : Pattern) : Bool :=
   wildcardMatchIdx text.toList pattern 0 0 (by simp) (by simp) |>.run' HashMap.empty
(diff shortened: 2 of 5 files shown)


cedar-lean/Cedar/Thm/Partial/Evaluation/Evaluate.lean
```diff
@@ -418,7 +418,6 @@ theorem subst_preserves_evaluation_to_value {expr : Expr} {req req' : Partial.Re
       have := List.sizeOf_lt_of_mem h₂
       exact subst_preserves_evaluation_to_value wf_r wf_e wf_s h_req hx
     exact Call.subst_preserves_evaluation_to_value ih h_req v h₁
-termination_by expr
```
cdisselkoen, line 421: for this one, I think I remember leaving this one because if the proof does break, having the `termination_by` explicit led to a better error message than if you remove this line.  (Without this line, something like "failed to find a termination measure"; with this line, some actual failed proof obligation where the `expr` can't be proved to be smaller at a particular call site.)  That said, I'm fine if we want to adopt the convention to omit this line whenever Lean allows it.  It's easy enough to add it back (even locally) while troubleshooting termination proofs.

cdisselkoen (approved):

cedar-lean/Cedar/Thm/Partial/Evaluation/Evaluate.lean
khieta: Yeah, I'd buy that leaving them makes it easier to debug in the future if things break when heuristics change. Super satisfying to delete them though 😅 @emina any opinions?

khieta:
I have emerged victorious in my battle with the CI ⚔️ 

I made some updates to the build process to get things working. Most notably I deleted the autogenerated `lake-manifest.json` file. As a result, you will need to run `lake update` before you build the project. I've updated our READMEs accordingly.

.github/workflows/deploy_docs.yml
```diff
@@ -34,14 +34,14 @@ jobs:
...
+      - name: Download dependencies
         working-directory: ./cedar-lean
         shell: bash
-        run: source ~/.profile && sed -i '/^meta if get_config? env = some "dev" then/d' lakefile.lean && lake update doc-gen4
+        run: source ~/.profile && lake -R -Kenv=dev update
       - name: Build docs
         working-directory: ./cedar-lean
         shell: bash
-        run: source ~/.profile && lake -Kenv=dev build Cedar:docs
+        run: source ~/.profile && lake -R -Kenv=dev build Cedar:docs
```
cdisselkoen, line 44: for my curiosity, why is `-R` required?

cedar-lean/lakefile.lean
```diff
@@ -18,9 +18,9 @@ import Lake
 open Lake DSL
 
 meta if get_config? env = some "dev" then -- dev is so not everyone has to build it
-require «doc-gen4» from git "https://github.com/leanprover/doc-gen4" @ "c7f4ac84b973b6efd8f24ba2b006cad1b32c9c53"
+require «doc-gen4» from git "https://github.com/leanprover/doc-gen4" @ "v4.10.0"
 
-require "leanprover-community" / "batteries"
+require «batteries» from git "https://github.com/leanprover-community/batteries" @ "v4.10.0"
```
cdisselkoen, line 23: any way we can avoid this reversion to the old syntax? see https://lean-lang.org/blog/2024-8-1-lean-4100/ for more on the new syntax and why it's better

cdisselkoen (approved):

cedar-lean/lakefile.lean
khieta: Updated in the latest commit. Note that we still need to pin to a specific git version, but I _think_ the new syntax is still going through Reservoir.

.github/workflows/deploy_docs.yml
khieta: Tbh I'm not sure. It wasn't working without the `-R` in `ci.yml` so I figured I might as well make the change in `deploy_docs.yml` too. All the instructions on https://github.com/leanprover/doc-gen4 use that flag, which is why I tried it.

Here's the command description: `--reconfigure, -R     elaborate configuration files instead of using OLeans`

cdisselkoen (approved):

emina (approved):

emina (approved):

PR #447, opened 2024-10-01
generate hasTag expressions

cdisselkoen:
Updates the generators for expressions to generate `.hasTag()` expressions.

### DIFF
--- a/cedar-policy-generators/src/expr.rs
+++ b/cedar-policy-generators/src/expr.rs
@@ -252,14 +252,6 @@ impl<'a> ExprGenerator<'a> {
                             attr_name,
                         ))
                     },
-                    7 => {
-                        let tag_name = uniform!(u,
-                            self.generate_expr(max_depth - 1, u)?,
-                            ast::Expr::val(self.schema.arbitrary_attr(u)?.0.clone())
-                        );
-                        let e = self.generate_expr(max_depth - 1, u)?;
-                        Ok(ast::Expr::get_tag(e, tag_name))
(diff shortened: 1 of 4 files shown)


cedar-policy-generators/src/expr.rs
```diff
@@ -251,7 +251,24 @@ impl<'a> ExprGenerator<'a> {
...
                         ))
-                    })
+                    },
+                    7 => {
+                        let tag_name = uniform!(u,
+                            self.generate_expr(max_depth - 1, u)?,
+                            ast::Expr::val(self.schema.arbitrary_attr(u)?.0.clone())
+                        );
+                        let e = self.generate_expr(max_depth - 1, u)?;
+                        Ok(ast::Expr::get_tag(e, tag_name))
```
john-h-kastner-aws, line 261: Noting that you do generate `get_tag` also

john-h-kastner-aws (approved):
assuming you'll generate entity data with tags in a future PR

cedar-policy-generators/src/expr.rs
cdisselkoen: Good catch.  We do not generate `get_tag` in all the places we would need to -- in particular we do not generate `get_tag` in the type-directed generators in this PR.  I will just remove this from this PR for clarity.

aaronjeline (approved):

cdisselkoen:
CI failures are due to #448 not merged yet.  Merging this.

PR #457, opened 2024-10-10
update to Lean 4.12

cdisselkoen:


emina (approved):

cedar-lean/Cedar/Data/Set.lean
```diff
@@ -126,7 +126,7 @@ instance : EmptyCollection (Set α) where
 
 -- enables ∈
 instance : Membership α (Set α) where
-  mem v s := v ∈ s.elts
```
aaronjeline, line 129: What's up with this (and the other `mem`) change?

cdisselkoen: One of the changes in Lean 4.11 or 4.12 (forget which) was swapping the argument order for Membership.  They said this was to support currying in some important usecase

aaronjeline: Cool, just was double checking the argument swap was intentional

aaronjeline (approved):

PR #464, opened 2024-11-05
Fix arbitrary to 1.3.2

shaobo-he-aws:
*Issue #, if available:*
Apparently 1.4 is a breaking change.

*Description of changes:*

### DIFF
--- a/cedar-policy-generators/Cargo.toml
+++ b/cedar-policy-generators/Cargo.toml
@@ -5,7 +5,8 @@ version = "4.0.0"
 publish = false
 
 [dependencies]
-arbitrary = "1.3"
+arbitrary = "=1.3.2"
+derive_arbitrary = "=1.3.2"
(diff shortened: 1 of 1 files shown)


cdisselkoen (approved):

adpaco-aws:
Would it make sense to commit the `Cargo.lock` file instead?

If we decided to go with this solution, could you attach an issue so we know the reason and keep track of the work to upgrade?

cdisselkoen:
Since this is a library, its Cargo.lock will not be respected by packages depending on this library.  If 1.4 truly doesn't work with this code, I prefer the solution here, not (just) committing the Cargo.lock.  I agree a comment would be great though.

adpaco-aws (approved):

john-h-kastner-aws:
> Would it make sense to commit the Cargo.lock file instead?

> I prefer the solution here, not (just) committing the Cargo.lock.

I think we agree that it makes sense to start committing `Cargo.lock` in this repository for the same reasons we commit it in the main `cedar` repo.

PR #470, opened 2024-11-08
Changes needed to test `datetime` extension

shaobo-he-aws:
*Issue #, if available:*
https://github.com/cedar-policy/cedar/pull/1276

*Description of changes:*

### DIFF
--- a/cedar-policy-generators/src/abac.rs
+++ b/cedar-policy-generators/src/abac.rs
@@ -296,6 +296,84 @@ impl ConstantPool {
         .map(SmolStr::new)
     }
 
+    // Generate a roughly valid datetime string
+    fn arbitrary_datetime_str_inner(&self, u: &mut Unstructured<'_>) -> Result<String> {
+        let mut result = String::new();
+        // Generate YYYY-MM-DD
+        let y = u.int_in_range(0..=999)?;
+        let m = u.int_in_range(1..=12)?;
+        let d = u.int_in_range(1..=31)?;
+        result.push_str(&format!("{:04}-{:02}-{:02}", y, m, d));
+        // There's a 25% chance where just a date is generated
+        if u.ratio(1, 4)? {
+            return Ok(result);
+        }
+        // Generate hh:mm:ss
+        result.push('T');
+        let h = u.int_in_range(0..=23)?;
+        let m = u.int_in_range(0..=59)?;
+        let s = u.int_in_range(0..=59)?;
+        result.push_str(&format!("{:02}:{:02}:{:02}", h, m, s));
+        match u.int_in_range(0..=3)? {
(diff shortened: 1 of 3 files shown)


john-h-kastner-aws (commented):
I'm concerned we might not generate schema with the new types

cedar-policy-generators/src/schema.rs
```diff
@@ -1179,6 +1179,12 @@ impl Schema {
             Type::Decimal => Some(json_schema::TypeVariant::Extension {
                 name: "decimal".parse().unwrap(),
             }),
+            Type::DateTime => Some(json_schema::TypeVariant::Extension {
```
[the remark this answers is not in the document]
shaobo-he-aws: Good catch. There should.

john-h-kastner-aws (approved):

cedar-policy-generators/src/abac.rs
```diff
@@ -296,6 +296,84 @@ impl ConstantPool {
         .map(SmolStr::new)
     }
 
+    // Generate a roughly valid datetime string
+    fn arbitrary_datetime_str_inner(&self, u: &mut Unstructured<'_>) -> Result<String> {
+        let mut result = String::new();
+        // Generate YYYY-MM-DD
+        let y = u.int_in_range(0..=999)?;
```
cdisselkoen, line 303: this doesn't include the current year, should this be `0..=3000` or something?

cedar-policy-generators/src/abac.rs
```diff
@@ -767,3 +940,86 @@ impl From<ABACRequest> for ast::Request {
...
+            .collect();
+        valid_strs.len()
+    }
+
+    #[test]
+    fn test_valid_extension_value_ratio() {
+        let mut rng = StdRng::seed_from_u64(666);
+        let mut bytes = [0; 4096];
+        rng.fill_bytes(&mut bytes);
+        let mut u = Unstructured::new(&bytes);
```
cdisselkoen, line 1004: Using an RNG here means the test is nondeterministic.  Would it be better to have a deterministic test?  For instance, some fixed-but-arbitrary 4096-byte array instead of filling it randomly each time the test is executed?

cdisselkoen (approved):

cedar-policy-generators/src/abac.rs
shaobo-he-aws: Good catch. Fixed.

cedar-policy-generators/src/abac.rs
shaobo-he-aws: It should be deterministic because the RNG is seeded.

cdisselkoen (approved):

Issue #461, opened 2024-10-25
Avoid needing to build Lean lib for fuzz targets that don't use it


john-h-kastner-aws:
### Category

DRT infrastructure

### Describe the feature you'd like to request

To compile any fuzz target we currently  link in the lean library even if the fuzz target doesn't do any differential testing (e.g., parser targets and validator prop-test target). We could probably do something with feature flags so that these fuzz targets could be built and run without first building the lean lib.

### Describe alternatives you've considered

.

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

john-h-kastner-aws:
wip: https://github.com/cedar-policy/cedar-spec/tree/lean-feature-flag

john-h-kastner-aws:
Noting that this could have helped localize issues with recent protobuf changes if that was also behind this flag.

PR #477, opened 2024-11-18
Fail DRT build quickly if lean build dir isn't present

john-h-kastner-aws:
Trying to improve on another papercut in the DRT build process. Add this to remind myself that I need to build the lean library.

### DIFF
--- a/cedar-drt/build.rs
+++ b/cedar-drt/build.rs
@@ -15,10 +15,16 @@
  */
 
 use std::env;
+use std::path::Path;
 fn main() {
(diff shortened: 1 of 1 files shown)


cdisselkoen (commented):
Is this in opposition to #461?

john-h-kastner-aws:
> Is this in opposition to https://github.com/cedar-policy/cedar-spec/issues/461?

Not really That will already require adding a condition build file to not error when `env::var("LEAN_LIB_DIR")` is missing, so this extra check will just go under that

cdisselkoen (approved):

PR #479, opened 2024-11-19
Document vscode settings

john-h-kastner-aws:
I was editing in a new workspace and had to remember that wanted these settings.


shaobo-he-aws (approved):

cdisselkoen (approved):
Should we just check-in a `.vscode/settings.json` with these settings?

PR #482, opened 2024-11-20
various cleanups in Protobuf

cdisselkoen:
- refactors `BParsec.ParseResult` from an `inductive` where the `pos` field was duplicated across variants, to a `structure` that also gets to make use of the built-in `Except` type
- redefines `BParsec.run!` in terms of `BParsec.run`, and improves the panic message
- defines a new helper `BParsec.inspect` and redefines a lot of existing helpers using it.  (also uses it elsewhere in other files)
- redefines `BParsec.eof` and `utf8DecodeChar` using `do` notation for clarity
- adjusts proofs for all of the above changes

### DIFF
--- a/cedar-lean/Protobuf/BParsec.lean
+++ b/cedar-lean/Protobuf/BParsec.lean
@@ -21,9 +21,9 @@ Monadic Parser Combinator for ByteArrays
 -/
 
 namespace BParsec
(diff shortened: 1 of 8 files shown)


john-h-kastner-aws (approved):
Seems reasonable

PR #478, opened 2024-11-19
add Lean protobuf library

cdisselkoen:
First step towards merging the code in #400.  I thought it would make it easier to split #400 into some smaller chunks.  This self-contained PR has only the general-purpose Lean protobuf-parsing library, which generally implements protobuf parsing.  Not included in this PR is any functionality specific to Cedar, e.g., parsing Cedar structures from protobuf formats.

All of this code is credit to @Brandon-Rozek.  I simply rebased on latest main, updated for latest Lean, renamed a couple things, small tweaks like that.

PR #400, opened 2024-07-29
Lean Protobuffers

Brandon-Rozek:
*Issue #, if available:* N/A

*Description of changes:*

Initial work on adding Protobuffer support to lean

PR #483, opened 2024-11-21
[cedar-lean] actually run protobuf tests

cdisselkoen:
Protobuf tests introduced in #478 were not actually running in CI, and in fact were broken in the version of the PR that was merged.  This PR corrects that, and ensures that the tests are indeed checked in CI.

~Note that Brandon wrote the tests using `#guard` rather than our unit test framework, so this PR has no changes to e.g. `UnitTest/Main.lean`; it's sufficient to ensure that the `UnitTest/Proto.lean` file is imported.~ Edit: also updated the tests to use our unit test framework.

While we're at it, cleans up a proof in `UnitTest/Proto.lean`, and introduces a helper function `Field.guardedParse` which will be used both by `UnitTest/Proto.lean` and also the `CedarProto` package from #400 which I'll put up a cleaned-up PR for shortly.

### DIFF
--- a/cedar-lean/UnitTest/Proto.lean
+++ b/cedar-lean/UnitTest/Proto.lean
@@ -22,12 +22,13 @@ import Protobuf.Message
 import Protobuf.Enum
 import Protobuf.Varint
 import Protobuf.Packed
+import UnitTest.Run
 
 /-! Test Cases for Protobuffer functions -/
 
 open Proto
 
--- Show DecidableEquality of Except for unit tests
+-- Show DecidableEq of Except for unit tests
 namespace Except
 instance [DecidableEq α] [DecidableEq β] : DecidableEq (Except α β) := by
   unfold DecidableEq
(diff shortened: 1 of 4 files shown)


john-h-kastner-aws (approved):

cedar-lean/UnitTest/Proto.lean
```diff
@@ -32,19 +32,50 @@ namespace Except
...
+instance : Message HardCodeStruct := {
+  parseField := parseField
+  merge := merge
+}
+
+end HardCodeStruct
+
+
 
 #guard (@Field.interpret? Bool) (ByteArray.mk #[0]) = Except.ok false
```
emina, line 80: Is there a reason for why we don't want to use our unit testing framework here instead of `#guard`?

cdisselkoen: What is the advantage of our unit testing framework over `#guard` (or `theorem ... := by native_decide`)?  The latter two will fail at compile time, which seems better than waiting until the unit tests are run?  (Obviously, this doesn't work for tests that require IO, but for tests that don't require IO ...)

emina: Yeah, the main reason for our unit testing framework is to cover use cases like requiring IO. Another potential advantage is that the tests don't immediately fail the execution, but act more like a regular unit testing framework where results are gathered and then presented in the end. 

I don't have strong feelings here other than liking consistency :)

If we want to keep these static, maybe turn them into `theorem` so it consistent with other tests?

cdisselkoen: > Another potential advantage is that the tests don't immediately fail the execution

Why is this an advantage?

> maybe turn them into `theorem`

Definitely can do

emina: > Why is this an advantage?

If you have N failing tests, and the framework fails after the first one, you won't see the next one fail after you fix the first one.  I prefer to see all the failures up front (similar to how Lean tries to continue compiling as much of your code as it can after encountering the first failure).  That's just a personal preference though.

cdisselkoen: That's fair, and a good reason to convert

cdisselkoen: Refactored

emina (approved):

PR #489, opened 2024-12-10
ignore the `run_all_tests` test in benchmark.rs

cdisselkoen:
This test doesn't actually test anything for correctness, it just prints timing results.  We don't need to run it by default in CI.  So, this PR adds the `#[ignore]` attribute.  The test can still be run on demand, and CI will still ensure it builds.


cedar-drt/tests/benchmark.rs
```diff
@@ -164,6 +164,11 @@ fn print_summary(auth_times: HashMap<&str, Vec<f64>>, val_times: HashMap<&str, V
 }
 
 #[test]
+#[ignore] // doesn't test anything, just prints timing results -- run this on
+          // demand but not in CI.
```
adpaco-aws, line 168: Can we please rename the test then?

adpaco-aws (approved):

john-h-kastner-aws (approved):

PR #492, opened 2024-12-14
Changes needed for RFC 48

shaobo-he-aws:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cedar-drt/fuzz/src/schemas.rs
+++ b/cedar-drt/fuzz/src/schemas.rs
@@ -189,6 +189,12 @@ impl<K: Eq + Ord + Display, V: Equiv> Equiv for BTreeMap<K, V> {
     }
 }
 
+impl<N: Clone + PartialEq + Debug + Display + TypeName + Ord> Equiv for json_schema::CommonType<N> {
+    fn equiv(lhs: &Self, rhs: &Self) -> Result<(), String> {
+        Equiv::equiv(&lhs.ty, &rhs.ty)
+    }
+}
+
 impl<N: Clone + PartialEq + Debug + Display + TypeName + Ord> Equiv for json_schema::EntityType<N> {
     fn equiv(lhs: &Self, rhs: &Self) -> Result<(), String> {
         Equiv::equiv(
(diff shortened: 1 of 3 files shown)


cedar-policy-generators/src/schema.rs
```diff
@@ -152,6 +153,7 @@ fn arbitrary_typeofattribute_with_bounded_depth<N: From<ast::Name>>(
     Ok(json_schema::TypeOfAttribute {
         ty: arbitrary_schematype_with_bounded_depth::<N>(settings, entity_types, max_depth, u)?,
         required: u.arbitrary()?,
+        annotations: Annotations::new(),
```
john-h-kastner-aws, line 156: could generate some arbitrary annotations here for the schema roundtrip targets

shaobo-he-aws: Yeah. I will create an issue and set up another PR for it. This PR is to ensure RFC 48 doesn't cause issues for existing properties.

john-h-kastner-aws (approved):

cedar-policy-generators/src/schema.rs
shaobo-he-aws: Issue: https://github.com/cedar-policy/cedar-spec/issues/493

cedar-drt/fuzz/src/schemas.rs
```diff
@@ -189,6 +189,12 @@ impl<K: Eq + Ord + Display, V: Equiv> Equiv for BTreeMap<K, V> {
     }
 }
 
+impl<N: Clone + PartialEq + Debug + Display + TypeName + Ord> Equiv for json_schema::CommonType<N> {
+    fn equiv(lhs: &Self, rhs: &Self) -> Result<(), String> {
+        Equiv::equiv(&lhs.ty, &rhs.ty)
+    }
```
cdisselkoen, line 195: should we check more than just the `.ty` fields?  is the other field annotations, should we check that the annotations are the same?

cdisselkoen (approved):

cedar-drt/fuzz/src/schemas.rs
shaobo-he-aws: We'll do it later: https://github.com/cedar-policy/cedar-spec/pull/492#discussion_r1887196692

PR #498, opened 2024-12-18
Add dockerignore to 3.4.x

john-h-kastner-aws:
Avoids copying build artifacts into the docker image. We already have this on `main`, but it should be helpful for anyone else who might try to debug something on this version using the docker container

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -17,5 +17,5 @@ cd ..
 
 # Build DRT
 cd cedar-drt
-cargo fuzz build
+cargo fuzz build -s none
 cargo test


build.sh
```diff
@@ -17,8 +17,4 @@ cd ..
 
 # Build DRT
 cd cedar-drt
-cargo build
-
-# Build inner fuzz crate
-cd fuzz && RUSTFLAGS="--cfg=fuzzing" cargo build
-cargo test
+cargo fuzz build
```
cdisselkoen, line 20: should we still `cargo test`?

cdisselkoen (approved):

build.sh
john-h-kastner-aws: yeah, looks like this is also used in CI, so it should stay

shaobo-he-aws (approved):

PR #500, opened 2024-12-18
[cedar-lean/Protobuf] fix bug parsing empty maps

cdisselkoen:
Fix incorrect logic when parsing a `Proto.Map` that has no entries.  While we're here, added an `HAppend` instance for `Proto.Map` to clean up some calling code slightly


cdisselkoen:
Possible that it only fails on a policyset like this when the `LiteralPolicySet` is used as an embedded `Message` in some larger message (like `AuthorizationRequest`).  Which would make this hard to test in a unit test, since none of the types in `cedar-policy` have protobuf representations that include a `LiteralPolicySet` as an embedded `Message` (right?)

john-h-kastner-aws (approved):
Changes look reasonable, though I don't understand the details of the problems you mention in comments. Does it make sense to merge this as-is given that should fix at least one problem?

cdisselkoen:
I do think we should merge as-is, even though the added test doesn't catch the problem this is fixing, it is definitely fixing a problem

shaobo-he-aws (approved):

PR #501, opened 2024-12-18
Enable generation of arbitrary annotations and test their equivalence

shaobo-he-aws:
*Issue #, if available:*

*Description of changes:*

Depend on https://github.com/cedar-policy/cedar/pull/1382

### DIFF
--- a/cedar-drt/fuzz/src/schemas.rs
+++ b/cedar-drt/fuzz/src/schemas.rs
@@ -109,7 +109,7 @@ impl Equiv for Option<cedar_policy_core::ast::Annotation> {
             }
             (None, None) => return Ok(()),
         };
-        Err(format!("two annotations mismatch: {lhs:?} and {rhs:?}"))
+        Err(format!("{lhs:?} and {rhs:?} are not equivalent"))
     }
 }
 
--- a/cedar-drt/fuzz/src/schemas.rs
+++ b/cedar-drt/fuzz/src/schemas.rs
@@ -88,6 +88,31 @@ impl<'a, T: Equiv> Equiv for &'a T {
     }
 }
 
+impl Equiv for cedar_policy_core::est::Annotations {
+    fn equiv(lhs: &Self, rhs: &Self) -> Result<(), String> {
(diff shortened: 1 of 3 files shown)


cedar-drt/fuzz/src/schemas.rs
```diff
@@ -88,13 +88,40 @@ impl<'a, T: Equiv> Equiv for &'a T {
...
+                }
+            }
+            (Some(a), None) | (None, Some(a)) => {
+                if a.val.is_empty() {
+                    return Ok(());
+                }
+            }
+            (None, None) => return Ok(()),
+        };
+        Err(format!("two annotations mismatch: {lhs:?} and {rhs:?}"))
```
adpaco-aws, line 112:
```suggestion
        Err(format!("{lhs:?} and {rhs:?} are not equivalent"))
```

adpaco-aws: The error is already wrapped in `mismatch in namespace annotations: {e}` so there's no need to repeat the description.

cedar-drt/fuzz/src/schemas.rs
```diff
@@ -517,3 +552,70 @@ impl Equiv for cedar_policy_validator::ValidatorSchema {
...
+                serde_json::from_value(serde_json::json!({"b": "🥯🍩", "a": "🥨"})).unwrap(),
+            ),
+        ];
+        pairs
+            .iter()
+            .for_each(|(a, b)| assert!(Equiv::equiv(a, b).is_ok()));
+
+        // negative cases
+        let pairs: [(Annotations, Annotations); 5] = [
+            (
```
adpaco-aws, line 596: Nit: we could have this be a separate `#[test]`

john-h-kastner-aws (approved):

cedar-drt/fuzz/src/schemas.rs
shaobo-he-aws: Will do in another PR.

adpaco-aws (approved):

PR #502, opened 2024-12-19
Try to ignore path dependencies for depndabot

john-h-kastner-aws:
Depndabot failed because it couldn't find the dependencies. They're path dependencies that we assume will be in `./cedar`, but that's not populated for dependabot. Don't know if this will work. AFAIK the only way to check is merging


adpaco-aws (approved):

john-h-kastner-aws:
tried this on a for and it didn't work

PR #504, opened 2024-12-30
Update typechecking to construct type annoted AST

john-h-kastner-aws:
*Issue #, if available:*

Adds a new `TypedExpr` type which is constructed by the updated typechecking procedure and updates the soundness proof for these changes.

*Description of changes:*

### DIFF
--- a/cedar-lean/Cedar/Thm/Typechecking.lean
+++ b/cedar-lean/Cedar/Thm/Typechecking.lean
@@ -51,8 +51,8 @@ theorem typecheck_is_sound (policy : Policy) (env : Environment) (t : CedarType)
   split at h₂ <;> simp at h₂
   rename_i ht
   have hc := empty_capabilities_invariant request entities
-  split_type_of h₃ ; rename_i h'₃
-  have ⟨_, v, h₄, h₅⟩ := type_of_is_sound hc h₁ h'₃
+  split_type_of h₃ ; rename_i h₃
+  have ⟨_, v, h₄, h₅⟩ := type_of_is_sound hc h₁ h₃
   simp at h₂
   rw [h₂] at ht h₅
   have ⟨b, h₆⟩ := instance_of_type_bool_is_bool v t h₅ ht
--- a/cedar-lean/Cedar/Thm/Typechecking.lean
+++ b/cedar-lean/Cedar/Thm/Typechecking.lean
@@ -26,7 +26,7 @@ open Cedar.Spec
 open Cedar.Validation
 
 def typecheck (policy : Policy) (env : Environment) : Except TypeError CedarType := do
-  let (ty, _) ← typeOf policy.toExpr ∅ env
+  let (ty, _) ← (typeOf policy.toExpr ∅ env).typeOf
   if ty ⊑ .bool .anyBool
   then .ok ty
   else .error (.unexpectedType ty)
@@ -47,11 +47,13 @@ theorem typecheck_is_sound (policy : Policy) (env : Environment) (t : CedarType)
 := by
   intro h₁ h₂
   simp [typecheck] at h₂
-  cases h₃ : typeOf (Policy.toExpr policy) [] env <;> simp [h₃] at h₂
+  cases h₃ : (typeOf (Policy.toExpr policy) [] env).typeOf <;> simp [h₃] at h₂
   split at h₂ <;> simp at h₂
   rename_i ht
   have hc := empty_capabilities_invariant request entities
-  have ⟨_, v, h₄, h₅⟩ := type_of_is_sound hc h₁ h₃
+  split_type_of h₃ ; rename_i h'₃
+  have ⟨_, v, h₄, h₅⟩ := type_of_is_sound hc h₁ h'₃
+  simp at h₂
   rw [h₂] at ht h₅
   have ⟨b, h₆⟩ := instance_of_type_bool_is_bool v t h₅ ht
   subst h₆
--- a/cedar-lean/Cedar/Thm/Validation/Typechecker/Call.lean
+++ b/cedar-lean/Cedar/Thm/Validation/Typechecker/Call.lean
@@ -175,7 +175,7 @@ theorem type_of_call_decimal_comparator_inversion {xfn : ExtFun} {xs : List Expr
     split at h₁ <;> try { contradiction }
     all_goals {
       simp [ok] at h₁
-      obtain ⟨ h₁ₗ, h₁ᵣ ⟩ := h₁
+      have ⟨ h₁ₗ, h₁ᵣ ⟩ := h₁
       subst h₁ₗ h₁ᵣ
       simp only [List.empty_eq, true_and, TypedExpr.typeOf]
       rename_i h₃
@@ -184,7 +184,7 @@ theorem type_of_call_decimal_comparator_inversion {xfn : ExtFun} {xs : List Expr
       cases tys <;> try simp at h₃
       rename_i tys
       cases tys <;> try simp at h₃
-      obtain ⟨ h₃ₗ, h₃ᵣ ⟩ := h₃
+      have ⟨ h₃ₗ, h₃ᵣ ⟩ := h₃
       rw (config := {occs := .pos [1]}) [←h₃ₗ]
       rw [←h₃ᵣ]
       apply typeOf_of_binary_call_inversion h₂
@@ -259,7 +259,7 @@ theorem type_of_call_isInRange_inversion {xs : List Expr} {c c' : Capabilities} 
     cases tys <;> try simp at h₃
     rename_i tys
     cases tys <;> try simp at h₃
-    obtain ⟨ h₃ₗ, h₃ᵣ ⟩ := h₃
+    have ⟨ h₃ₗ, h₃ᵣ ⟩ := h₃
     rw (config := {occs := .pos [1]}) [←h₃ₗ]
     rw [←h₃ᵣ]
     apply typeOf_of_binary_call_inversion h₂
--- a/cedar-lean/Cedar/Thm/Validation/Typechecker/Call.lean
+++ b/cedar-lean/Cedar/Thm/Validation/Typechecker/Call.lean
@@ -26,9 +26,9 @@ open Cedar.Data
 open Cedar.Spec
 open Cedar.Validation
 
-theorem type_of_call_decimal_inversion {xs : List Expr} {c c' : Capabilities} {env : Environment} {ty : CedarType}
+theorem type_of_call_decimal_inversion {xs : List Expr} {c c' : Capabilities} {env : Environment} {ty : TypedExpr}
   (h₁ : typeOf (Expr.call .decimal xs) c env = Except.ok (ty, c')) :
-  ty = .ext .decimal ∧
+  ty.typeOf = .ext .decimal ∧
   c' = ∅ ∧
   ∃ (s : String),
     xs = [.lit (.string s)] ∧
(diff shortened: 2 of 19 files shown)


cedar-lean/Cedar/Validation/Typechecker.lean
```diff
@@ -59,74 +64,80 @@ def typeOfLit (p : Prim) (env : Environment) : ResultType :=
...
   match v with
   | .principal => ok (.entity env.reqty.principal)
   | .action    => ok (.entity env.reqty.action.ty)
   | .resource  => ok (.entity env.reqty.resource)
   | .context   => ok (.record env.reqty.context)
 
-def typeOfIf (r₁ : CedarType × Capabilities) (r₂ r₃ : ResultType) : ResultType :=
-  match r₁ with
+def typeOfIf (r₁ : TypedExpr × Capabilities) (r₂ r₃ : ResultType) : ResultType :=
+  match (r₁.fst.typeOf, r₁.snd) with
```
emina, line 75: I suggest using a parallel match here instead of creating a tuple and matching on it (Lean's automation sometimes doesn't like the tuples but is fine with the parallel match):

```
match r₁.fst.typeOf, r₁.snd with
| .bool .tt, c₁  => do ...
```

cedar-lean/Cedar/Validation/Typechecker.lean
```diff
@@ -59,74 +64,80 @@ def typeOfLit (p : Prim) (env : Environment) : ResultType :=
...
   | (.bool .tt, c₁) => do
     let (ty₂, c₂) ← r₂
```
emina, line 77: The `ty₂` threw me off here, because we use `ty` for CedarType values.  We may want to rename this to something like `tx` (typed expression).

cedar-lean/Cedar/Validation/Typechecker.lean
```diff
@@ -59,74 +64,80 @@ def typeOfLit (p : Prim) (env : Environment) : ResultType :=
...
+    match ty₂.typeOf ⊔ ty₃.typeOf with
+    | .some ty => ok (.ite r₁.fst ty₂ ty₃ ty) ((c₁ ∪ c₂) ∩ c₃)
+    | .none    => err (.lubErr ty₂.typeOf ty₃.typeOf)
   | (ty₁, _) => err (.unexpectedType ty₁)
 
-def typeOfAnd (r₁ : CedarType × Capabilities) (r₂ : ResultType) : ResultType :=
-  match r₁ with
-  | (.bool .ff, _)  => ok (.bool .ff)
+def typeOfAnd (r₁ : TypedExpr × Capabilities) (r₂ : ResultType) : ResultType :=
+  match (r₁.fst.typeOf, r₁.snd) with
```
emina, line 89: Consider using a parallel match here.

cedar-lean/Cedar/Validation/Typechecker.lean
```diff
@@ -59,74 +64,80 @@ def typeOfLit (p : Prim) (env : Environment) : ResultType :=
...
     | .bool _       => ok (.bool .anyBool) (c₁ ∪ c₂)
-    | _             => err (.unexpectedType ty₂)
+    | _             => err (.unexpectedType ty₂.typeOf)
   | (ty₁, _)        => err (.unexpectedType ty₁)
 
-def typeOfOr (r₁ : CedarType × Capabilities) (r₂ : ResultType) : ResultType :=
-  match r₁ with
-  | (.bool .tt, _)  => ok (.bool .tt)
+def typeOfOr (r₁ : TypedExpr × Capabilities) (r₂ : ResultType) : ResultType :=
+  match (r₁.fst.typeOf, r₁.snd) with
```
emina, line 102: Consider using a parallel match here.

cedar-lean/Cedar/Validation/Typechecker.lean
```diff
@@ -179,95 +190,118 @@ def typeOfHasTag (ety : EntityType) (x : Expr) (t : Expr) (c : Capabilities) (en
...
   | .sub,    .int, .int                     => ok .int
   | .mul,    .int, .int                     => ok .int
-  | .contains, .set ty₃, _                  => ifLubThenBool ty₂ ty₃
-  | .containsAll, .set ty₃, .set ty₄        => ifLubThenBool ty₃ ty₄
-  | .containsAny, .set ty₃, .set ty₄        => ifLubThenBool ty₃ ty₄
-  | _, _, _                                 => err (.unexpectedType ty₁)
-
-def hasAttrInRecord (rty : RecordType) (x : Expr) (a : Attr) (c : Capabilities) (knownToExist : Bool) : ResultType :=
+  | .contains, .set ty₃, _                  => do
+    let (ty, c) <- ifLubThenBool ty₂.typeOf ty₃
```
emina, line 222: Can we change `<-` to `←` for consistency?

cedar-lean/Cedar/Validation/Typechecker.lean
```diff
@@ -167,7 +178,7 @@ def typeOfInₛ (ety₁ ety₂ : EntityType) (x₁ x₂ : Expr) (env : Environme
     then .anyBool
     else .ff
 
-def typeOfHasTag (ety : EntityType) (x : Expr) (t : Expr) (c : Capabilities) (env : Environment) : ResultType :=
+def typeOfHasTag (ety : EntityType) (x : Expr) (t : Expr) (c : Capabilities) (env : Environment) : Except TypeError (CedarType × Capabilities)  :=
```
emina, line 181: I'm sad that `typeOfHasTag` and `typeOfGetTag` don't return `ResultType` like other `typeOf` functions :D 

(I see below why it's done, but the asymmetry makes me sad.)

cedar-lean/Cedar/Validation/Typechecker.lean
```diff
@@ -179,95 +190,118 @@ def typeOfHasTag (ety : EntityType) (x : Expr) (t : Expr) (c : Capabilities) (en
...
+  let ok ty₂ (c := ∅) := ok (TypedExpr.hasAttr  ty a ty₂) c
+  match ty.typeOf with
+  | .record rty => do
+    let (ty', c) ← hasAttrInRecord rty x a c true
+    ok ty' c
   | .entity ety =>
     match env.ets.attrs? ety with
-    | .some rty => hasAttrInRecord rty x a c false
+    | .some rty => do
+      let (ty', c) ← hasAttrInRecord rty x a c false
```
emina, line 249: This pattern repeats a lot in the code:

```
do 
  let (ty, c) ← ...
  ok ty c
```

I might be tempted to hide it in a helper function.  If we had a variant for `ok` that accepts a pair rather than two arguments, we could rewrite these as just one line, e.g.:

```
do ok (← hasAttrInRecord rty x a c false)
```

cedar-lean/Cedar/Validation/Typechecker.lean
john-h-kastner-aws: I actually found this easier since it caused a smaller change to the code and hence less change to the proof. In fact, matching in `r\1.snd` shouldn't be necessary at all since I don't deconstruct it. Perhaps the proof will ultimately end up simpler if I push the change through.

cedar-lean/Cedar/Validation/Typechecker.lean
john-h-kastner-aws: I was thinking about doing some renaming later. Maybe changing the alias for `Except TypeError (CedarType × Capabilities)` to `ResultTypedExpr` and keeping `ResultType` as `Except TypeError (CedarType × Capabilities)` so things look a bit more symmetric at these definitions.  I like small diffs though, and that kind of blows it up.

cedar-lean/Cedar/Validation/Typechecker.lean
john-h-kastner-aws: I tried to do something clever here but Lean's handling of default arguments was getting in the way. Might take another look

emina: Yeah, that's fair.  We can do a second pass after this one is merged.  

(It all looks good / correct to me.  These are just nits.)

cedar-lean/Cedar/Validation/Typechecker.lean
john-h-kastner-aws: Turns out to not a difficult change to make in isolation now that everything else is already fixed up

emina (approved):

cedar-lean/Cedar/Thm/Typechecking.lean
```diff
@@ -26,7 +26,7 @@ open Cedar.Spec
 open Cedar.Validation
 
 def typecheck (policy : Policy) (env : Environment) : Except TypeError CedarType := do
-  let (ty, _) ← typeOf policy.toExpr ∅ env
+  let (ty, _) ← (typeOf policy.toExpr ∅ env).typeOf
```
cdisselkoen, line 29: do we have two different functions both named `typeOf`?

cedar-lean/Cedar/Thm/Validation/Typechecker/Call.lean
```diff
@@ -172,29 +175,43 @@ theorem type_of_call_decimal_comparator_inversion {xfn : ExtFun} {xs : List Expr
     split at h₁ <;> try { contradiction }
     all_goals {
       simp [ok] at h₁
-      simp only [h₁, List.empty_eq, true_and]
+      obtain ⟨ h₁ₗ, h₁ᵣ ⟩ := h₁
```
cdisselkoen, line 178: is `obtain` different from `have`?

cedar-lean/Cedar/Thm/Validation/Typechecker/HasAttr.lean
```diff
@@ -180,13 +189,18 @@ theorem type_of_hasAttr_is_sound {x₁ : Expr} {a : Attr} {c₁ c₂ : Capabilit
...
     simp [EvaluatesTo] at h₆ <;>
     simp [EvaluatesTo, evaluate] <;>
+    rw [hl₄] at h₇ <;>
     rcases h₆ with h₆ | h₆ | h₆ | h₆ <;> simp [h₆]
-    <;> try exact type_is_inhabited ty
-    · exact type_of_hasAttr_is_sound_for_entities h₁ h₂ h₃ h₄ h₆ h₇
-    · exact type_of_hasAttr_is_sound_for_records h₁ h₃ h₄ h₆ h₇
+    <;> try exact type_is_inhabited ty.typeOf
+    · exact type_of_hasAttr_is_sound_for_entities h₁ h₂ h₃ (by simp [h'₄, ResultType.typeOf, Except.map]; exact h₄) h₆ h₇
+    · exact type_of_hasAttr_is_sound_for_records h₁ h₃ (by simp [h'₄, ResultType.typeOf, Except.map]; exact h₄) h₆ h₇
```
cdisselkoen, line 203: might be clearer to `apply` with an `_` in the place of the `(by )` expression, then put the `by` expression on the next line

cedar-lean/Cedar/Thm/Validation/Typechecker/LitVar.lean
```diff
@@ -26,10 +26,10 @@ namespace Cedar.Thm
 open Cedar.Spec
 open Cedar.Validation
 
-theorem type_of_lit_is_sound {l : Prim} {c₁ c₂ : Capabilities} {env : Environment} {ty : CedarType} {request : Request} {entities : Entities}
-  (h₃ : typeOf (Expr.lit l) c₁ env = Except.ok (ty, c₂)) :
+theorem type_of_lit_is_sound {l : Prim} {c₁ c₂ : Capabilities} {env : Environment} {e' : TypedExpr} {request : Request} {entities : Entities}
```
cdisselkoen, line 29: why `e'` instead of `e`?

cedar-lean/Cedar/Validation/TypedExpr.lean
```diff
@@ -0,0 +1,132 @@
+
+import Cedar.Spec
+import Cedar.Validation.Types
+
+namespace Cedar.Validation
+
+open Cedar.Data
+open Cedar.Spec
+
+inductive TypedExpr where
```
cdisselkoen, line 10: maybe a doc comment explaining what `ty` represents?

cdisselkoen (approved):
Looks great

cedar-lean/Cedar/Thm/Validation/Typechecker/Call.lean
john-h-kastner-aws: looks like they're almost the same but `obtain` can recursively deconstruct objects. We don't need that, so this can use `have` for consistency.

emina (approved):

PR #505, opened 2024-12-31
fix bug in protobuf map parsing

cdisselkoen:
This bug was introduced by #500.  #500 did indeed identify that the behavior on `len_size = 0` was incorrect, but unfortunately, #500's behavior is not correct either.  This PR has the correct behavior: when we have a map, but the len is 0, that's not empty map but rather a map with a single key-value pair where both key and value are default.

This also explains why the test case added in #500 wasn't actually broken before #500.  It tests the empty-map case, which is handled (correctly) elsewhere.  The new test case added in this PR is actually broken before this PR (and after #500), and fixed with this PR.


john-h-kastner-aws (approved):
Odd, but ok

shaobo-he-aws (approved):

PR #506, opened 2024-12-31
simplify Protobuf.Map.parse

cdisselkoen:
Following up on #505, this is a refactor to significantly simplify the implementation of `Protobuf.Map.parse`, which passes all tests including the regression tests introduced in #500 and #505.  It's a little bit more permissive than the old implementation, but in ways that I believe align with the spirit of the Protobuf spec, and won't affect the parsing of anything that used to parse with the old code.  

Also reindents this code, the old code was using 5-space tabs for some reason (!).  It now uses 2-space tabs like the rest of the file and other files in the module.

### DIFF
--- a/cedar-lean/Protobuf/Map.lean
+++ b/cedar-lean/Protobuf/Map.lean
@@ -46,59 +46,23 @@ instance [Field α] [Field β] : HAppend (Map α β) (Map α β) (Map α β) whe
 
 @[inline]
 def parse [Inhabited KeyT] [Inhabited ValueT] [Field KeyT] [Field ValueT] : BParsec (Map KeyT ValueT) := do
-     let len_size ← Len.parseSize
-     if len_size = 0 then
-          -- CAREFUL! You might think this means "empty map".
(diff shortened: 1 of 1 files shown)


john-h-kastner-aws (approved):

cedar-lean/Protobuf/Map.lean
```diff
@@ -46,59 +46,23 @@ instance [Field α] [Field β] : HAppend (Map α β) (Map α β) (Map α β) whe
...
-     let endPos ← BParsec.pos
-
-     if endPos - startPos != len_size then
-          throw s!"[Map parse] LEN size invariant not maintained: expected {len_size}, parsed {endPos - startPos}"
-
-     pure result
+  let len_size ← Len.parseSize
+  let startPos ← BParsec.pos
+
+  let mut key := (default : KeyT)
```
shaobo-he-aws, line 52: 🫣

shaobo-he-aws (approved):

PR #512, opened 2025-01-08
Upgrade to lean 4.15

emina:
This PR updates the Lean models and proofs to Lean version 4.15.

### DIFF
--- a/cedar-lean/Cedar/Thm/Data/List/Lemmas.lean
+++ b/cedar-lean/Cedar/Thm/Data/List/Lemmas.lean
@@ -111,14 +111,14 @@ theorem map₁_eq_map (f : α → β) (as : List α) :
   as.map₁ (λ x : {x // x ∈ as} => f x.val) =
   as.map f
 := by
-  simp [map₁, attach_def, map_pmap_subtype]
+  simp only [map₁, attach_def, map_pmap_subtype]
 
 
 theorem map_attach₂ {α : Type u} {β : Type v} [SizeOf α] [SizeOf β] {xs : List (α × β)} (f : (α × β) → γ) :
   xs.attach₂.map (λ x : { x : α × β // sizeOf x.snd < 1 + sizeOf xs } => f x.1) =
   xs.map f
 := by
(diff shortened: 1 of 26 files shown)


cedar-lean/Cedar/Thm/Data/List/Lemmas.lean
```diff
@@ -111,14 +111,14 @@ theorem map₁_eq_map (f : α → β) (as : List α) :
   as.map₁ (λ x : {x // x ∈ as} => f x.val) =
   as.map f
 := by
-  simp [map₁, attach_def, map_pmap_subtype]
+  simp only [map₁, attach_def, map_pmap_subtype]
```
adpaco-aws, line 114: Can't believe this `only` is all we needed 😅

adpaco-aws (approved):
Thanks!

cedar-lean/Cedar/Thm/Data/List/Lemmas.lean
emina: In this particular case, `simp` would have worked too, but I took the opportunity to harden it by using `simp only` instead :)

john-h-kastner-aws (approved):

cedar-lean/Cedar/Thm/Validation/Typechecker/LUB.lean
```diff
@@ -30,7 +30,7 @@ open Cedar.Data
 open Cedar.Spec
 open Cedar.Validation
 
-inductive IsLubOfRecordTypes : List (Attr × Qualified CedarType) → List (Attr × Qualified CedarType) → List (Attr × Qualified CedarType) → Prop :=
+inductive IsLubOfRecordTypes : List (Attr × Qualified CedarType) → List (Attr × Qualified CedarType) → List (Attr × Qualified CedarType) → Prop where
```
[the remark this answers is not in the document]
emina: Lean is changing the syntax, so any `inductive` that uses `:=` gets a warning to switch to `where`.

PR #517, opened 2025-01-15
Fix protobuf parsing for unrecognized extension functions

john-h-kastner-aws:
* Change protobuf parsing to return an error result instead of panicking on unknown extension functions
* Update fuzz targets to look for these error messages rather than errors messages that were specific to JSON parsing
* Call `lean_set_exit_on_panic` when initializing the lean runtime so that any other panics will actually cause tests to fail

This should fix the remaining issues in `abac-type-directed` and `validation-drt-type-directed` targets

### DIFF
--- a/cedar-lean/CedarProto/Expr.lean
+++ b/cedar-lean/CedarProto/Expr.lean
@@ -86,16 +86,16 @@ def parseField (t : Proto.Tag) : BParsec (MergeFn Prim) := do
   match t.fieldNum with
     | 1 =>
       let x : Bool ← Field.guardedParse t
-      pure (merge_bool · x)
+      pure (pure $ merge_bool · x)
     | 2 =>
       let x : Proto.Int64 ← Field.guardedParse t
-      pure (merge_int · x)
+      pure (pure $ merge_int · x)
     | 3 =>
       let x : String ← Field.guardedParse t
-      pure (merge_string · x)
+      pure (pure $ merge_string · x)
     | 4 =>
       let x : EntityUID ← Field.guardedParse t
-      pure (merge_euid · x)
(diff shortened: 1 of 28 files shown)


shaobo-he-aws (approved):
LGTM. Though having two `pure`s looks a little bit weird.

cedar-lean/CedarProto/Expr.lean
```diff
@@ -915,56 +915,56 @@ partial def Proto.ExprKind.parseField (t : Proto.Tag) : BParsec (Proto.ExprKind
   match t.fieldNum with
     | 1 =>
       let x : Prim ← Field.guardedParse t
-      pure (Proto.ExprKind.mergePrim · x)
+      pure (pure $ Proto.ExprKind.mergePrim · x)
```
cdisselkoen, line 918: here and elsewhere: kinda weird to mix the `f (expr)` and `f $ expr` styles.  Maybe change this to `pure $ pure $ Proto.ExprKind.mergePrim \. x`?

cdisselkoen (approved):
Definitely better than the status quo.  I also kinda dislike having the doubled `pure` and it feels like we should be able to flatten that, but I don't feel too strongly

cedar-lean/CedarProto/Expr.lean
john-h-kastner-aws: It would be something like `pure $ pure ∘ (Proto.ExprKind.mergePrim · x)` which to me at least is harder to mentally parse that one `pure` operators on the function return and the other on the function

cdisselkoen: Understood

PR #519, opened 2025-01-15
Lean: Add parsing for `datetime` in the `datetime` extension

adpaco-aws:
*Issue #, if available:*

*Description of changes:*

This PR adds parsing for `datetime` strings in the `datetime` extension. See [RFC 80: datetime extension](https://cedar-policy.github.io/rfcs/0080-datetime-extension.html), its [Rust implementation](https://github.com/cedar-policy/cedar/pull/1276), and #471 which added parsing for `duration` in Lean.

The parsing procedure in this PR leverages the new datetime functionality in Lean 4 (see https://github.com/leanprover/lean4/pull/4904). The `datespec` macro, which allows users to define custom formats, is used here to define the five datetime formats specified in the RFC. As a consequence, **this implementation can handle some cases (e.g., leap seconds) which the Rust implementation does not handle at the moment**. I've marked such cases with a comment `TODO: LEAN-RUST DIFF` so they are easier to locate in the PR, and I'm hoping that the discussion in this PR helps me understand what cases we need to artificially reject in order to achieve parity with the Rust implementation.

When writing tests for dates before 1970, I noticed some confusing results. The folks working on Lean 4 have confirmed to me in private that there are some bugs in such dates, and they're working on fixes now. See the test marked with `TODO: Bug in dates before 1970` for an example.

### DIFF
--- a/cedar-lean/Cedar/Spec/Ext/Datetime.lean
+++ b/cedar-lean/Cedar/Spec/Ext/Datetime.lean
@@ -28,11 +28,16 @@ namespace Datetime
 /--
   A datetime value is measured in milliseconds and constructed from a datetime string.
   A datetime string must be of one of the forms:
-  - `YYYY-MM-DD` (date only)
-  - `YYYY-MM-DDThh:mm:ssZ` (UTC)
-  - `YYYY-MM-DDThh:mm:ss.SSSZ` (UTC with millisecond precision)
-  - `YYYY-MM-DDThh:mm:ss(+|-)hhmm` (With timezone offset in hours and minutes)
-  - `YYYY-MM-DDThh:mm:ss.SSS(+|-)hhmm` (With timezone offset in hours and minutes and millisecond precision)
+    - `YYYY-MM-DD` (date only)
+    - `YYYY-MM-DDThh:mm:ssZ` (UTC)
+    - `YYYY-MM-DDThh:mm:ss.SSSZ` (UTC with millisecond precision)
+    - `YYYY-MM-DDThh:mm:ss(+|-)hhmm` (With timezone offset in hours and minutes)
+    - `YYYY-MM-DDThh:mm:ss.SSS(+|-)hhmm` (With timezone offset in hours and minutes and millisecond precision)
+
+  Regardless of the timezone, offset is always normalized to UTC.
+
+  The datetime type does not provide a way to create a datetime from a Unix timestamp.
+  One of the readable formats listed above must be used instead.
 -/
 abbrev Datetime := Int64
 
@@ -59,24 +64,6 @@ def parse (str: String) : Option Datetime :=
   | Except.ok val => datetime? val.toTimestamp.toMillisecondsSinceUnixEpoch.toInt
   | _ => none
 
-#eval DateOnly.parse "2022-10-10"
-#eval DateUTC.parse "2022-10-10T00:00:00Z"
-#eval DateUTCWithMillis.parse "2022-10-10T03:35:00.001Z"
-#eval DateWithOffset.parse "2022-10-10T03:35:00+0500"
-#eval DateWithOffsetAndMillis.parse "2022-10-10T03:35:00.000+0500"
-
-#eval parse "2022-10-10"
-#eval parse "2022-10-10T00:00:00Z"
-#eval parse "2022-10-10T03:35:00.001Z"
-#eval parse "2022-10-10T03:35:00+0500"
-#eval parse "2022-10-10T03:35:00.000+0500"
-
-#eval "2022-10-10".length
-#eval "2022-10-10T00:00:00Z".length
-#eval "2022-10-10T03:35:00.001Z".length
-#eval "2022-10-10T03:35:00+05:00".length
-#eval "2022-10-10T03:35:00.000+05:00".length
--- #eval (parse "1969-12-31") == (Int64.ofInt (-86400000 :Int))
 /--
   A duration value is measured in milliseconds and constructed from a duration string.
   A duration string is a concatenated sequence of quantity-unit pairs where the quantity
--- a/cedar-lean/Cedar/Spec/Ext/Datetime.lean
+++ b/cedar-lean/Cedar/Spec/Ext/Datetime.lean
@@ -23,6 +23,60 @@ namespace Cedar.Spec.Ext
 
 open Cedar.Data
 
+namespace Datetime
+
+/--
+  A datetime value is measured in milliseconds and constructed from a datetime string.
+  A datetime string must be of one of the forms:
+  - `YYYY-MM-DD` (date only)
+  - `YYYY-MM-DDThh:mm:ssZ` (UTC)
+  - `YYYY-MM-DDThh:mm:ss.SSSZ` (UTC with millisecond precision)
+  - `YYYY-MM-DDThh:mm:ss(+|-)hhmm` (With timezone offset in hours and minutes)
+  - `YYYY-MM-DDThh:mm:ss.SSS(+|-)hhmm` (With timezone offset in hours and minutes and millisecond precision)
+-/
+abbrev Datetime := Int64
+
+def DateOnly : Std.Time.GenericFormat .any := datespec("uuuu-MM-dd")
+def DateUTC : Std.Time.GenericFormat .any := datespec("uuuu-MM-dd'T'HH:mm:ss'Z'")
+def DateUTCWithMillis : Std.Time.GenericFormat .any := datespec("uuuu-MM-dd'T'HH:mm:ss.SSS'Z'")
+def DateWithOffset : Std.Time.GenericFormat .any := datespec("uuuu-MM-dd'T'HH:mm:ssxx")
+def DateWithOffsetAndMillis : Std.Time.GenericFormat .any := datespec("uuuu-MM-dd'T'HH:mm:ss.SSSxx")
+
+abbrev datetime? := Int64.ofInt?
+
+def parse (str: String) : Option Datetime :=
+  let parseFun := match str.length with
+  | 10 => DateOnly.parse
+  | 20 => DateUTC.parse
+  | 24 => if str.get? ⟨23⟩ == some 'Z'
+          then DateUTCWithMillis.parse
+          else DateWithOffset.parse
+  | 28 => DateWithOffsetAndMillis.parse
+  | _ => (fun _ => Except.error "invalid string length")
+
+  let datetime := parseFun str
+  match datetime with
+  | Except.ok val => datetime? val.toTimestamp.toMillisecondsSinceUnixEpoch.toInt
+  | _ => none
+
+#eval DateOnly.parse "2022-10-10"
+#eval DateUTC.parse "2022-10-10T00:00:00Z"
+#eval DateUTCWithMillis.parse "2022-10-10T03:35:00.001Z"
+#eval DateWithOffset.parse "2022-10-10T03:35:00+0500"
+#eval DateWithOffsetAndMillis.parse "2022-10-10T03:35:00.000+0500"
+
+#eval parse "2022-10-10"
+#eval parse "2022-10-10T00:00:00Z"
+#eval parse "2022-10-10T03:35:00.001Z"
+#eval parse "2022-10-10T03:35:00+0500"
+#eval parse "2022-10-10T03:35:00.000+0500"
+
+#eval "2022-10-10".length
+#eval "2022-10-10T00:00:00Z".length
+#eval "2022-10-10T03:35:00.001Z".length
+#eval "2022-10-10T03:35:00+05:00".length
+#eval "2022-10-10T03:35:00.000+05:00".length
+-- #eval (parse "1969-12-31") == (Int64.ofInt (-86400000 :Int))
 /--
   A duration value is measured in milliseconds and constructed from a duration string.
   A duration string is a concatenated sequence of quantity-unit pairs where the quantity
@@ -40,8 +94,6 @@ open Cedar.Data
   A duration may be negative. Negative duration strings must begin with `-`.
 -/
 abbrev Duration := Int64
-
-namespace Datetime
 
 def MILLISECONDS_PER_SECOND: Int := 1000
 def MILLISECONDS_PER_MINUTE: Int := 60000
@@ -113,53 +165,5 @@ def Duration.parse (str : String) : Option Duration :=
 deriving instance Repr for Duration
 
 abbrev duration := Duration.parse
-
-abbrev Datetime := Int64
-
--- /-
---   * `YYYY-MM-DD` (date only)
---   * `YYYY-MM-DDThh:mm:ssZ"` (UTC)
---   * `"YYYY-MM-DDThh:mm:ss.SSSZ"` (UTC with millisecond precision)
---   * `"YYYY-MM-DDThh:mm:ss(+|-)hhmm"` (With timezone offset in hours and minutes)
---   * `"YYYY-MM-DDThh:mm:ss.SSS(+|-)hhmm"` (With timezone offset in hours and minutes and millisecond precision)
--- -/
-def DateOnly : Std.Time.GenericFormat .any := datespec("uuuu-MM-dd")
-def DateUTC : Std.Time.GenericFormat .any := datespec("uuuu-MM-dd'T'HH:mm:ss'Z'")
-def DateUTCWithMillis : Std.Time.GenericFormat .any := datespec("uuuu-MM-dd'T'HH:mm:ss.SSS'Z'")
-def DateWithOffset : Std.Time.GenericFormat .any := datespec("uuuu-MM-dd'T'HH:mm:ssxxx")
-def DateWithOffsetAndMillis : Std.Time.GenericFormat .any := datespec("uuuu-MM-dd'T'HH:mm:ss.SSSxxx")
-
-def parse (str: String) : Option Datetime :=
-  let parseFun := match str.length with
-  | 10 => DateOnly.parse
-  | 20 => DateUTC.parse
-  | 24 => DateUTCWithMillis.parse
-  | 25 => DateWithOffset.parse
-  | 29 => DateWithOffsetAndMillis.parse
-  | _ => (fun _ => Except.error "invalid string length")
-
-  let datetime := parseFun str
-  match datetime with
-  | Except.ok val => Int64.ofInt? val.toTimestamp.toMillisecondsSinceUnixEpoch.toInt
-  | _ => none
-
-#eval DateOnly.parse "2022-10-10"
-#eval DateUTC.parse "2022-10-10T00:00:00Z"
-#eval DateUTCWithMillis.parse "2022-10-10T03:35:00.001Z"
-#eval DateWithOffset.parse "2022-10-10T03:35:00+05:00"
-#eval DateWithOffsetAndMillis.parse "2022-10-10T03:35:00.000+05:00"
-
-#eval parse "2022-10-10"
-#eval (parse "1969-12-31") == (Int64.ofInt (-86400000 :Int))
-#eval parse "2022-10-10T00:00:00Z"
-#eval parse "2022-10-10T03:35:00.001Z"
-#eval parse "2022-10-10T03:35:00+05:00"
-#eval parse "2022-10-10T03:35:00.000+05:00"
-
-#eval "2022-10-10".length
-#eval "2022-10-10T00:00:00Z".length
-#eval "2022-10-10T03:35:00.001Z".length
-#eval "2022-10-10T03:35:00+05:00".length
-#eval "2022-10-10T03:35:00.000+05:00".length
 
 end Datetime
--- a/cedar-lean/Cedar/Spec/Ext/Datetime.lean
+++ b/cedar-lean/Cedar/Spec/Ext/Datetime.lean
@@ -49,20 +49,14 @@ def DateWithOffsetAndMillis : Std.Time.GenericFormat .any := datespec("uuuu-MM-d
 
 abbrev datetime? := Int64.ofInt?
 
-def parse (str: String) : Option Datetime :=
-  let parseFun := match str.length with
-  | 10 => DateOnly.parse
-  | 20 => DateUTC.parse
-  | 24 => if str.get? ⟨23⟩ == some 'Z'
-          then DateUTCWithMillis.parse
-          else DateWithOffset.parse
-  | 28 => DateWithOffsetAndMillis.parse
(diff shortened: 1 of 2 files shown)


cdisselkoen (approved):

cedar-lean/Cedar/Spec/Ext/Datetime.lean
```diff
@@ -15,13 +15,55 @@
...
+def DateWithOffset : Std.Time.GenericFormat .any := datespec("uuuu-MM-dd'T'HH:mm:ssxx")
+def DateWithOffsetAndMillis : Std.Time.GenericFormat .any := datespec("uuuu-MM-dd'T'HH:mm:ss.SSSxx")
+
+abbrev datetime? := Int64.ofInt?
+
+def parse (str: String) : Option Datetime :=
+  let parseFun := match str.length with
+  | 10 => DateOnly.parse
+  | 20 => DateUTC.parse
+  | 24 => if str.get? ⟨23⟩ == some 'Z'
```
shaobo-he-aws, line 56: I'm confused by the Lean doc here:
```
Strings are indexed by [String.Pos](https://lean-lang.org/doc/reference/latest/Basic-Types/Strings/#String___Pos___mk), which internally records byte counts rather than character counts, and thus takes constant time.
Aside from 0, these should not be constructed directly, but rather updated using [String.next](https://lean-lang.org/doc/reference/latest/Basic-Types/Strings/#String___next) and [String.prev](https://lean-lang.org/doc/reference/latest/Basic-Types/Strings/#String___prev).
```
Does it mean we cannot `str.get? <23>`?

shaobo-he-aws (approved):
LGTM.

cedar-lean/Cedar/Spec/Ext/Datetime.lean
```diff
@@ -15,13 +15,55 @@
...
+
+def DateOnly : Std.Time.GenericFormat .any := datespec("uuuu-MM-dd")
+def DateUTC : Std.Time.GenericFormat .any := datespec("uuuu-MM-dd'T'HH:mm:ss'Z'")
+def DateUTCWithMillis : Std.Time.GenericFormat .any := datespec("uuuu-MM-dd'T'HH:mm:ss.SSS'Z'")
+def DateWithOffset : Std.Time.GenericFormat .any := datespec("uuuu-MM-dd'T'HH:mm:ssxx")
+def DateWithOffsetAndMillis : Std.Time.GenericFormat .any := datespec("uuuu-MM-dd'T'HH:mm:ss.SSSxx")
+
+abbrev datetime? := Int64.ofInt?
+
+def parse (str: String) : Option Datetime :=
```
emina, line 52:
```suggestion
def parse (str: String) : Option Datetime := do
  let val :=
    DateOnly.parse str <|>
    DateUTC.parse str <|>
    DateUTCWithMillis.parse str <|>
    DateWithOffset.parse str <|>
    DateWithOffsetAndMillis.parse str
  datetime? (← val.toOption).toTimestamp.toMillisecondsSinceUnixEpoch.toInt
```

emina: We can implement this more compactly using the "orElse" operator (`<|>`).  It tries all the parsers in turn until one works (or returns error if none work).

cedar-lean/UnitTest/Datetime.lean
```diff
@@ -23,77 +23,143 @@ namespace UnitTest.Datetime
...
+    testInvalidDatetime "2024-01-01T" "separator not needed",
+    testInvalidDatetime "2024-01-01Ta" "invalid time: character",
+    testInvalidDatetime "2024-01-01T01:" "invalid time: only hours",
+    testInvalidDatetime "2024-01-01T01:02" "invalid time: no seconds",
+    testInvalidDatetime "2024-01-01T01:02:0b" "invalid time: character",
+    testInvalidDatetime "2024-01-01T01::02:03" "invalid time: double colon",
+    testInvalidDatetime "2024-01-01T01::02::03" "invalid time: double colons",
+    testInvalidDatetime "2024-01-01T31:02:03Z" "invalid time: hour offset",
+    testInvalidDatetime "2024-01-01T01:60:03Z" "invalid time: minute offset",
+    -- testInvalidDatetime "2016-12-31T23:59:60Z" "invalid time: leap second", -- TODO: LEAN-RUST DIFF
```
emina, line 73: The problem here is that Rust doesn't accept the last ":60" because it's in general non-sensical to specify 60 seconds, right? 

If so, I think the question is not one of Lean vs Rust but how we want to handle this in the RFC too. 

Do we know what the Go implementation does?  

That could break the tie (to change either the Rust code or the Lean code and, potentially, update the RFC).

cedar-lean/UnitTest/Datetime.lean
```diff
@@ -23,77 +23,143 @@ namespace UnitTest.Datetime
...
+    testInvalidDatetime "2024-01-01T00:00:00x001Z" "typo in milliseconds separator",
+    testInvalidDatetime "2024-01-01T00:00:00.001ZZ" "double Zulu code w/ millis",
+    testInvalidDatetime "2024-01-01T00:00:00➕0000" "sign `+` is an emoji",
+    testInvalidDatetime "2024-01-01T00:00:00➖0000" "sign `-` is an emoji",
+    testInvalidDatetime "2024-01-01T00:00:00.0001Z" "fraction of seconds is 4 digits",
+    testInvalidDatetime "2024-01-01T00:00:00.001➖0000" "sign `+` is an emoji",
+    testInvalidDatetime "2024-01-01T00:00:00.001➕0000" "sign `-` is an emoji",
+    testInvalidDatetime "2024-01-01T00:00:00.001+00000" "timezone offset is 5 digits",
+    testInvalidDatetime "2024-01-01T00:00:00.001-00000" "timezone offset is 5 digits",
+    -- testInvalidDatetime "2016-12-31T00:00:00+1160" "timezone offset is ???", -- TODO: LEAN-RUST DIFF
```
emina, line 87: Same comment here on Lean vs Rust as above.  What does the Go implementation do?  And what behavior do we want in general (RFC / spec question).

cedar-lean/UnitTest/Datetime.lean
adpaco-aws: Well, it's a complex topic. It's true that in general a 60-second value is not expected. Ideally, the usage of 60-second in any application should be validated against a leap second database like [this one](https://github.com/nyx-space/hifitime/blob/master/data/leap-seconds.list) since leap seconds are declared on the fly. But this would force such applications to be "online", so in practice datetime library designers do either:
 * Forbid the usage of 60-second values
 * Allow the usage of 60-second values and assume that any leap second is valid.

I agree that we should specify the valid range for seconds and timezones in the RFC. Let me post below the results I've got from using the Go implementation in addition to other popular datetime frameworks.

cedar-lean/UnitTest/Datetime.lean
adpaco-aws: Same answer here, I'll post results on a separate comment.

cedar-lean/Cedar/Spec/Ext/Datetime.lean
adpaco-aws: Thanks, this looks much better!

adpaco-aws:
In order to compare different datetime frameworks, I've written nine corner cases and run them in those frameworks. In particular, I'm comparing:
 * our current implementation in Rust (branch [here](https://github.com/adpaco-aws/cedar/tree/datetime-corners))
 * this implementation in Lean (branch [here](https://github.com/adpaco-aws/cedar-spec/tree/datetime-corners))
 * the current implementation in Go (branch [here](https://github.com/adpaco-aws/cedar-go/tree/datetime-corners))
 * the `datetime` library in Python (Gist [here](https://gist.github.com/adpaco-aws/20a28e6f6213a70cf99602b61e577fc4))
 * the `datetime` library in Typescript (Gist [here](https://gist.github.com/adpaco-aws/ce23cb1febb9d989d38ea6373db6e785), or run [on the playground](https://www.typescriptlang.org/play/?#code/PTAEHUFMBsGMHsC2lQBd5oBYoCoE8AHSAZVgCcBLA1UABWgEM8BzM+AVwDsATAGiwoBnUENANQAd0gAjQRVSQAUCEmYKsTKGYUAbpGF4OY0BoadYKdJMoL+gzAzIoz3UNEiPOofEVKVqAHSKymAAmkYI7NCuqGqcANag8ABmIjQUXrFOKBJMggBcISGgoAC0oACCbvCwDKgU8JkY7p7ehCTkVDQS2E6gnPCxGcwmZqDSTgzxxWWVoASMFmgYkAAeRJTInN3ymj4d-jSCeNsMq-wuoPaOltigAKoASgAywhK7SbGQZIIz5VWCFzSeCrZagNYbChbHaxUDcCjJZLfSDbExIAgUdxkUBIursJzCFJtXydajBDIKMjJBhLHD6VAAYQYghQAG9FCUShkCOxUPkrqhKJxmABuDmciGQWAKbj8wSC4ZigC+wQQnHlaHpTJZBW8WuZkAA2gBdUAAXlAhvFJRUAEYAlgUIwNS0CFcpY1uNbQKyRJweXzQAAiABMAAZbQA2Uq2kOlADMtpwIfj+QArABOfKRsMALSD-El0sgsuD4dtAHZShHq0mw2H8vXG2GAvW80HQEr+CpHvoooHADLkoAAYhUAJLPEoAcgAomQ2NiCI45ML+Y8zMxIHOF-yMjoGNAKK5uHVIFPvSoQw6xkG9wejx3Xe61V7OT6-QH+aGI1Ww9H42Gyappm2btoW6xSjKX7lr+v51g2Tb1q29b5p23ZgL2gj9vyQ6jhO07bvAi7LsMa4blu85EbunD7oex6nueb4qPG15eHedGgIgGS8kob6+tyvLQT+1b-oBKagc2+bgUQxalt+lbVnBOCIQhLZtqhXagD2fbQIOI7jpOoCzpRxE-KRoDrsKFE7n6tFHnCDEXmAAAsDpfKADIzjgaBQpAABejQoCkyQsqg3r8f6gllhG0axgm8HNs2ADUEb1gW4IQbJQlRjGcYAUBiVIepHaadpWG6fygCg5HQFTEMQTmgGm179JAzB1LoKB+d8GD1MgAWcEFiKheFH5RfJsV5QlKnVm26VFlB0U5XFibKapzbIe2aFaRhOmBtVtC1fVTFgJGzWbgNZAHj5fWBUkQ2QGFfGjYG425fFq2FWGSW2ramZzZlC2vctgGxo2trrcVW1ldhoD7YdDUVqxgy9FoKLfFdvX+bdwXDU9AkveWE3vSpyW-Wl0mQSW2VE-lYZZipG0aehFm7VVNV1Q1AAcbm+f1g0hQ9kgfEGT5cZwPFBiN+PU29K0k-WSUpjm-0yYDhOy4B8tqShJXM5hMO4QZBEmfMJGrhZ5GEWQ1F2fRCiMZyKgZjzN0DXdAswpoHCoKUKSlJdVmgLR7D6FLkUEzFstTapivORmYYq5Tcnq3FIYZgVAEQzrUM7eVel4YZxkLqbZnm5Zm5Wzb9522eijGmKijJFw0oNF4S4-JAAAip44L5AAU3CoMQgpygqwoAJSj0KIzsm+gp4D63pvmqGqD+aLUSKA3cKAPQ+CuPYpvm+TioPiXiDwE6BjsQADyw-T33B-ekqoyoBooB95A4+L0fnIn2foAAAGVsS4rmYPyAAJKySASpAGH05CqFUjdm71EaKAMgXA6TykEI-fkOh4D2VnpyBQ8ptT6ACMkIiM4aSYD7n3CKAYKayU7N-M0AA+H+v80TqhoASfs6924sm3pAXuyA+74yflw7hgh4DuACNAeAzA+6ALHOHfknIoH41gZIrhK9ZGQHkYo5R+sKqcinFAvhuklRTkATo3+ei5EKKUcAgGVMjJQPmiWaxtj4H2MaDIxxRjLE0DNKEjKqsSygAAPzBjhnVDsX5DYTiDHYo+DiDFOMfr4lhypggYM4Fg1AOCn7BBUDgZoHgyBeEQERZwwJeSOjcBudgDBNz8FgIeWAiQGDAj0H6YMM5VgMEQAsfQHYiLBnAA4VAU5hAADlIASCDEEFQN8vhkHeCyfgm4jioEcDKcYC8nA1J0MMR0LI0SIGhMIS47kJBEWiCIAwHBFyMBYGwLg3AgglCAA))

I've executed the tests for each framework and compiled the results into the Markdown table below. Note that ✅ means that the test passed, ❌ means it failed for some reason.


| Test Case | Description | Rust | Lean | Go | Python | TypeScript |
|-----------|-------------|------|------|-----|---------|------------|
| 1 | Last leap second: `2016-12-31T23:59:60Z`     | ❌ | ✅ | ✅ | ❌ | ❌ |
| 2 | Invalid leap second: `2017-06-30T23:59:60Z`  | ❌ | ✅ | ✅ | ❌ | ❌ |
| 3 | Invalid minute: `2017-06-30T23:60:00Z`       | ❌ | ❌ | ✅ | ❌ | ❌ |
| 4 | CET timezone: `2016-12-31T00:00:00+0100`     | ✅ | ✅ | ✅ | ✅ | ✅ |
| 5 | Negative zero TZ: `2016-12-31T00:00:00-0000` | ✅ | ✅ | ✅ | ✅ | ✅ |
| 6 | General TZ: `2016-12-31T00:00:00+1159`       | ✅ | ✅ | ✅ | ✅ | ✅ |
| 7 | General TZ: `2016-12-31T00:00:00+1500`       | ❌ | ✅ | ✅ | ✅ | ✅ |
| 8 | Leap minute TZ: `2016-12-31T00:00:00+2360`   | ❌ | ✅ | ✅ | ❌ | ❌ |
| 9 | Out-of-range TZ: `2016-12-31T00:00:00+2490`  | ❌ | ✅ | ✅ | ❌ | ❌ |

### Observations

We should all discuss these results, but first let me add my observations:
 1. The Python and TypeScript columns agree in all test cases. It's important to take these frameworks into account given their popularity. They also agree with our current Rust implementation except in one case: the +1500 general timezone.
 2. The Lean and Go implementations seem too permissive. They accept both of the leap-second test cases (note: test case num. 2 is not an actual leap second although it was considered at some point). In fact, Go goes as far as to accept a "leap minute".
 3. All of them accept the case with a negative zero offset (test case num. 5). This was actually quite surprising to me, since [ISO 8601 specifies it as invalid](https://en.wikipedia.org/wiki/ISO_8601#Other_time_offset_specifications):
 > It is not permitted to state a zero value time offset with a negative sign, as "−00:00", "−0000", or "−00".

Given this experience, I would recommend changing the RFC to reject 60-second values, and allow timezone offsets in between +24:00 and -24:00. This should result in the following implementation changes:
 * Tweak our Lean models to reject (1), (2), (8) and (9).
 * Change our Rust implementation to accept (7)
 * Change the Go implementation to do all of above and also reject (3).

What do you think about these changes?

cedar-lean/Cedar/Spec/Ext/Datetime.lean
adpaco-aws: My interpretation is that `str.get? <23>` is unsafe. They want you to use the `next` and `prev` iterators to be safe from character/byte discrepancies.

cedar-lean/Cedar/Spec/Ext/Datetime.lean
```diff
@@ -15,13 +15,55 @@
...
+  | 10 => DateOnly.parse
+  | 20 => DateUTC.parse
+  | 24 => if str.get? ⟨23⟩ == some 'Z'
+          then DateUTCWithMillis.parse
+          else DateWithOffset.parse
+  | 28 => DateWithOffsetAndMillis.parse
+  | _ => (fun _ => Except.error "invalid string length")
+
+  let datetime := parseFun str
+  match datetime with
```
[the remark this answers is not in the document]
adpaco-aws: Thanks for the suggestion, but since I've changed to use the `orElse` operator this doesn't appear anymore 😬

shaobo-he-aws:
> They also agree with our current Rust implementation except in one case: the +1500 general timezone.

Very nice table. The Rust implementation used to be very permissive about timezones until I read the wikipedia page and decided to follow it. The current time zones range from [UTC−12:00](https://en.wikipedia.org/wiki/UTC%E2%88%9212:00) to [UTC+14:00](https://en.wikipedia.org/wiki/UTC%2B14:00).

shaobo-he-aws:
To expand on why Rust rejects leap seconds/minutes: The rationale is that Cedar can't determine their validity as pointed out in the RFC. Hence, it just rejects them, hoping the developer to rewrite the policy/attribute values to the equivalent times that have valid representations.

emina:
> Given this experience, I would recommend changing the RFC to reject 60-second values, and allow timezone offsets in between +24:00 and -24:00. This should result in the following implementation changes:
> 
> * Tweak our Lean models to reject (1), (2), (8) and (9).
> * Change our Rust implementation to accept (7)
> * Change the Go implementation to do all of above and also reject (3).
> 
> What do you think about these changes?

Thank you for putting together this detailed analysis!  

I agree that we should re-open the discussion on the RFC and update the proposal accordingly.

My vote is to follow the JavaScript and Python behavior, and make the changes proposed by @adpaco-aws.

adpaco-aws:
Just posted the update to the RFC in https://github.com/cedar-policy/rfcs/pull/94

cedar-lean/Cedar/Spec/Ext/Datetime.lean
```diff
@@ -15,13 +15,60 @@
...
+
+def DateOnly : Std.Time.GenericFormat .any := datespec("uuuu-MM-dd")
+def DateUTC : Std.Time.GenericFormat .any := datespec("uuuu-MM-dd'T'HH:mm:ss'Z'")
+def DateUTCWithMillis : Std.Time.GenericFormat .any := datespec("uuuu-MM-dd'T'HH:mm:ss.SSS'Z'")
+def DateWithOffset : Std.Time.GenericFormat .any := datespec("uuuu-MM-dd'T'HH:mm:ssxx")
+def DateWithOffsetAndMillis : Std.Time.GenericFormat .any := datespec("uuuu-MM-dd'T'HH:mm:ss.SSSxx")
+
+abbrev datetime? := Int64.ofInt?
+
+def date_contains_leap_seconds (str: String) : Bool :=
```
emina, line 55: Naming convention:  we use camelCase for function names.  See https://github.com/cedar-policy/cedar-spec/blob/main/cedar-lean/GUIDE.md

cedar-lean/Cedar/Spec/Ext/Datetime.lean
```diff
@@ -15,13 +15,60 @@
...
+def parse (str: String) : Option Datetime := do
+  if date_contains_leap_seconds str then failure
+  let val :=
+    DateOnly.parse str <|>
+    DateUTC.parse str <|>
+    DateUTCWithMillis.parse str <|>
+    DateWithOffset.parse str <|>
+    DateWithOffsetAndMillis.parse str
+
+  let zoned_time ← val.toOption
```
emina, line 67: Naming convention:  we use camelCase for variable names.  See https://github.com/cedar-policy/cedar-spec/blob/main/cedar-lean/GUIDE.md

cedar-lean/UnitTest/Datetime.lean
```diff
@@ -23,77 +23,152 @@ namespace UnitTest.Datetime
...
-theorem test10 : toString ((Duration.parse "2d5h20m").get!) = "175800000" := by native_decide
-theorem test11 : toString ((Duration.parse "-1d12h").get!) = "-90720000" := by native_decide
-theorem test12 : toString ((Duration.parse "-3h45m").get!) = "-3780000" := by native_decide
-theorem test13 : toString ((Duration.parse "1d1ms").get!) = "86400001" := by native_decide
-theorem test14 : toString ((Duration.parse "59m59s999ms").get!) = "3599999" := by native_decide
-theorem test15 : toString ((Duration.parse "23h59m59s999ms").get!) = "11879999" := by native_decide
-theorem test16 : toString ((Duration.parse "0d0h0m0s0ms").get!) = "0" := by native_decide
-
-private def testValid (str : String) (rep : Int) : TestCase IO :=
+theorem testDatetime1 : toString ((parse "2022-10-10").get!) = "1665360000000" := by native_decide
```
emina, line 26: Can we turn these theorem-tests into unit tests for consistency?

cedar-lean/Cedar/Spec/Ext/Datetime.lean
adpaco-aws: Apologies, I'm not sure why I stopped using camelCase last week 😭

cedar-lean/UnitTest/Datetime.lean
```diff
@@ -23,22 +23,30 @@ namespace UnitTest.Datetime
...
 
-theorem testDatetime1 : toString ((parse "2022-10-10").get!) = "1665360000000" := by native_decide
-theorem testDatetime2 : toString ((parse "1969-12-31").get!) = "-86400000" := by native_decide
-theorem testDatetime3 : toString ((parse "2024-10-15T11:38:02Z").get!) = "1728992282000" := by native_decide
-theorem testDatetime4 : toString ((parse "2024-10-15T11:38:02.101Z").get!) = "1728992282101" := by native_decide
-theorem testDatetime5 : toString ((parse "2024-10-15T11:38:02.101-1134").get!) = "1729033922101" := by native_decide
-theorem testDatetime6 : toString ((parse "2024-10-15T11:38:02.101+1134").get!) = "1728950642101" := by native_decide
-theorem testDatetime7 : toString ((parse "2024-10-15T11:38:02+1134").get!) = "1728950642000" := by native_decide
-theorem testDatetime8 : toString ((parse "2024-10-15T11:38:02-1134").get!) = "1729033922000" := by native_decide
+theorem testDatetime01 : toString ((parse "2022-10-10").get!) = "1665360000000" := by native_decide
```
emina, line 26: Sorry, I should have been clearer:  let's use the unit testing format instead of theorems for all tests and remove the theorems.

cedar-lean/Cedar/Spec/Ext/Datetime.lean
emina: No worries; I figured you went into Rust mode :D

cedar-lean/UnitTest/Datetime.lean
adpaco-aws: No problem, I was about to ask if that's what we wanted. I've removed theorem tests for both datetime and duration, so we only have unit tests now.

emina (approved):

PR #520, opened 2025-01-15
Fix entity manifest fuzz target

john-h-kastner-aws:
Another fuzz target that we weren't running. The target was failing because manifests don't support tags, so I've updated the target to ignore unsupported feature errors. This should be fine since it's an experimental feature. Alternatively, we could add another option on the generator to disable tag generation, but this was the easier fix for now. Eventually we should support tags in manifests


shaobo-he-aws (approved):
LGTM

PR #523, opened 2025-02-03
Compare tags in `ValidatorEntityType`

shaobo-he-aws:
*Issue #, if available:*

*Description of changes:*

We didn't compare tags in `impl Equiv for cedar_policy_validator::ValidatorEntityType`. This PR fixes it.


cdisselkoen (approved):
do we need to instead test whether the tag types are `Equiv`, not equal?

john-h-kastner-aws (approved):

shaobo-he-aws:
> do we need to instead test whether the tag types are `Equiv`, not equal?

We don't have `Equiv` implemented for `Type`. If we do, it will be equivalence checking like this PR.

PR #526, opened 2025-02-04
RFC 53 Lean model

shaobo-he-aws:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cedar-lean/Cedar/Validation/Types.lean
+++ b/cedar-lean/Cedar/Validation/Types.lean
@@ -61,12 +61,29 @@ abbrev QualifiedType := Qualified CedarType
 
 abbrev RecordType := Map Attr QualifiedType
 
+inductive InputEntitySchemaEntityKind where
+  | Standard (info: RecordType × (Option CedarType))
+  | Enum (choices: List String)
+
+structure InputEntitySchemaEntry where
+  ancestors: Cedar.Data.Set EntityType
+  kind: InputEntitySchemaEntityKind
+
 structure EntitySchemaEntry where
   ancestors : Cedar.Data.Set EntityType
   attrs : RecordType
   tags : Option CedarType
 
+def InputEntitySchemaEntry.toEntitySchemaEntry (as: InputEntitySchemaEntry) : EntitySchemaEntry :=
+  match as.kind with
+  | .Standard (attrs, tags) => ⟨ as.ancestors, attrs, tags ⟩
+  | .Enum _ => ⟨ as.ancestors, Map.empty, none ⟩
+
 abbrev EntitySchema := Map EntityType EntitySchemaEntry
+abbrev InputEntitySchema := Map EntityType InputEntitySchemaEntry
+
+def InputEntitySchema.toEntitySchema (as: InputEntitySchema) : EntitySchema :=
+  as.mapOnValues InputEntitySchemaEntry.toEntitySchemaEntry
 
 def EntitySchema.contains (ets : EntitySchema) (ety : EntityType) : Bool :=
   (ets.find? ety).isSome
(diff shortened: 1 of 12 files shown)


cedar-lean/Cedar/Validation/Types.lean
```diff
@@ -61,12 +61,33 @@ abbrev QualifiedType := Qualified CedarType
 
 abbrev RecordType := Map Attr QualifiedType
 
+structure EntityInfo where
```
emina, line 64: Thank you for the PR! 

This approach works but there are two disadvantages in terms of keeping the overall model minimal:

1.  It introduces quite a few auxiliary layers (definitions and operators) into the model, and it touches part of the model that should not be affected by the addition of enums (`Spec`).  
2. Having two parallel schemas, one of which is abandoned after the first checking pass, won't play nicely with the symbolic compilation code that also uses and depends on the schema data structure.

I propose that we take a different approach, similar to how enums are modeled in the symbolic compiler.

The approach involves changing `EntitySchemaEntry` as follows and adding the following auxiliary function to `EntitySchema`:

```
structure EntitySchemaEntry where
  ancestors : Cedar.Data.Set EntityType
  attrs : RecordType
  tags : Option CedarType
  members: Set String
  
def EntitySchema.isValidEntityUID (ets : EntitySchema) (uid : EntityUID) : Bool :=
  match ets.find? uid.ty with
  | .some d => d.members.isEmpty || d.members.contains uid.eid 
  | .none   => false
```

Here, we add the `.members` field to `EntitySchemaEntry` and we use the empty set to indicate that the Entity is _not_ an enum.  This works because we cannot have an empty enum.  If we feel that is too subtle, we can declare `.members` to be `Option (Set String)`, and then `EntitySchema.isValidEntityUID` uses `none` to decide that an entity is not an enum.

With this change, we won't have to add any new functions or datastructures to the rest of the model.  Instead, we'd make localized changes to the base-cases of the validation algorithms and proofs that examine entity literals.

emina (changes requested):

cedar-lean/Cedar/Validation/Types.lean
shaobo-he-aws: This is the approach I thought about. In fact, the protobufs interface uses a similar struct. The main reason I had the current design is that `EntitySchemaEntry` above cannot inherently guarantee well-formedness - specifically, that enumerated entity types do not have attributes or tags - without adding a predicate.

The current design also offers the advantage of a modular validation process. This allows us to first validate input well-formedness (such as checking for invalid UIDs of enumerated entity types) before proceeding with type checking. Since the type checking phase doesn't require the information specified by members, we can maintain existing proofs without modification.

Regarding the symbolic compiler, implementing this approach should be straightforward. We can maintain `EntitySchemaEntry` above while adding a simple conversion function from `InputEntitySchemaEntry` to `EntitySchemaEntry`.

emina: It's fine to add additional well-formedness constraints on `EntitySchemaEntry`s if they are necessary for the proofs.  I'm guessing these extra constraints are only needed for entity validation, not for type checking.  And that's okay.  We should aim to keep models as simple as possible.  It's okay to push some of the complexity into proofs / well-formedness definitions to keep models simple.

cdisselkoen: @emina are you saying that the invariant (enum entities don't have attributes/tags) is ok to be implicit in most of the code, and only a precondition for the proofs that need it?  In particular, the evaluator would effectively support attributes/tags on enum entities?

Is that at odds with one stated purpose of the Lean model, which is to be a clear, readable spec for Cedar behavior?  Structurally enforcing the invariant would make it crystal clear that the invariant exists, even in the evaluator code

cdisselkoen: Got it, in that case I agree

shaobo-he-aws: Following a discussion with @emina, we want to implement entity validation and type checking in one pass, rather than two passes implemented in this PR.. This decision is driven by our future plans to extend enumerated entities to support hierarchies, as outlined in the RFC.  This extension makes enumerated entities similar to actions. Having one pass unifies the treatment of actions and extended enumerated entities.

shaobo-he-aws:
Closed in favor of #530

PR #529, opened 2025-02-07
update to Lean 4.16

cdisselkoen:
Biggest change for us is (quoting from the [Lean changelog](https://github.com/leanprover/lean4/releases/tag/v4.16.0)):

> https://github.com/leanprover/lean4/pull/6379 replaces the inductive predicate `List.lt` with an upstreamed version of `List.Lex` from Mathlib.
(Previously `Lex.lt` was defined in terms of `<`; now it is generalized to take an arbitrary relation.)
This subtly changes the notion of ordering on `List α`.

> `List.lt` was a weaker relation: in particular if `l₁ < l₂`, then `a :: l₁ < b :: l₂` may hold according to `List.lt` even if `a` and `b` are merely incomparable (either neither `a < b` nor `b < a`), whereas according to `List.Lex` this would require `a = b`.

> When `<` is total, in the sense that `¬ · < ·` is antisymmetric, then the two relations coincide.

> Mathlib was already overriding the order instances for `List α`, so this change should not be noticed by anyone already using Mathlib.

> We simultaneously add the boolean valued `List.lex` function, parameterised by a `BEq` typeclass and an arbitrary `lt` function. This will support the flexibility previously provided for `List.lt`, via a `==` function which is weaker than strict equality.

Separately, Lean introduced its own `DecidableLT`.  For now, to get things working, I kept our `DecidableLT` and just qualified it as `Cedar.Data.DecidableLT` when ambiguous.  In a future PR we can look at replacing our `DecidableLT` with the Lean one.

### DIFF
--- a/cedar-lean/Cedar/Data/LT.lean
+++ b/cedar-lean/Cedar/Data/LT.lean
@@ -114,7 +114,7 @@ theorem List.slt_trans [LT α] [StrictLT α] {xs ys zs : List α} :
       apply List.Lex.cons
       exact List.slt_trans h₃ h₆
 
-theorem List.lt_asymm' [LT α] [StrictLT α] [Cedar.Data.DecidableLT α] {xs ys : List α} :
+theorem List.slt_asymm [LT α] [StrictLT α] [Cedar.Data.DecidableLT α] {xs ys : List α} :
   xs < ys → ¬ ys < xs
 := by
   intro h₁
@@ -171,7 +171,7 @@ theorem List.lt_conn [LT α] [StrictLT α] {xs ys : List α} :
         exact StrictLT.if_not_lt_eq_then_gt xhd yhd h₄ h₅
 
 instance List.strictLT (α) [LT α] [StrictLT α] [Cedar.Data.DecidableLT α] : StrictLT (List α) where
-  asymmetric _ _   := List.lt_asymm'
+  asymmetric _ _   := List.slt_asymm
   transitive _ _ _ := List.slt_trans
   connected  _ _   := List.lt_conn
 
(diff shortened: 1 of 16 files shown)


cedar-lean/Cedar/Thm/Data/LT.lean
```diff
@@ -164,13 +164,16 @@ instance Ext.strictLT : StrictLT Ext where
...
 instance Name.strictLT : StrictLT Name where
   asymmetric a b   := by
-    simp [LT.lt, Name.lt]
-    apply List.lt_asymm
+    simp only [LT.lt, Bool.not_eq_true]
+    simp only [Name.lt, decide_eq_true_eq, decide_eq_false_iff_not, List.not_lt]
+    apply List.lt_asymm'
   transitive a b c := by
-    simp [LT.lt, Name.lt]
+    simp only [LT.lt]
```
emina, line 171: Curious why we need to consecutive `simp only`s here.  Stylistic choice or Lean does the wrong thing if we use a single `simp only` that calls both sets of theorems?

cedar-lean/Cedar/Data/LT.lean
```diff
@@ -62,69 +70,51 @@ end Cedar.Data
...
-      have h₉ := StrictLT.if_not_lt_gt_then_eq xhd yhd h₃ h₄
-      subst h₉
-      apply List.lt.tail h₆ h₇
-      apply List.slt_trans h₅ h₈
+    case cons _ _ ztl h₆ =>
+      apply List.Lex.cons
+      exact List.slt_trans h₃ h₆
 
-theorem List.lt_asymm [LT α] [StrictLT α] {xs ys : List α} :
+theorem List.lt_asymm' [LT α] [StrictLT α] [Cedar.Data.DecidableLT α] {xs ys : List α} :
```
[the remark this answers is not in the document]
cdisselkoen: sure. What does the `s` stand for?

cedar-lean/Cedar/Thm/Data/LT.lean
cdisselkoen: Yes, if we combine it all then Lean throws `maximum recursion depth has been reached`

cdisselkoen: I think that `LT.lt` and `Name.lt` recurse infinitely with each other, so those are the two that needed separating

PR #530, opened 2025-02-07
RFC 53 Lean model and proofs

shaobo-he-aws:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cedar-lean/Cedar/Thm/Validation/Typechecker/BinaryApp.lean
+++ b/cedar-lean/Cedar/Thm/Validation/Typechecker/BinaryApp.lean
@@ -903,7 +903,10 @@ private theorem no_tags_type_implies_no_tags {uid : EntityUID} {env : Environmen
     replace ⟨e', h₂, h₃⟩ := h₂
     simp only [hf', Option.some.injEq] at h₂
     subst h₂
-    simp only [h₃] at h₁
+    simp only [EntitySchemaEntry.tags] at h₁
+    split at h₃
+    · simp only [h₃] at h₁
+      simp only [h₁, map_empty_contains_instance_of_ff]
     simp only [h₁, map_empty_contains_instance_of_ff]
   · exact map_empty_contains_instance_of_ff
 
@@ -1071,6 +1074,8 @@ theorem type_of_getTag_is_sound {x₁ x₂ : Expr} {c₁ c₂ : Capabilities} {e
   replace ⟨_, h₅, h₇⟩ := h₅
   simp only [hf₂, Option.some.injEq] at h₅
   subst h₅
+  simp only [EntitySchemaEntry.tags] at h₂
+  split at h₇
   simp only [h₇] at h₂
   have hf₃ := Map.findOrErr_returns d.tags s Error.tagDoesNotExist
   rcases hf₃ with ⟨v, hf₃⟩ | hf₃ <;>
@@ -1085,6 +1090,7 @@ theorem type_of_getTag_is_sound {x₁ x₂ : Expr} {c₁ c₂ : Capabilities} {e
     simp only [Entities.tagsOrEmpty, hf₁, Map.contains_iff_some_find?] at h₁
     replace ⟨_, h₁⟩ := h₁
     simp only [Map.findOrErr_err_iff_find?_none, h₁, reduceCtorEq] at hf₃
+  · cases h₇
 
 theorem type_of_binaryApp_is_sound {op₂ : BinaryOp} {x₁ x₂ : Expr} {c₁ c₂ : Capabilities} {env : Environment} {ty : TypedExpr} {request : Request} {entities : Entities}
   (h₁ : CapabilitiesInvariant c₁ request entities)
--- a/cedar-lean/Cedar/Validation/RequestEntityValidator.lean
+++ b/cedar-lean/Cedar/Validation/RequestEntityValidator.lean
@@ -158,7 +158,7 @@ def updateSchema (schema : Schema) (actionSchemaEntities : Entities) : Schema :=
       let entriesWithType := actionSchemaEntities.filter (λ k _ => k.ty == ty)
       let allAncestorsForType := List.flatten (entriesWithType.values.map (λ edt =>
         edt.ancestors.elts.map (·.ty) ))
-      let ese : EntitySchemaEntry := {
+      let ese : EntitySchemaEntry := .standard {
         ancestors := Set.make allAncestorsForType,
         attrs := Map.empty,
         tags := Option.none
--- a/cedar-lean/Cedar/Validation/RequestEntityValidator.lean
+++ b/cedar-lean/Cedar/Validation/RequestEntityValidator.lean
@@ -39,7 +39,11 @@ def instanceOfBoolType (b : Bool) (bty : BoolType) : Bool :=
   | _, .anyBool => true
   | _, _ => false
 
-def instanceOfEntityType (e : EntityUID) (ety : EntityType ) : Bool := ety == e.ty
+def instanceOfEntityType (e : EntityUID) (ety : EntityType ) (eids: Option (List String)) : Bool :=
+  ety == e.ty &&
+  match eids with
+  | .some eids => eids.contains e.eid
+  | _ => true
 
 def instanceOfExtType (ext : Ext) (extty: ExtType) : Bool :=
 match ext, extty with
@@ -52,17 +56,17 @@ match rty.find? k with
     | .some qty => if qty.isRequired then r.contains k else true
     | _ => true
 
-def instanceOfType (v : Value) (ty : CedarType) : Bool :=
+def instanceOfType (v : Value) (ty : CedarType) (schema: EntitySchema) : Bool :=
   match v, ty with
   | .prim (.bool b), .bool bty => instanceOfBoolType b bty
   | .prim (.int _), .int => true
   | .prim (.string _), .string => true
-  | .prim (.entityUID e), .entity ety => instanceOfEntityType e ety
-  | .set s, .set ty => s.elts.attach.all (λ ⟨v, _⟩ => instanceOfType v ty)
+  | .prim (.entityUID e), .entity ety => instanceOfEntityType e ety $ schema.getEIDRange ety
+  | .set s, .set ty => s.elts.attach.all (λ ⟨v, _⟩ => instanceOfType v ty schema)
   | .record r, .record rty =>
     r.kvs.all (λ (k, _) => rty.contains k) &&
     (r.kvs.attach₂.all (λ ⟨(k, v), _⟩ => (match rty.find? k with
-        | .some qty => instanceOfType v qty.getType
+        | .some qty => instanceOfType v qty.getType schema
         | _ => true))) &&
     rty.kvs.all (λ (k, _) => requiredAttributePresent r rty k)
   | .ext x, .ext xty => instanceOfExtType x xty
@@ -79,11 +83,11 @@ def instanceOfType (v : Value) (ty : CedarType) : Bool :=
         simp only [Map.mk.sizeOf_spec]
         omega
 
-def instanceOfRequestType (request : Request) (reqty : RequestType) : Bool :=
-  instanceOfEntityType request.principal reqty.principal &&
+def instanceOfRequestType (request : Request) (reqty : RequestType) (schema: EntitySchema) : Bool :=
+  instanceOfEntityType request.principal reqty.principal (schema.getEIDRange reqty.principal) &&
   request.action == reqty.action &&
-  instanceOfEntityType request.resource reqty.resource &&
-  instanceOfType request.context (.record reqty.context)
+  instanceOfEntityType request.resource reqty.resource (schema.getEIDRange reqty.resource) &&
+  instanceOfType request.context (.record reqty.context) schema
 
 /--
 For every entity in the store,
@@ -98,12 +102,12 @@ def instanceOfEntitySchema (entities : Entities) (ets : EntitySchema) : EntityVa
 where
   instanceOfEntityTags (data : EntityData) (entry : EntitySchemaEntry) : Bool :=
     match entry.tags with
-    | .some tty => data.tags.values.all (instanceOfType · tty)
+    | .some tty => data.tags.values.all (instanceOfType · tty ets)
     | .none     => data.tags == Map.empty
   instanceOfEntityData uid data :=
     match ets.find? uid.ty with
     |  .some entry =>
-      if instanceOfType data.attrs (.record entry.attrs) then
+      if instanceOfType data.attrs (.record entry.attrs) ets then
         if data.ancestors.all (λ ancestor => entry.ancestors.contains ancestor.ty) then
           if instanceOfEntityTags data entry then .ok ()
           else .error (.typeError s!"entity tags inconsistent with type store")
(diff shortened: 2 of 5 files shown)


cedar-lean/Cedar/Validation/Types.lean
```diff
@@ -61,21 +61,62 @@ abbrev QualifiedType := Qualified CedarType
...
 
-structure EntitySchemaEntry where
+structure StandardSchemaEntry where
   ancestors : Cedar.Data.Set EntityType
   attrs : RecordType
   tags : Option CedarType
 
+inductive EntitySchemaEntry where
+  | standard (ty: StandardSchemaEntry)
+  | enum (eids: List String)
```
emina, line 71: Let's change this to `enum (eids : Set String)`.  This is more accurate since the enum IDs form a set (e.g., no duplicates are allowed).

cedar-lean/Cedar/Validation/Types.lean
```diff
@@ -61,21 +61,62 @@ abbrev QualifiedType := Qualified CedarType
...
+
+def EntitySchemaEntry.validEid? (entry: EntitySchemaEntry) (eid: String): Bool :=
```
emina, line 73: Note that Lean uses `?` for functions that return `Option`.  

Let's rename this to `EntitySchemaEntry.isValidEntityUID` for consistency.

cedar-lean/Cedar/Validation/Types.lean
```diff
@@ -61,21 +61,62 @@ abbrev QualifiedType := Qualified CedarType
...
+  match entry with
+  | .standard _ => true
+  | .enum eids => eids.contains eid
+
+def EntitySchemaEntry.tags (entry: EntitySchemaEntry): Option CedarType :=
```
emina, line 78:
```suggestion
def EntitySchemaEntry.tags : EntitySchemaEntry → Option CedarType  
  | .standard ty => ty.tags
  | .enum _ => none
```

cedar-lean/Cedar/Validation/Types.lean
```diff
@@ -61,21 +61,62 @@ abbrev QualifiedType := Qualified CedarType
...
+  match entry with
+  | .standard ty => ty.tags
+  | .enum _ => none
+
+def EntitySchemaEntry.attrs (entry: EntitySchemaEntry): RecordType :=
```
emina, line 83: Same suggestion as for `EntitySchemaEntry.tags` for shorter code without an explicit `match`.

cedar-lean/Cedar/Validation/Types.lean
```diff
@@ -61,21 +61,62 @@ abbrev QualifiedType := Qualified CedarType
...
+  match entry with
+  | .standard ty => ty.attrs
+  | .enum _ => Map.empty
+
+def EntitySchemaEntry.ancestors (entry: EntitySchemaEntry): Set EntityType :=
```
emina, line 88: Same suggestion as for `EntitySchemaEntry.tags` for shorter code without an explicit `match`.

cedar-lean/Cedar/Validation/Types.lean
```diff
@@ -61,21 +61,62 @@ abbrev QualifiedType := Qualified CedarType
...
+  match entry with
+  | .standard ty => ty.ancestors
+  | .enum _ => Set.empty
+
 abbrev EntitySchema := Map EntityType EntitySchemaEntry
 
+def EntitySchema.getEIDRange (ets: EntitySchema) (et: EntityType) : Option (List String) :=
```
emina, line 95: What is this used for?  

Let's change the name to reflect the fact that the set of enums is a set (not a list / range).

cedar-lean/Cedar/Validation/Types.lean
```diff
@@ -61,21 +61,62 @@ abbrev QualifiedType := Qualified CedarType
...
+  match ets.find? et with
+  | .some (.enum eids) => .some eids
+  | .some (.standard _) | .none => .none
```
emina, line 98: The last case of the `match` can be simply `_ => .none`.

cedar-lean/Cedar/Validation/Types.lean
```diff
@@ -61,21 +61,62 @@ abbrev QualifiedType := Qualified CedarType
...
+  | .some (.enum eids) => eids.contains uid.eid
+  | .some (.standard _) => true
+  | .none   => false
+
 def EntitySchema.contains (ets : EntitySchema) (ety : EntityType) : Bool :=
   (ets.find? ety).isSome
 
 def EntitySchema.attrs? (ets : EntitySchema) (ety : EntityType) : Option RecordType :=
-  (ets.find? ety).map EntitySchemaEntry.attrs
+  (ets.find? ety).map λ e =>
```
emina, line 110: Why an explicit lambda here instead of `(ets.find? ety).map EntitySchemaEntry.attrs`? 

It looks like code should still work here.

cedar-lean/Cedar/Validation/Types.lean
```diff
@@ -61,21 +61,62 @@ abbrev QualifiedType := Qualified CedarType
...
+  match e with
+  | .standard ty => ty.attrs
+  | .enum _ => Map.empty
 
 def EntitySchema.tags? (ets : EntitySchema) (ety : EntityType) : Option (Option CedarType) :=
```
emina, line 115: Same comment as for `EntitySchema.attrs?`

cedar-lean/Cedar/Validation/Types.lean
emina: If the old code doesn't work, I'd suggest avoiding the lambda in favor of an explicit match:

```
match ets.find? ety with
  | .standard ty => ty.attrs
  | .enum _ => Map.empty
```

cedar-lean/CedarProto/ValidationRequest.lean
```diff
@@ -28,10 +28,21 @@ namespace Cedar.Validation.Proto
 
 deriving instance DecidableEq for ActionSchemaEntry
 deriving instance DecidableEq for ActionSchema
-deriving instance DecidableEq for EntitySchemaEntry
 deriving instance DecidableEq for EntitySchema
 deriving instance DecidableEq for Schema
 
+deriving instance Repr for StandardSchemaEntry
+
+instance : Repr EntitySchemaEntry where
```
emina, line 36: Do we need to give an implementation of `Repr` here?  We usually leave these as default Lean implementations.

emina (changes requested):
Nice PR!  I like this approach.  

Left small comments / suggestions on the model.

cedar-lean/CedarProto/ValidationRequest.lean
shaobo-he-aws: `CedarProto` requires `Schema` to implement `Repr`, which transitively does the same for `EntitySchemaEntry`.

cedar-lean/Cedar/Validation/Types.lean
shaobo-he-aws: It's used by `instanceOfEntityType` to determine if an `EntityUID` is valid.

cedar-lean/Cedar/Validation/Types.lean
```diff
@@ -68,34 +68,32 @@ structure StandardSchemaEntry where
...
-  match entry with
+def EntitySchemaEntry.ancestors : EntitySchemaEntry → Set EntityType
   | .standard ty => ty.ancestors
   | .enum _ => Set.empty
 
 abbrev EntitySchema := Map EntityType EntitySchemaEntry
 
-def EntitySchema.getEIDRange (ets: EntitySchema) (et: EntityType) : Option (List String) :=
+-- Get the possible range of EIDs specified by an enumerated entity type
+def EntitySchema.getEIDRange (ets: EntitySchema) (et: EntityType) : Option (Set String) :=
```
emina, line 93: Can we rename this to `EntitySchema.entityTypeMembers`?  

The members of an entity type don’t form a range in the sense that we don’t have an ordering on them that’s exposed in the Cedar semantics.

emina (changes requested):

cedar-lean/CedarProto/ValidationRequest.lean
emina: Why do we need a custom implementation of `Repr` though?  Note that we use the default Lean implementations for other `Repr` instances:  e.g., https://github.com/cedar-policy/cedar-spec/blob/2481def3f4ba8d5700364ccaa12d7746933d29c9/cedar-lean/Cedar/Spec/Value.lean#L146

shaobo-he-aws: `deriving` `Repr` works.

cedar-lean/Cedar/Validation/Types.lean
shaobo-he-aws: Done.

emina:
Thinking about this more, I think we need to change the definition of [`InstanceOfEntityType`](https://github.com/cedar-policy/cedar-spec/blob/2481def3f4ba8d5700364ccaa12d7746933d29c9/cedar-lean/Cedar/Thm/Validation/Typechecker/Types.lean#L41) and fix the upstream definitions and proofs accordingly.  

In particular, the definition should be this:

```
def InstanceOfEntityType (e : EntityUID) (ety: EntityType) (ets : EntitySchema) : Prop :=
  ety = e.ty && 
  sch.isValidEntityUID ety 
```

Without this change, `InstanceOfType` admits invalid enum literals. And this means that our [type soundness theorem](https://github.com/cedar-policy/cedar-spec/blob/2481def3f4ba8d5700364ccaa12d7746933d29c9/cedar-lean/Cedar/Thm/Validation/Typechecker.lean#L46) becomes too weak.  It says that the evaluator may produce invalid entity literals even for type-correct expressions.  The type checker code actually prevents this, but it also needs to be reflected in the definition of `InstanceOfType` and the soundness theorem.

Note that we have already made the corresponding change in `RequestEntityValidator` and `instanceOfEntityType`.

emina:
Actually, I now realize that the definition of `InstanceOfEntityType` has always been too weak (!). 

Note that it admits entity literals of arbitrary entity types that aren’t declared in the schema. 

The actual type checking algorithm doesn’t admit this. 

So, even before the addition of entity enums, this definition should have taken the environment  as input and looked like this instead:

```
def InstanceOfEntityType (e : EntityUID) (ety: EntityType) (env : Environment) : Prop :=
  env.ets.contains uid.ty  ∨ env.acts.contains uid
```

shaobo-he-aws:
> Actually, I now realize that the definition of `InstanceOfEntityType` has always been too weak (!).
> 
> Note that it admits entity literals of arbitrary entity types that aren’t declared in the schema.
> 
> The actual type checking algorithm doesn’t admit this.
> 
> So, even before the addition of entity enums, this definition should have taken the environment as input and looked like this instead:
> 
> ```
> def InstanceOfEntityType (e : EntityUID) (ety: EntityType) (env : Environment) : Prop :=
>   e.ty = ety && (env.ets.contains e.ty  ∨ env.acts.contains e)
> ```

To make the scope of this PR limited to RFC 53, I propose that we fix the definition first in a separate PR and rebase this PR based on it. I'll work on the fix first.

emina:
> In fact, `env.ets.isValidEntityUID e.ty` implies `env.ets.contains e.ty` so we might as well fix the definition in this PR.

I'll approve this PR as-is, in case you find it's easier to fix this in a separate PR.

emina (approved):

cedar-lean/Cedar/Thm/Validation/Typechecker/BinaryApp.lean
```diff
@@ -903,7 +903,10 @@ private theorem no_tags_type_implies_no_tags {uid : EntityUID} {env : Environmen
     replace ⟨e', h₂, h₃⟩ := h₂
     simp only [hf', Option.some.injEq] at h₂
     subst h₂
-    simp only [h₃] at h₁
+    simp only [EntitySchemaEntry.tags?] at h₁ h₃
+    split at h₃
+    · simp only [h₃] at h₁
+      simp only [h₁, map_empty_contains_instance_of_ff]
     simp only [h₁, map_empty_contains_instance_of_ff]
```
cdisselkoen, line 910: For readability, I think line 910 should also be indented with a `·`?

cedar-lean/Cedar/Validation/RequestEntityValidator.lean
```diff
@@ -158,7 +164,7 @@ def updateSchema (schema : Schema) (actionSchemaEntities : Entities) : Schema :=
       let entriesWithType := actionSchemaEntities.filter (λ k _ => k.ty == ty)
       let allAncestorsForType := List.flatten (entriesWithType.values.map (λ edt =>
         edt.ancestors.elts.map (·.ty) ))
-      let ese : EntitySchemaEntry := {
+      let ese : EntitySchemaEntry := .standard {
```
cdisselkoen, line 167: maybe for a future PR -- should action entities be modeled as enumerated entities instead of standard?

cedar-lean/Cedar/Validation/Typechecker.lean
```diff
@@ -359,6 +359,6 @@ def typeOf (x : Expr) (c : Capabilities) (env : Environment) : ResultType :=
 ---- Derivations -----
 
 deriving instance Repr for RequestType
-deriving instance Repr for Environment
+--deriving instance Repr for Environment
```
cdisselkoen, line 362: instead of commenting this out, should we just remove it?

cedar-lean/Cedar/Validation/Types.lean
```diff
@@ -61,21 +61,53 @@ abbrev QualifiedType := Qualified CedarType
...
+def EntitySchema.entityTypeMembers? (ets: EntitySchema) (et: EntityType) : Option (Set String) :=
+  match ets.find? et with
+  | .some (.enum eids) => .some eids
+  | _ => .none
+
+def EntitySchema.isValidEntityUID (ets : EntitySchema) (uid : EntityUID) : Bool :=
+  match ets.find? uid.ty with
+  | .some (.enum eids) => eids.contains uid.eid
+  | .some (.standard _) => true
+  | .none   => false
```
cdisselkoen, line 101: this duplicates logic from `EntitySchemaEntry.isValidEntityUID`.  Can we refactor this to call `EntitySchemaEntry.isValidEntityUID`?  That way if updates are needed in the future, they're needed in only one place

cedar-lean/Cedar/Validation/Types.lean
```diff
@@ -124,8 +156,8 @@ deriving instance Repr, DecidableEq, Inhabited for ExtType
 deriving instance Repr, DecidableEq, Inhabited for Qualified
 deriving instance Repr, Inhabited for CedarType
 deriving instance Repr for ActionSchemaEntry
-deriving instance Repr for EntitySchemaEntry
-deriving instance Repr for Schema
+-- deriving instance Repr for EntitySchemaEntry
+--deriving instance Repr for Schema
```
cdisselkoen, line 160: instead of commenting these out, should we just remove them?

cdisselkoen (commented):
Looks good, just some small nits

cedar-lean/Cedar/Validation/RequestEntityValidator.lean
shaobo-he-aws: Not yet. Enumerated entity types currently cannot form hierarchies, which actions can.

cdisselkoen: Got it.

cedar-lean/Cedar/Validation/Types.lean
shaobo-he-aws: Good catch. Done.

cdisselkoen (approved):

shaobo-he-aws:
Will merge this PR first and then address the weakness of `IsInstanceOfEntityType` later in a separate PR.

PR #533, opened 2025-02-07
Level checking soundness proof

john-h-kastner-aws:
Proves that level based entity slicing at level `n` cannot change the authorization decision when using polices that validate at level `n`.

The main theorem is

```lean
theorem validate_with_level_is_sound {ps : Policies} {schema : Schema} {n : Nat} {request : Request} {entities slice : Entities}
  (hr : validateRequest schema request = .ok ())
  (he : validateEntities schema entities = .ok ())
  (hs : slice = entities.sliceAtLevel request n)
  (htl : validateWithLevel ps schema n = .ok ()) :
  isAuthorized request entities ps = isAuthorized request slice ps
```


cedar-lean/Cedar/Spec/Slice.lean
```diff
@@ -0,0 +1,73 @@
...
+-/
+
+namespace Cedar.Spec
+
+open Cedar.Data
+
+def flatten_union {α} [LT α] [DecidableLT α] : List (Set α) → Set α :=
+  List.foldl (· ∪ ·) ∅
+
+def Value.sliceEUIDs (v : Value) : Set EntityUID :=
```
shaobo-he-aws, line 33: Maybe we can let it return a list of `EntityUID` or do the `List` to `Set` conversion after getting the list. I found termination proofs much easier this way (see: https://github.com/cedar-policy/cedar-spec/pull/526/files#diff-1670759f9561792dce989f71cca5ecdc85b5026ae00e2661b54e054158ffb9c0R74-R88)

cedar-lean/Cedar/Validation/Levels.lean
```diff
@@ -0,0 +1,119 @@
...
+  checked: Bool
+  -- Was the checked expression an entity root? Returns `false` for entity
+  -- literals, which are not roots used by the slicing algorithm. Used
+  -- internally by the level checking procedure to forbid access on entity
+  -- literals.
+  root: Bool
+
+def LevelCheckResult.and (r₀ r₁ : LevelCheckResult) : LevelCheckResult := LevelCheckResult.mk (r₀.checked && r₁.checked) (r₀.root && r₁.root)
+
+def checkLevel (tx : TypedExpr) (n : Nat) : LevelCheckResult :=
```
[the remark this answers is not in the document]
john-h-kastner-aws: Ultimately we want to diff test this same as the evaluator and typechecker, so using (only) an inductive won't work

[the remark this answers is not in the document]
emina: Yes, in the typechecker proof we use both:  inductive predicates for proofs, and then functions for request validation (with proofs that those functions enforce the constraints expressed in the inductive predicates).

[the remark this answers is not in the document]
john-h-kastner-aws: I think we do need some notion of "root", though the naming might not be the best.

If I have `User::"alice"`, that is well typed and should check at any level. But, as soon as I try to access an attribute it on like `User::"alice".name` we can't ever allow it. The slicing algorithm doesn't account for entity literals. It selects entities by starting from "roots" in the scope variables and any context entities, and the literal entity isn't one of those root.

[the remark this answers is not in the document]
john-h-kastner-aws: Thinking about this, it might be clearer to encode `root` with a separate recursive function (or inductive as the case may be) instead of shoehorning it into the same tree traversal. Should make the core level checking logic easier to follow

cedar-lean/Cedar/Validation/Levels.lean
```diff
@@ -0,0 +1,99 @@
...
+type-annotated AST instead of being built into the primary typechecking
+algorithm.
+-/
+
+namespace Cedar.Validation
+
+open Cedar.Data
+open Cedar.Spec
+
+def checkRoot : TypedExpr → Bool
```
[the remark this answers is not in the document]
emina: On second thought, I believe that this check is okay in the context of its sole caller, `checkLevel`, assuming that the input TypedExpr is well-formed.  

Still, this requires a lot of mental effort to understand and we want the model to be easy to grok :) 

So, I'd still favor refactoring this to make the intent more explicit.  (Or alternatively, add a detailed comment explaining why this does the right thing in the context of its caller and well-formedness of the input.)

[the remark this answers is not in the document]
john-h-kastner-aws: > I think I understand what this check is intended to do: ensure that the input expression contains no entity UID references.

Not quite. It ensures that it won't evaluate to value coming from a literal entity UID reference, so `if principal in Group::"Users" then foo else bar` isn't a problem, but it would be a problem if one if the arms was an entity literal

> Note that it rejects all literals, not just entity UIDs.

Yeah, I could make it only reject entity literals if that's clearer. As you noticed later, this doesn't actually matter given how it's used.

> I might also be tempted to use a more explicit though less efficient method of checking this.

This check would be too strict. See the `if` example. On a side note, I think we also don't need to look at set elements, so the current implementation could always return `true` for sets (but it doesn't atm).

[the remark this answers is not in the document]
emina: Yeah, this is very subtle. 

Is the record case necessary (by analogy to the set case)?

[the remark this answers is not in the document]
john-h-kastner-aws: We need something there definitely

We need to reject something like `{a: User::"Alice"}.a.b` which the current check accomplishes. It is a little conservative though. `{a: principal, b: User::"alice"}.a.c` is currently rejected when I don't think it needs to be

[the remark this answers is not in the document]
john-h-kastner-aws: > Note that it rejects all literals, not just entity UIDs.

Actually, this is an important point when combine with what I just said about records. An expression like `{a: principal, b: 0}.a.c` will be rejected by the current def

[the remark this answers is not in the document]
emina: Right, and we'd want to accept this.  (Admittedly, it seems like a really weird to write---maybe it could come from some auto-generated policy.)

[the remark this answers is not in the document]
john-h-kastner-aws: I had an idea which i think might solve this while avoiding a lot of complexity

```lean4
inductive EntityLit : TypedExpr → List Attr → Prop where
  | entity_lit {euid : EntityUID} {ty : CedarType} {path : List Attr}:
    EntityLit (.lit (.entityUID euid) ty) path
  | ite_true {x₁ x₂ x₃ : TypedExpr} {path : List Attr} {ty : CedarType}
    (hx₂ : EntityLit x₂ path) :
    EntityLit (.ite x₁ x₂ x₃ ty) path
  | ite_false {x₁ x₂ x₃ : TypedExpr} {path : List Attr} {ty : CedarType}
    (hx₃ : EntityLit x₃ path) :
    EntityLit (.ite x₁ x₂ x₃ ty) path
  | get_attr {x₁ : TypedExpr} {a : Attr} {path : List Attr} {ty : CedarType}
    (hx : EntityLit x₁ (a :: path)) :
    EntityLit (.getAttr x₁ a ty) path
  | get_tag {x₁ x₂ : TypedExpr} {path : List Attr} {ty : CedarType}
    (hx₁ : EntityLit x₁ path) :
    EntityLit (.binaryApp .getTag x₁ x₂ ty) path
  | record {attrs : List (Attr × TypedExpr)} {ty : CedarType} {a : Attr} {path : List Attr} {tx : TypedExpr}
    (hx₁ : (Map.make attrs).find? a = some tx)
    (hx₂ : EntityLit tx path) :
    EntityLit (.record attrs ty) (a :: path)

```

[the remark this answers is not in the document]
emina: Or maybe even define it as an inductive predicate on TypedExpr, like:

```
inductive TypedExpr.HasAccessPath : TypedExpr → List Attr → Prop where
```

cedar-lean/Cedar/Spec/Slice.lean
```diff
@@ -0,0 +1,67 @@
...
+  match v with
+  | .prim (.entityUID uid) => Set.singleton uid
+  -- TODO: You can't access these except by `in`, so maybe this could just be `Set.empty`
+  | .set s => flatten_union $ s.elts.attach.map λ e => e.val.sliceEUIDs
+  | .record r => flatten_union $ r.toList.attach.map λ e => e.val.snd.sliceEUIDs
```
emina, line 38: We can usually avoid having to provide termination proofs by using `List.attach₂` or `List.attach₃` for records.  (They carry the right proof obligations.)

Another thing that helps is explicitly deconstructing the set and record during the match, like so:

```
def Value. sliceEUIDs : Value → Set EntityUID
  | .prim (.entityUID uid) => Set.singleton uid
  | .record (Map.mk avs) => avs.attach₃.unionAll (λ ⟨(_, v), _⟩ => v. sliceEUIDs)
  |  _               => Set.empty
```

cedar-lean/Cedar/Validation/Levels.lean
```diff
@@ -0,0 +1,119 @@
...
+    match tx, path with
+    | .lit (.entityUID _) _, _ => false
+    | .ite _ x₂ x₃ _, _ =>
+      notEntityLit' x₂ path && notEntityLit' x₃ path
+    | .getAttr x₁ a _, _=>
+      notEntityLit' x₁ (a :: path)
+    | .binaryApp .getTag x₁ _ _, _ =>
+      notEntityLit' x₁ path
+    | .record axs ty, (a :: path) =>
+      match (Map.make axs).find? a with
```
emina, line 47: Proofs will be easier with just `axs.find?` here.  This lemma would be useful for termination: [List.mem_of_find?_eq_some](https://leanprover-community.github.io/mathlib4_docs/Init/Data/List/Find.html#List.mem_of_find?_eq_some).

john-h-kastner-aws: I think this has to `Map.make` since `axs` isn't sorted

emina: Ah, yes, that's right.  With `Map.make`, this theorem (and some the surrounding ones) will be helpful: [Map.find?_mem_toList](https://github.com/cedar-policy/cedar-spec/blob/91ac1f173ffb81236e27d4332ab0ff43af7121cd/cedar-lean/Cedar/Thm/Data/Map.lean#L296)

cedar-lean/Cedar/Spec/Slice.lean
john-h-kastner-aws: nice! That works for the `record` case, but I couldn't find the right incantation to get ride of the termination proof for sets

john-h-kastner-aws: actually, `attach₃` messes with some other proof  that need to step through that function. Is there an easy way unwrap it?

emina: Yes, you'll want to first `simp` it away, and then search for uses of this theorem: https://github.com/cedar-policy/cedar-spec/blob/91ac1f173ffb81236e27d4332ab0ff43af7121cd/cedar-lean/Cedar/Thm/Data/List/Lemmas.lean#L345 

Note that unification can be finicky with this one, so you'll want to provide explicit arguments (as we need to do in most other proofs that use this theorem).

That lets you strip away all the proof terms you're carrying around.

In this case though, I would suggest just getting rid of the Set case.  (I agree with your comment in the original code that the Set case can be skipped b/c there is no way to reference Set elements---just check for membership.)

cedar-lean/Cedar/Spec/Slice.lean
john-h-kastner-aws: termination proof here went away with Emina's advice on using `attach₃`

cedar-lean/Cedar/Spec/Slice.lean
john-h-kastner-aws: got it

emina: Yay!  And here is the PR with `List.unionAll` lemmas, now known as `List.unionMap`: https://github.com/cedar-policy/cedar-spec/pull/541

The lemmas are in `Cedar.Thm.Data.MapUnion` (see `mem_mapUnion_iff_mem_exists`).

cedar-lean/Cedar/Validation/Levels.lean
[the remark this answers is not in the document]
john-h-kastner-aws: yeah, the idea here is that the typed expression is an entity literal when dereferenced along a path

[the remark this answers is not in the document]
emina: Ah, got it.  In that case, maybe a better name would be `HasPathToEntityLit` or something like that.

cedar-lean/Cedar/Thm/Validation/Levels/Slice.lean
```diff
@@ -0,0 +1,633 @@
...
+    have ⟨ hc', rtxs, htx, hfat ⟩ := type_of_record_inversion ht
+    subst hc' htx
+
+    simp [evaluate] at he
+    cases he₁ : rxs.mapM₂ λ x => bindAttr x.1.fst (evaluate x.1.snd request entities) <;> simp [he₁] at he
+    subst he
+    cases ha
+    rename_i rvs v a path' hf' hv
+
+    have ⟨ x, hfx, hex ⟩ : ∃ x, (Map.make rxs).find? a = some x ∧ evaluate x request entities = .ok v := by
```
john-h-kastner-aws, line 489: @emina this bit is becoming anoying. Do you know if we've had to prove something like this somewhere else? I have a record expr which evaluates to value containing a pair `(a, v)` and want to show there was an `(a, x)` pair the original record where `x` evals to `v`.

emina: Here is the proof:

```
have ⟨ x, hfx, hex ⟩ : ∃ x, (Map.make rxs).find? a = some x ∧ evaluate x request entities = .ok v := by
      rw [List.mapM₂, List.attach₂] at he₁
      rw [List.mapM_pmap_subtype λ (x : Attr × Expr) => bindAttr x.fst (evaluate x.snd request entities)] at he₁
      rw [List.mapM_ok_iff_forall₂] at he₁
      simp only [Map.make, Map.find?, Map.kvs] at *
      split at hf' <;> simp at hf'
      subst hf'
      rename_i a v hf'
      have hf'' := List.find?_some hf'
      simp only [beq_iff_eq] at hf''
      subst hf''
      replace he₁ : List.Forallᵥ (λ x y => evaluate x request entities = Except.ok y) rxs rvs := by
        simp only [List.forallᵥ_def]
        apply List.Forall₂.imp _ he₁
        intro x y h
        simp only [bindAttr] at h
        cases hx : evaluate x.snd request entities <;> simp [hx] at h
        simp only [← h, and_self]
      replace he₁ := List.canonicalize_preserves_forallᵥ _ _ _ he₁
      simp only [List.forallᵥ_def] at he₁
      have ⟨(a', x), he₂, he₃, he₄⟩ := List.forall₂_implies_all_right he₁ (a, v) (List.mem_of_find?_eq_some hf')
      simp only at he₃ he₄
      subst he₃
      exists x
      simp only [List.mem_of_sortedBy_implies_find? he₂ (List.canonicalize_sortedBy _ _),
        he₄, and_self]
```

Can we break up the proof of the outer theorem into case-by-case lemmas?  This will help with proof maintenance and performance.  (The proof is sluggish on my ancient laptop.)

john-h-kastner-aws: Yeah, i was probably going to do that eventually. It's starting to take some time even on the more powerful dev desktop

cedar-lean/Cedar/Thm/Validation/Levels/CheckLevel.lean
```diff
@@ -0,0 +1,418 @@
...
+    (hl₁ : Level tx₁ n)
+    (hl₂ : Level tx₂ (n + 1)) :
+    Level (.binaryApp .getTag tx₁ tx₂ ty) (n + 1)
+  | hasTag (tx₁ tx₂ : TypedExpr) (ty : CedarType) (n : Nat)
+    (he : ¬ EntityLit tx₁ [])
+    (hl₁ : Level tx₁ n)
+    (hl₂ : Level tx₂ (n + 1)) :
+    Level (.binaryApp .hasTag tx₁ tx₂ ty) (n + 1)
+  | binaryApp (op : BinaryOp) (tx₁ tx₂ : TypedExpr) (ty : CedarType) (n : Nat)
+    (hop :
```
emina, line 188: I'd suggest factoring out the `match op with` check into a separate Prop.  It will make proof contexts nicer to work with (because you can unroll the definition of the new Prop as needed).

cedar-lean/Cedar/Thm/Validation/Levels/CheckLevel.lean
```diff
@@ -0,0 +1,418 @@
...
+            replace hf := List.sizeOf_lt_of_mem hf
+            rw [Prod.mk.sizeOf_spec a tx'] at hf
+            omega
+          have ih := not_entity_lit_spec' tx'
+          rw [ih]
+          intros h₃
+          apply h₁
+          constructor <;> assumption
+
+inductive Level : TypedExpr → Nat → Prop where
```
emina, line 151: I might call this something like `TypedExpr.AtLevel`.  We generally name Props to sound like property descriptions for a target type.  For example, `InstanceOfType` for validation.  Same for `EntityLit` above---I'd pick some more descriptive / proposition-ish name :).  Maybe `TypedExpr.AccessedViaPath` or some shorter version of that.

cedar-lean/Cedar/Thm/Validation/Levels/CheckLevel.lean
```diff
@@ -0,0 +1,418 @@
...
+    Level (.and tx₁ tx₂ ty) n
+  | or (tx₁ tx₂ : TypedExpr) (ty : CedarType) (n : Nat)
+    (hl₁ : Level tx₁ n)
+    (hl₂ : Level tx₂ n) :
+    Level (.or tx₁ tx₂ ty) n
+  | unaryApp (op : UnaryOp) (tx₁ : TypedExpr) (ty : CedarType)
+    (hl₁ : Level tx₁ n) :
+    Level (.unaryApp op tx₁ ty) n
+  | mem (tx₁ tx₂ : TypedExpr) (ty : CedarType) (n : Nat)
+    (he : ¬ EntityLit tx₁ [])
```
emina, line 173: I think I follow the logic of `he`:  we don't want `tx₁` to evaluate to an EntityLit via a path of any length.  This is where a more descriptive name for `EntityLit` would help re-inforce what this predicate is checking.

cedar-lean/Cedar/Thm/Validation/Levels/CheckLevel.lean
```diff
@@ -0,0 +1,418 @@
...
+    (hl₁ : Level tx₁ n)
+    (hty : tx₁.typeOf = .entity ety) :
+    Level (.getAttr tx₁ a ty) (n + 1)
+  | hasAttr (tx₁ : TypedExpr) (a : Attr) (ty : CedarType) {ety : EntityType} (n : Nat)
+    (he : ¬ EntityLit tx₁ [])
+    (hl₁ : Level tx₁ n)
+    (hty : tx₁.typeOf = .entity ety) :
+    Level (.hasAttr tx₁ a ty) (n + 1)
+  | getAttrRecord (tx₁ : TypedExpr) (a : Attr) (ty : CedarType) (n : Nat)
```
emina, line 205: I think it may make proofs simpler to state this definition positively (rather than by exclusion), like this:

```
| getAttrRecord (tx₁ : TypedExpr) (a : Attr) (rty : RecordType) (n : Nat)
    (hl₁ : Level tx₁ n)
    (hty : tx₁.typeOf = .record rty) :
    Level (.getAttr tx₁ a ty) n
```

Same for `hasAttrRecord`.

cedar-lean/Cedar/Thm/Validation/Levels/CheckLevel.lean
```diff
@@ -0,0 +1,418 @@
...
+    (hl₁ : Level tx₁ n)
```
emina, line 206: Doesn't the level depend on what attribute is being accessed?  

In other words, I was expecting this to have some sort of `EntityLit` precondition for the value of the specific attribute being accessed.  Because it's okay for record fields that aren't being dereferenced to be at whatever level; this is the case of `{ "a": principal, "b": User::"Alice" }.a.name` being allowed at level 1.

cedar-lean/Cedar/Thm/Validation/Levels/CheckLevel.lean
john-h-kastner-aws: Since the function is written

```lean
    match x₁.typeOf with
    | .entity _ => ...
    | _ => ...
```

Lean  give me ` ∀ (ety : EntityType), tx₁.typeOf = CedarType.entity ety → False` after a `split`, so at least for the equivlance proof, this is easier, and it didn't cause trouble in the other proof using this def.

cedar-lean/Cedar/Thm/Validation/Levels/CheckLevel.lean
john-h-kastner-aws: This applies to accessing a record attribute at any level, so that example will work

cedar-lean/Cedar/Thm/Validation/Levels/CheckLevel.lean
john-h-kastner-aws: done. that's a little bit nicer

cedar-lean/Cedar/Thm/Validation/Levels/CheckLevel.lean
john-h-kastner-aws: I've changed it to be `EntityLitViaPath` which I think expresses what it does

cedar-lean/Cedar/Thm/Validation/Levels/CheckLevel.lean
john-h-kastner-aws: Is there a benefit to naming this `TypedExpr.AtLevel` rather than just `AtLevel`? (I don't want to type out `TypedExpr` every time I reference it)

emina: Ah, yes, you can use method call notation to reference the predicate.  So, instead of writing `TypedExpr.AtLevel tx n`, you can write `tx.AtLevel n`.  Whether to use one or the other depends on which you find nicer to look at.

cedar-lean/Cedar/Thm/Validation/Levels/Or.lean
```diff
@@ -61,59 +61,10 @@ theorem level_based_slicing_is_sound_or {e₁ e₂ : Expr} {n : Nat} {c₀ c₁:
...
-          simp [evaluate]
-          simp [checkLevel] at hl
-          specialize ih₁ hs hc hr htx₁ (by simp [hl])
-          rw [←ih₁]
-          rcases he₁ with he₁ | he₁ | he₁ | he₁ <;>
-          simp [he₁, Result.as, Coe.coe, Value.asBool]
-          cases b₁ <;> simp
-          specialize ih₂ hs hc hr htx₂ (by simp [hl])
-          simp [ih₂]
+    replace ⟨ bt, tx₂, bty₂, c₂, htx, htx₂, hty₂, ht ⟩ := ht
```
emina, line 64: Nice!

cedar-lean/Cedar/Thm/Validation/Levels.lean
```diff
@@ -0,0 +1,62 @@
...
+import Cedar.Thm.Validation.Levels.GetAttr
+import Cedar.Thm.Validation.Levels.HasAttr
+
+namespace Cedar.Thm
+
+open Cedar.Data
+open Cedar.Spec
+open Cedar.Validation
+
+theorem level_based_slicing_is_sound {e : Expr} {tx : TypedExpr} {n : Nat} {c c₁ : Capabilities} {env : Environment} {request : Request} {slice entities : Entities}
```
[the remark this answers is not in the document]
john-h-kastner-aws: Added. The main theorem is now

```lean
theorem level_based_slicing_is_sound {p : Policy} {n : Nat} {env : Environment} {request : Request} {entities slice : Entities}
  (hs : slice = entities.sliceAtLevel request n)
  (hr : RequestAndEntitiesMatchEnvironment env request entities)
  (htl : typecheckAtLevel p env n = true) :
  evaluate p.toExpr request entities = evaluate p.toExpr request slice
```

where

```lean
def typecheckAtLevel (policy : Policy) (env : Environment) (n : Nat) : Bool :=
  match typeOf policy.toExpr ∅ env with
  | .ok (tx, _) => checkLevel tx n
  | _           => false

```

cedar-lean/Cedar/Thm/Validation/Levels.lean
```diff
@@ -14,4 +14,149 @@
...
+    replace htl₁ := List.forall₂_implies_all_left ∘ List.mapM_ok_iff_forall₂.mp $ htl₁
+    intro env he
+    specialize htl₁ env he
+    simp only [htl₁.imp, and_imp, imp_self, implies_true]
+  have ⟨env, ⟨he₁, hr⟩⟩ := he
+  specialize htl env he₁
+  replace ⟨_, htl⟩ := htl
+  exact typecheck_policy_with_level_is_sound hs hr htl
+
+theorem validate_with_level_is_sound {ps : Policies} {schema : Schema} {n : Nat} {request : Request} {entities slice : Entities}
```
emina, line 150: Nice!

cedar-lean/Cedar/Thm/Validation/Slice.lean
```diff
@@ -0,0 +1,215 @@
...
+    simp only [Entities.sliceAtLevel.sliceAtLevel, Option.some.injEq] at hs
+    cases hs₁ : List.mapM (Map.find? entities) work.elts <;>
+      simp only [hs₁, Option.map_eq_map, Option.bind_eq_bind, Option.bind_none_fun, Option.bind_some_fun, reduceCtorEq] at hs
+    rename_i eds
+    cases hs₂ : List.mapM (fun ed : EntityData => Entities.sliceAtLevel.sliceAtLevel entities ed.sliceEUIDs n) eds <;>
+      simp only [hs₂, Option.map_none', Option.map_some', Option.none_bind, Option.some_bind, reduceCtorEq, Option.some.injEq] at hs
+    subst slice
+    simp [Union.union, Set.union, Set.make_wf]
+
+theorem mapm_key_id_sorted_by_key {α β : Type} [LT α] [BEq α] [LawfulBEq α] {kvs : List (α × β)} {ks : List α} {fn : α → Option β}
```
emina, line 53: This is a generally useful lemma bout SortedBy.  I'd move it here: https://github.com/cedar-policy/cedar-spec/blob/7c2cc67f772068326c7ddbdf436aa4308c06e1cb/cedar-lean/Cedar/Thm/Data/List/Basic.lean#L180

And also rename to `mapm_key_id_sortedBy_key` to match the names of the other theorems about SortedBy.

cedar-lean/Cedar/Thm/Validation/Slice.lean
```diff
@@ -0,0 +1,215 @@
...
+      simpa using hlt
+
+    replace hs : List.SortedBy Prod.fst ((head₁, v₁) :: kvs') := by
+      have hm₄ : List.mapM (fun k => (fn k).bind fun v => some (k, v)) (head₁ :: tail) = some ((head₁, v₁) :: kvs') := by
+        simp [hm₂, hm₃]
+      exact mapm_key_id_sorted_by_key hm₄ hs
+
+    exact List.SortedBy.cons_cons hlt hs
+
+theorem map_find_mapm_value {α β : Type} [BEq α] [LawfulBEq α] {kvs : List (α × β)} {ks : List α} {fn : α → Option β} {k: α}
```
emina, line 88: Useful generic Map lemma.  Let's move it to here-ish: https://github.com/cedar-policy/cedar-spec/blob/7c2cc67f772068326c7ddbdf436aa4308c06e1cb/cedar-lean/Cedar/Thm/Data/Map.lean#L289

Also rename to  `map_find?_mapm_value` to match other theorems in the Map file.

cedar-lean/Cedar/Thm/Validation/Slice.lean
```diff
@@ -0,0 +1,215 @@
...
+  replace hmake : Map.mk eds = Map.make eds := by
+    have hsorted := eids.wf_iff_sorted.mp (slice_at_level_inner_well_formed hs₁)
+    have hsbfst := mapm_key_id_sorted_by_key hs₂ hsorted
+    have hwf : (Map.mk eds).WellFormed := by
+      simpa [Map.wf_iff_sorted, Map.toList, Map.kvs] using hsbfst
+    simpa [Map.WellFormed, Map.toList, Map.kvs] using hwf
+  have hmk := map_find_mapm_value hs₂ hi
+  simpa [hmake] using hmk
+
+theorem not_in_list_not_in_mapm_list {α γ : Type} [BEq α] [LT α] [DecidableLT α] {l : List α} {l' : List (α × γ)} {f : α → Option γ} {e: α}
```
emina, line 145: Move to https://github.com/cedar-policy/cedar-spec/blob/7c2cc67f772068326c7ddbdf436aa4308c06e1cb/cedar-lean/Cedar/Thm/Data/List/Lemmas.lean#L317

cedar-lean/Cedar/Thm/Validation/Slice.lean
```diff
@@ -0,0 +1,215 @@
...
+    cases hm₂ : (List.mapM (fun e => (f e).bind fun v => some (e, v)) tail) <;> simp only [hm₂, Option.none_bind, Option.some_bind, Option.some.injEq, reduceCtorEq] at hm
+    subst l'
+    cases hl'
+    case head =>
+      exact hl (List.Mem.head _)
+    case tail tail' ht' =>
+      replace hl : e ∉ tail := (hl $ List.Mem.tail _ ·)
+      exact not_in_list_not_in_mapm_list hm₂ hl _ ht'
+
+theorem mapm_none_find_none {α γ : Type} [DecidableEq α] [LT α] [StrictLT α] [DecidableLT α] {l : List α} {l' : List (α × γ)} {f : α → Option γ} {e: α}
```
emina, line 168: Move and rename as for `map_find_mapm_value`.

emina (approved):
Looks good to me!

chaluli (approved):
Looks good.

PR #534, opened 2025-02-10
fail directly when lean --print-libdir returns no output

cdisselkoen:
Test for the case where `lean --print-libdir` returns no output, and fail directly if so.  Currently, we instead set `LD_PRELOAD` to `/lib/glibc/libm.so` which is unlikely to exist, which causes more confusing errors.

### DIFF
--- a/cedar-drt/set_env_vars.sh
+++ b/cedar-drt/set_env_vars.sh
@@ -19,6 +19,10 @@ if ! command -v lean &> /dev/null; then
     return 1
 else
     export LEAN_LIB_DIR=$(lean --print-libdir)
+    if [ -z "$LEAN_LIB_DIR" ] then
+        echo "lean --print-libdir returned no output"
+        return 1
+    fi
     export LD_LIBRARY_PATH=${LD_LIBRARY_PATH+$LD_LIBRARY_PATH:}$(lean --print-libdir)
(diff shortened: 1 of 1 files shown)


cedar-drt/set_env_vars.sh
```diff
@@ -19,6 +19,10 @@ if ! command -v lean &> /dev/null; then
     return 1
 else
     export LEAN_LIB_DIR=$(lean --print-libdir)
+    if [ -z "$LEAN_LIB_DIR" ]; then
+        echo "lean --print-libdir returned no output"
```
adpaco-aws, line 23:
```suggestion
        echo "error: `lean --print-libdir` returned no output"
```

adpaco-aws (approved):

cedar-drt/set_env_vars.sh
cdisselkoen: Oh wait -- these backticks would be interpreted. that's not what we want

john-h-kastner-aws (approved):

cedar-drt/set_env_vars.sh
adpaco-aws: Oops, true. It's fine to remove them, the important part is to make clear that we ran into an error.

PR #547, opened 2025-02-21
`WellTyped` relation for `TypedExpr`

shaobo-he-aws:
*Issue #, if available:*

*Description of changes:*


cedar-lean/Cedar/Thm/Validation/TypedExpr.lean
```diff
@@ -0,0 +1,340 @@
...
+This file contains useful definitions and lemmas about well-typedness of `TypedExpr`
+-/
+
+namespace Cedar.Thm
+
+open Cedar.Validation
+open Cedar.Spec
+open Cedar.Data
+
+inductive Var.WellTyped : Var → CedarType → Prop
```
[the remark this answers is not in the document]
shaobo-he-aws: I'm curious if we can get away with `var`s just being either entity uids or records, rather than constrained by `env`. I'm fine with adding the constraints back because we will be likely to easily prove them.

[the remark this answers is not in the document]
emina: My guess is that it will be needed but we won't know for sure until we try to prove various things :) 

Since the Typechecker enforces these stronger constraints, and we have to carry the environment around in the upstream definitions anyway, it's probably safest to start with the stronger definition.  If we don't end up needing these, it would be easy to drop the stronger constraints.  If we do end up needing them, it might be harder to strengthen the definitions later.

cedar-lean/Cedar/Thm/Validation/WellTyped.lean
```diff
@@ -0,0 +1,380 @@
...
+namespace Cedar.Thm
+
+open Cedar.Validation
+open Cedar.Spec
+open Cedar.Data
+
+inductive Var.WellTyped (env : Environment) : Var → CedarType → Prop
+  | principal { ety : EntityType }
+    (h₁ : env.reqty.principal = ety) :
+    WellTyped env .principal (.entity ety)
```
emina, line 32: Why not omit the assumption and use `WellTyped env .principal (.entity env.reqty.principal)`?

cedar-lean/Cedar/Thm/Validation/WellTyped.lean
```diff
@@ -0,0 +1,380 @@
...
+  | toDate {x₁ : TypedExpr}
+    (h₁ : x₁.typeOf = .ext .datetime) :
+    WellTyped .toDate [x₁] (.ext .datetime)
+  | toTime {x₁ : TypedExpr}
+    (h₁ : x₁.typeOf = .ext .datetime) :
+    WellTyped .toTime [x₁] (.ext .duration)
+
+inductive TypedExpr.WellTyped (env : Environment) : TypedExpr → Prop
+| lit {p : Prim} {ty : CedarType}
+  (h₁ : InstanceOfType (.prim p) ty) :
```
emina, line 186: I think this is not quite strong enough in two ways: 
(1) `InstanceOfType` will accept boolean singleton types (tt/ff) but we want to allow only `.anyBool`
(2) We need to rule out entity literals that aren't declared in the environment (since these are ruled out by the typechecker).

I think the easiest way to enforce this is to write a separate `Prim.WellTyped` predicate that checks these conditions.

cedar-lean/Cedar/Thm/Validation/WellTyped.lean
```diff
@@ -0,0 +1,380 @@
...
+  WellTyped env (.getAttr x₁ attr ty)
+| set {ls : List TypedExpr} {ty : CedarType}
+  (h₁ : ∀ x, x ∈ ls → WellTyped env x)
+  (h₂ : ∀ x, x ∈ ls → x.typeOf = ty) :
+  WellTyped env (.set ls (.set ty))
+| record {rty : RecordType} {m : List (Attr × TypedExpr)}
+  (h₁ : ∀ k v, (k,v) ∈ m → WellTyped env v)
+  -- should we require well-formedness of `m` and then rewrite h₁ using quantifiers?
+  (h₂ : rty = Map.make (m.map (λ (a, ty) => (a, .required ty.typeOf)))) :
+  WellTyped env (.record m ty)
```
emina, line 246: The conclusion doesn't seem right.  I think it should be `WellTyped env (.record m (.record rty)))`.

cedar-lean/Cedar/Thm/Validation/WellTyped.lean
shaobo-he-aws: Good catch.

cedar-lean/Cedar/Thm/Validation/WellTyped.lean
shaobo-he-aws: Good catch.

cedar-lean/Cedar/Thm/Validation/WellTyped.lean
```diff
@@ -0,0 +1,261 @@
...
+    WellTyped env .hasTag x₁ x₂ (.bool .anyBool)
+  | getTag {x₁ x₂ : TypedExpr} {ety : EntityType}
+    (h₁ : x₁.typeOf = .entity ety)
+    (h₂ : x₂.typeOf = .string)
+    (h₃ : env.ets.tags? ety = .some (.some ty)) :
+    WellTyped env .getTag x₁ x₂ ty
+
+inductive ExtFun.WellTyped : ExtFun → List TypedExpr → CedarType → Prop
+  | decimal {x₁ : TypedExpr}
+    (h₁ : x₁.typeOf = .string) :
```
emina, line 135: This constraint can be stronger for all constructors (and would be useful for symbolic compilation proofs):

```
| decimal {s₁ : String} {d₁ : Decimal}
    (h₁ : d₁ = Decimal.decimal s) :
    WellTyped .decimal [.lit (.prim (.string s₁))] (.ext .decimal)
```

cedar-lean/Cedar/Thm/Validation/WellTyped.lean
```diff
@@ -0,0 +1,261 @@
...
+This file contains useful definitions and lemmas about well-typedness of `TypedExpr`
+-/
+
+namespace Cedar.Thm
+
+open Cedar.Validation
+open Cedar.Spec
+open Cedar.Data
+
+inductive Prim.WellTyped (env : Environment) : Prim → CedarType → Prop
```
emina, line 29: If we put these predicates into the namespaces in which their receivers are defined, then we can use the receiver syntax :) 

For example, if `Prim.WellTyped` is in the namespace `Cedar.Spec` (instead of `Cedar.Thm`), then we can just write `p.WellFormed env`. 

Same for UnaryOp etc.

They can all stay in the same file and in the same order.  It's fine to open / close multiple namespaces in one file.

cedar-lean/Cedar/Thm/Validation/WellTyped.lean
```diff
@@ -0,0 +1,261 @@
...
+  | bool (b : Bool) :
+    WellTyped env (.bool b) (.bool .anyBool)
```
john-h-kastner-aws, line 31: Can you explain the design here? The relation consistently doesn't handle `.tt` and `.ff`, so I assume this is deliberate, but I don't see why

cedar-lean/Cedar/Thm/Validation/WellTyped.lean
shaobo-he-aws: Got it. Updated this file accordingly.

cedar-lean/Cedar/Thm/Validation/WellTyped.lean
shaobo-he-aws: Yes, it's intentional. The well-typedness property only rules out type errors and ensures the resulting values to be certain types (i.e., specified by `InstanceofType`). This should be sufficient for our use cases, particularly TPE soundness proofs. If we allow Boolean subtypes like `.tt` or `.ff`, we will have to prove that the resulting values are true or false, which involves reasoning about capabilities and makes the definitions overly complicated.

emina: I can also confirm that the symbolic compiler ignores tt/ff types, so to prove that the compiler reduces validated policies to terms (rather than error), all we need to know is that the policy is well-typed as defined here.

john-h-kastner-aws (approved):

PR #548, opened 2025-02-21
TPE Lean model

shaobo-he-aws:
*Issue #, if available:*

*Description of changes:*


cedar-lean/Cedar/TPE/Evaluator.lean
```diff
@@ -0,0 +1,222 @@
...
+    | .val (.prim (.entityUID uid)) _ =>
+      match es.find? uid with
+      | .some ⟨ .some m, _ , _⟩  =>
+        .ok $ .val (m.contains a) ty
+      | .some ⟨ .none, _ , _⟩  => .ok $ .hasAttr r a ty
+      | .none => .ok $ .val false ty
+    | _ => .ok $ .getAttr r a ty
+  | .set xs ty => do
+    let rs ← xs.mapM₁ (λ ⟨x₁, _⟩ ↦ tpeExpr x₁ req es)
+    match rs.mapM Residual.asValue with
```
emina, line 175: Let's move this `match` to a helper function.

cedar-lean/Cedar/TPE/Evaluator.lean
```diff
@@ -0,0 +1,222 @@
...
+    | .some vs => .ok $ .val (.set (Set.mk vs)) ty
+    | .none => .ok $ .set rs ty
+  | .record m ty => do
+    let m₁ ← m.mapM₁ (λ ⟨(a, x₁), _⟩ ↦ do
```
emina, line 179: This is pretty hard to read.  Let's break it up into helper functions.

cedar-lean/Cedar/TPE/Evaluator.lean
```diff
@@ -0,0 +1,222 @@
...
+      let v ← tpeExpr x₁ req es
+      pure (a, v))
+    match m₁.mapM λ (a, r₁) ↦ do
+      let v₁ ← r₁.asValue
+      pure (a, v₁) with
+    | .some xs => .ok $ .val (.record (Map.mk xs)) ty
+    | .none => .ok $ .record m₁ ty
+  | .call f args ty => do
+    let rs ← args.mapM₁ (λ ⟨x₁, _⟩ ↦ tpeExpr x₁ req es)
+    match rs.mapM Residual.asValue with
```
emina, line 189: Let's move to a helper function :)

cedar-lean/Cedar/TPE/Evaluator.lean
```diff
@@ -0,0 +1,222 @@
...
+  coe := Error.evaluation
+
+def inₑ (uid₁ uid₂ : EntityUID) (es : PartialEntities) (ty : CedarType): Residual :=
+  if uid₁ = uid₂
+    then .val true ty
+    else
+    match es.find? uid₁ with
+    | .some ⟨_, .some uids, _⟩ => .val (uids.contains uid₂) ty
+    | .some ⟨_, .none, _⟩ => .binaryApp .mem (.val (.prim (.entityUID uid₁)) (.entity uid₁.ty)) (.val (.prim (.entityUID uid₂)) (.entity uid₂.ty)) ty
+    | .none => .val false ty
```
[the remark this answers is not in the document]
shaobo-he-aws: I was thinking if it prevents users to explicitly specify that uid₁ is not in partial entity store.

cedar-lean/Cedar/TPE/Residual.lean
```diff
@@ -0,0 +1,108 @@
...
+    let v ← e.evaluate req es
+    Cedar.Spec.hasAttr v a es
+  | .getAttr e a _ => do
+    let v ← e.evaluate req es
+    Cedar.Spec.hasAttr v a es
+  | .set xs _ => do
+    let vs ← xs.mapM₁ (fun ⟨x₁, _⟩ => evaluate x₁ req es)
+    .ok (Set.make vs)
+  | .record axs _ => do
+    let avs ← axs.mapM₂ (fun ⟨(a₁, x₁), _⟩ => bindAttr a₁ (evaluate x₁ req es))
```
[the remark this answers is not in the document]
shaobo-he-aws: I tried it and using `mapM₃` actually makes termination proof harder because we need to prove `sizeOf ty > 0`.

cedar-lean/Cedar/TPE/Evaluator.lean
```diff
@@ -0,0 +1,250 @@
...
+
+def inₑ (uid₁ uid₂ : EntityUID) (es : PartialEntities) : Option Bool :=
+  if uid₁ = uid₂
+  then
+    pure true
+  else
+    ((es.ancestors uid₁).map
+      λ ancestors ↦ ancestors.contains uid₂)
+
+def inₛ (uid₁ : EntityUID) (vs : Set Value) (es : PartialEntities): Result (Option Bool) :=
```
emina, line 83: This implementation is inconsistent with the rest of the code in terms of the assumptions that it makes.  The rest of the code assumes that all inputs are well-typed, and doesn't check for type errors. So we have two options:  we either assume that the input is well-typed here too, or we don't assume that the input is well-typed in other places. 

Going with the first assumption, we would re-write this implementation along the following lines:

```
def inₛ (uid₁ : EntityUID) (vs : Set Value) (es : PartialEntities): Option Bool := do 
  let uids  ← vs.mapM (Except.toOption ∘ Value.asEntityUID)
  uids.anyM (inₑ uid · es)
```

cedar-lean/Cedar/TPE/Evaluator.lean
```diff
@@ -0,0 +1,250 @@
...
+  (vs.mapOrErr Value.asEntityUID .typeError).map λ uids ↦
+    uids.foldl disjunction (.some false)
+where disjunction accum uid₂ := do
+  let l ← inₑ uid₁ uid₂ es
+  let r ← accum
+  pure (l || r)
+
+def hasTag (uid : EntityUID) (tag : String) (es : PartialEntities) : Option Bool :=
+  (es.tags uid).map λ tags ↦ tags.contains tag
```
emina, line 92: We can write this more compactly as

```
def hasTag (uid : EntityUID) (tag : String) (es : PartialEntities) : Option Bool :=
  (es.tags uid).map (Set.contains · tag)
```

cedar-lean/Cedar/TPE/Evaluator.lean
```diff
@@ -0,0 +1,250 @@
...
+
+def getTag (uid : EntityUID) (tag : String) (es : PartialEntities) (ty : CedarType) : Option Residual :=
```
emina, line 94: This is pretty hard to read.  In general, the `do` notation is easier to read than chained maps.  I suggest rewriting this more explicitly as follows:

```
def getTag (uid : EntityUID) (tag : String) (es : PartialEntities) (ty : CedarType) : Option Residual := do 
  let tags ← es.tags uid 
  match tags.find? tag with 
  | . some v => .val v ty 
  | .none => .error ty 
```

cedar-lean/Cedar/TPE/Evaluator.lean
```diff
@@ -0,0 +1,250 @@
...
+    .val (vs₂.subset vs₁) ty
+  | .containsAny, .val (.set vs₁) _, .val (.set vs₂) _ =>
+    .val (vs₁.intersects vs₂) ty
+  | .mem, .val (.prim (.entityUID uid₁)) _, .val (.prim (.entityUID uid₂)) _ =>
+    ((inₑ uid₁ uid₂ es).map Coe.coe).getD self
+  | .mem, .val (.prim (.entityUID uid₁)) _, .val (.set vs) _ =>
+    match inₛ uid₁ vs es with
+    | .ok (.some b) => b
+    | .ok .none => self
+    | .error _ => error
```
emina, line 124: No error case; see comment above on `inₛ`.

cedar-lean/Cedar/TPE/Evaluator.lean
```diff
@@ -0,0 +1,250 @@
...
+def apply₂ (op₂ : BinaryOp) (r₁ r₂ : Residual) (es : PartialEntities) (ty : CedarType) : Residual :=
+  match op₂, r₁, r₂ with
+  | .eq, .val v₁ _, .val v₂ _ =>
+    .val (v₁ == v₂) ty
+  | .less, .val (.prim (.int i)) _, .val (.prim (.int j)) _ =>
+    .val (i < j : Bool) ty
+  | .lessEq, .val (.prim (.int i)) _, .val (.prim (.int j)) _ =>
+    .val (i ≤ j : Bool) ty
+  | .add, .val (.prim (.int i)) _, .val (.prim (.int j)) _ =>
+    ((i.add? j).map (Value.toResidual · ty)).getD error
```
emina, line 107: Let's create a helper function and rewrite this (and other similar cases) as follows:

```
optionValueToResidual (i.add? j) ty 
```

Where `optionValueToResidual` is defined more explicitly as follows:

```
def optionValueToResidual (v : Option Value) (ty : CedarType) : Residual := 
  match v with 
  | .some v => .value v ty 
  | .none => .error ty 
```

cedar-lean/Cedar/TPE/Evaluator.lean
```diff
@@ -0,0 +1,250 @@
...
+  | .getTag, .val (.prim (.entityUID uid₁)) _, .val (.prim (.string tag)) _ =>
+    (getTag uid₁ tag es ty).getD self
+  | _, _, _ =>
+    self
+where
+  self := .binaryApp op₂ r₁ r₂ ty
+  error := .error ty
+
+def hasAttr (r : Residual) (a : Attr) (es : PartialEntities) (ty : CedarType) : Residual :=
+  match r with
```
emina, line 136: Nit: align match cases with match head.

cedar-lean/Cedar/TPE/Evaluator.lean
```diff
@@ -0,0 +1,250 @@
...
+      match es.attrs uid with
+      | .some m  => m.contains a
+      | .none => self
+    | _ => self
+where self := .hasAttr r a ty
+
+def getAttr (r : Residual) (a : Attr) (es : PartialEntities) (ty : CedarType) : Residual :=
+  match r with
+  | .val (.record xs) _ =>
+    ((xs.find? a).map (Value.toResidual · ty)).getD (.error ty)
```
emina, line 148: Let's use `optionValueToResidual` from above to write this as `optionValueToResidual (xs.find? a) ty`.

cedar-lean/Cedar/TPE/Evaluator.lean
emina: I don't like the name `optionValueToResidual` but that's the most obvious one that comes to mind :). Would love a shorter one.

cedar-lean/Cedar/TPE/Evaluator.lean
```diff
@@ -0,0 +1,250 @@
...
+  | .val (.prim (.entityUID uid)) _ =>
+    match es.attrs uid with
+    | .some m  =>
+      ((m.find? a).map (Value.toResidual · ty)).getD (.error ty)
```
emina, line 152: Let's use `optionValueToResidual` from above to write this as `optionValueToResidual (m.find? a) ty`.

cedar-lean/Cedar/TPE/Evaluator.lean
```diff
@@ -0,0 +1,250 @@
...
+    ((hasTag uid₁ tag es).map Coe.coe).getD self
+  | .getTag, .val (.prim (.entityUID uid₁)) _, .val (.prim (.string tag)) _ =>
+    (getTag uid₁ tag es ty).getD self
+  | _, _, _ =>
+    self
+where
+  self := .binaryApp op₂ r₁ r₂ ty
+  error := .error ty
+
+def hasAttr (r : Residual) (a : Attr) (es : PartialEntities) (ty : CedarType) : Residual :=
```
emina, line 135: I suggest rewriting this function and also `getAttr` to follow the style of the concrete evaluator, which uses the helper function `attrsOf` for both.  

Keeping the parallel structure really helps with proofs.

cedar-lean/Cedar/TPE/Evaluator.lean
```diff
@@ -0,0 +1,250 @@
...
+  match r with
+    | .val (.record m) _ => m.contains a
+    | .val (.prim (.entityUID uid)) _ =>
+      match es.attrs uid with
+      | .some m  => m.contains a
+      | .none => self
+    | _ => self
+where self := .hasAttr r a ty
+
+def getAttr (r : Residual) (a : Attr) (es : PartialEntities) (ty : CedarType) : Residual :=
```
emina, line 145: I suggest rewriting this function to follow the style of the concrete evaluator, which uses the helper function `attrsOf`.

cedar-lean/Cedar/TPE/Evaluator.lean
```diff
@@ -0,0 +1,250 @@
...
+    | .none => self
+  | _ => self
+where self := .getAttr r a ty
+
+def set (rs : List Residual) (ty : CedarType) : Residual :=
+  match rs.mapM Residual.asValue with
+  | .some vs => .val (.set (Set.make vs)) ty
+  | .none => .set rs ty
+
+def bindAttr [Monad m] (a : Attr) (res : m α) : m (Attr × α) := do
```
emina, line 162: We should consider replacing `bindAttr` in Cedar.Spec.Evaluator with this function, which can then be used for both concrete and partial evaluation.

cedar-lean/Cedar/TPE/Evaluator.lean
```diff
@@ -0,0 +1,250 @@
...
+  pure (a, v)
+
+def record (m : List (Attr × Residual)) (ty : CedarType) : Residual :=
+  match m.mapM λ (a, r₁) ↦ bindAttr a r₁.asValue with
+  | .some xs => .val (.record (Map.make xs)) ty
+  | .none => .record m ty
+
+def call (f : ExtFun) (rs : List Residual) (ty : CedarType) : Residual :=
+  match rs.mapM Residual.asValue with
+    | .some vs => match (Spec.call f vs).map (Value.toResidual · ty) with
```
emina, line 173: Let's put the `match` on a separate line.

cedar-lean/Cedar/TPE/Evaluator.lean
```diff
@@ -0,0 +1,250 @@
...
+      | .ok v => v
+      | .error _ => .error ty
+    | .none => .call f rs ty
+
+def evaluate (x : TypedExpr)
+    (req : PartialRequest)
+    (es : PartialEntities)
+    : Residual :=
+  match x with
```
emina, line 182: Let's use the same names for matched variables in these cases that we use in the concrete evaluator. 

For example, `.lit l`, `.ite x₁ x₂ x₃` etc

cedar-lean/Cedar/TPE/Evaluator.lean
```diff
@@ -0,0 +1,250 @@
...
+    record m ty
+  | .call f args ty =>
+    let rs := args.map₁ (λ ⟨x₁, _⟩ ↦ evaluate x₁ req es)
+    call f rs ty
+termination_by x
+decreasing_by
+  all_goals
+    simp_wf
+    try omega
+  case _ h =>
```
emina, line 225: We should be able merge these cases as follows:

```
all_goals 
  rename_i h
  replace h := List.sizeOf_lt_of_mem h
  try simp at h
  omega 
```

cedar-lean/Cedar/TPE/Residual.lean
```diff
@@ -0,0 +1,116 @@
...
+  | .call xfn xs _ => do
+    let vs ← xs.mapM₁ (fun ⟨x₁, _⟩ => evaluate x₁ req es)
+    Cedar.Spec.call xfn vs
+  | .error _ => .error .extensionError
+termination_by x
+decreasing_by
+  all_goals
+    simp_wf
+    try omega
+  case _ h =>
```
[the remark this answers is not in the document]
shaobo-he-aws: It seems that these goals are anonymous.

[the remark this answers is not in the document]
emina: Ah, yes, in that case, we can do this right after `all_goals`:

```
any_goals
    have := List.sizeOf_lt_of_mem h
    omega
```

cedar-lean/Cedar/TPE/Evaluator.lean
shaobo-he-aws: This is much cleaner than mine.

cedar-lean/Cedar/TPE/Evaluator.lean
```diff
@@ -26,211 +26,190 @@ open Cedar.Validation
...
 def hasTag (uid : EntityUID) (tag : String) (es : PartialEntities) : Option Bool :=
-  (es.tags uid).map λ tags ↦ tags.contains tag
+  (es.tags uid).map (Map.contains · tag)
 
-def getTag (uid : EntityUID) (tag : String) (es : PartialEntities) (ty : CedarType) : Option Residual :=
-  (es.tags uid).map (λ tags ↦
-    ((tags.find? tag).map (.val · ty)).getD (.error ty))
+def getTag (uid : EntityUID) (tag : String) (es : PartialEntities) (ty : CedarType) : Option Residual := do
+  let tags ← es.tags uid
+  match tags.find? tag with
```
emina, line 95: Monadic calls can be inlined like this (so you don't have to introduce intermediate variables like `tags` that are used just to get the result):

```
match (← es.tags uid).find? tag with
```

I think the code is good as-is, but this is a useful pattern to know about :)

cedar-lean/Cedar/TPE/Evaluator.lean
```diff
@@ -26,211 +26,190 @@ open Cedar.Validation
...
-    match inₛ uid₁ vs es with
-    | .ok (.some b) => b
-    | .ok .none => self
-    | .error _ => error
+    someOrSelf (inₛ uid₁ vs es) ty self
   | .hasTag, .val (.prim (.entityUID uid₁)) _, .val (.prim (.string tag)) _ =>
-    ((hasTag uid₁ tag es).map Coe.coe).getD self
+    someOrSelf (hasTag uid₁ tag es) ty self
   | .getTag, .val (.prim (.entityUID uid₁)) _, .val (.prim (.string tag)) _ =>
     (getTag uid₁ tag es ty).getD self
```
emina, line 134: Any reason why this is not `someOrSelf (getTag uid₁ tag es ty) ty self` instead of `getD`?

shaobo-he-aws: It's because `getTag` used to return a `Optional Residual`, which can be `.some (.error ty)` when `tag` is not found in concrete `tags`. I changed its return type to `Residual`.

emina (approved):
Looks good to me!

cdisselkoen (approved):
Looks great

PR #551, opened 2025-02-26
generate protodatas for unit tests using cedar-policy interface

cdisselkoen:
This PR refactors the generator of protodatas for Lean unit tests, so that instead of using `cedar-policy-core` and `cedar-policy-validator`, it uses only functionality exposed by `cedar-policy` itself.

A few tests had to be tweaked as a result.  There's no way to encode _just_ a `TemplateBody` or _just_ a `Type` from the public interface, so tests that used to encode a `TemplateBody` now encode a singleton policyset (which still tests that `TemplateBody` encoding), and tests that used to encode a `Type` now encode a minimal `Schema` which uses that type.  I consider this a worthwhile trade for using only public-interface functions.


cedar-lean/UnitTest/CedarProto-test-data/record.protodata
```diff
@@ -1,10 +1,10 @@
...
 
 ham
 
-
\ No newline at end of file
+
+
+eggs
+
+
```
cdisselkoen, line 10: Regenerating the protodatas results in some random changes because of the use of Rust HashMap, which results in some protodatas having elements encoded in different orders (but equivalent).  This test, which shouldn't be changed by this PR, nonetheless seems to have had its two record elements (`eggs` and `ham`) encoded in the other order in this run, vs what was previously checked in.

john-h-kastner-aws (approved):

cedar-lean/UnitTest/CedarProto-test-data/record.protodata
john-h-kastner-aws: Do you think this is worth fixing? I'd definitely like to have a deterministic serialization, but requiring a sort before serializing isn't ideal if protobuf is supposed to be fast.

cdisselkoen: The generated protobufs are entirely equivalent (when deserialized by spec-conforming parsers) so most people will never notice that the serialization isn't technically deterministic.  I'm not sure if deterministic serialization is a goal worth trying to achieve?  What would be the motivation?  just avoiding this noise when regenerating some unit tests?

cdisselkoen: I don't think our JSON serializations are deterministic either?

cedar-lean/UnitTest/CedarProto.lean
```diff
@@ -50,20 +50,6 @@ namespace UnitTest.CedarProto
...
-    let buf ← IO.FS.readBinFile filename
-    let parsed : Except String α := Proto.Message.interpret? buf
-    match parsed with
-    | .ok req => checkEq req expected
-    | .error e => pure (.error e)
-  ⟩
-
 /--
   Similar to `testDeserializeProtodata`, but `f` is applied to the deserialized
   value before comparing to `expected`
```
adpaco-aws, line 55: Needs to be updated since `testDeserializeProtodata` is now based on this function.

adpaco-aws (approved):

PR #556, opened 2025-02-27
rely on impls more

cdisselkoen:
- Adds a generic instance for `Field (Option a)` to our generic `Protobuf` library, and uses it where applicable in `CedarProto`
- Adds `HAppend` impls for `Repeated` and `Packed` types, and use that instead of `Field.concatenate` which only worked for Arrays (not Repeated or Packed)

### DIFF
--- a/cedar-lean/Protobuf/Field.lean
+++ b/cedar-lean/Protobuf/Field.lean
@@ -37,11 +37,6 @@ namespace Merge
 @[inline]
 def override (_ : α) (x : α) : α := x
 
-/-- Concatenation semantics, combines two arrays -/
-@[inline]
-def concatenate (x1 : Array α) (x2 : Array α) : Array α :=
-  x1.append x2
-
 end Merge
 
 @[inline]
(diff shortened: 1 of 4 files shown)


cedar-lean/Protobuf/Field.lean
```diff
@@ -58,6 +53,15 @@ def guardWireType {α : Type} [Field α] (wt : WireType) : BParsec Unit := do
   if foundWt ≠ wt then
     throw s!"WireType mismatch: found {repr foundWt}, expected {repr wt}"
 
+instance [Field α] : Field (Option α) where
+  parse := do
```
shaobo-he-aws, line 57: Nit: Is it equivalent to `Field.parse.map some`?

cedar-lean/Protobuf/Field.lean
```diff
@@ -58,6 +53,15 @@ def guardWireType {α : Type} [Field α] (wt : WireType) : BParsec Unit := do
...
+    let a : α ← Field.parse
+    pure (some a)
+  merge
+  | some a1, some a2 => some (Field.merge a1 a2)
```
shaobo-he-aws, line 61: Nit: a1 -> a₁

shaobo-he-aws (approved):
LGTM

cedar-lean/Protobuf/Field.lean
cdisselkoen: Yes it is, thanks!

PR #557, opened 2025-02-27
refactor out some panics in CedarProto

cdisselkoen:
Refactors out some panics in CedarProto.  Specifically, the Protobuf parser now gracefully fails (rather than panicking) when encountering:
- unknown extension type (in a schema)
- `ip()` or `decimal()` argument that fails to parse
- expression of an unsupported variant (ie not a restricted expression) where a Value is expected

### DIFF
--- a/cedar-lean/Cedar/Validation/Types.lean
+++ b/cedar-lean/Cedar/Validation/Types.lean
@@ -61,6 +61,12 @@ def map {α β} (f : α → β) : Qualified α → Qualified β
   | optional a => optional (f a)
   | required a => required (f a)
 
+def transpose {α ε} : Qualified (Except ε α) → Except ε (Qualified α)
+  | optional (.ok a) => .ok (optional a)
+  | required (.ok a) => .ok (required a)
+  | optional (.error e) => .error e
+  | required (.error e) => .error e
+
 end Qualified
 
 inductive CedarType where
--- a/cedar-lean/CedarProto/Schema.lean
+++ b/cedar-lean/CedarProto/Schema.lean
@@ -49,32 +49,42 @@ private def descendantsToAncestors [LT α] [DecidableEq α] [DecidableLT α] (de
 
 namespace Schema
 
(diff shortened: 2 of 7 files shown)


cedar-lean/Cedar/Validation/Types.lean
```diff
@@ -61,6 +61,12 @@ def map {α β} (f : α → β) : Qualified α → Qualified β
   | optional a => optional (f a)
   | required a => required (f a)
 
+def transpose {α ε} : Qualified (Except ε α) → Except ε (Qualified α)
```
shaobo-he-aws, line 64: Nit: I'm uncertain if we need this function here. Intuitively,

```Lean
let attrs ← attrs.toList.mapM λ (k,v) => do
    let v ← v.map ProtoType.toCedarType |>.transpose
    .ok (k, v)
```
can be written like
```Lean
let attrs ← attrs.toList.mapM λ (k,q) => do
    let t ← ProtoType.toCedarType q.getType
    .ok (k, q.map ∘ Function.const t)
```

Or we can write this function as `q.getType.map fun t1 => q.map ∘ Function.const t1`

shaobo-he-aws (approved):
LGTM

cedar-lean/Cedar/Validation/Types.lean
shaobo-he-aws: Syntactically correct version:
```Lean
def transpose {α ε} (q : Qualified (Except ε α)) : Except ε (Qualified α) :=
  q.getType.map ((q.map ∘ Function.const _) ·)
```

cdisselkoen: I acknowledge you're correct, but I find that harder to read; anyone else have thoughts?

adpaco-aws: The current definition seems more readable to me. I don't know what `q.getType` or `Function.const` mean so I guess some familiarity with those methods is needed.

adpaco-aws (approved):

PR #558, opened 2025-02-27
Lean: Update models with the `datetime` extension

adpaco-aws:
*Description of changes:* Update the Lean models with the `datetime` extension, including:
 - Changes to the evaluator and validator to work with `datetime` types and methods.
 - Fixes existing theorems to work with `datetime` types and methods.
 - Adds new theorems for `datetime` types and methods (two for each constructor/method).


cedar-lean/Cedar/Spec/Ext.lean
```diff
@@ -26,16 +27,25 @@ open Cedar.Spec.Ext
...
 abbrev IPAddr := IPAddr.IPNet
+abbrev Duration := Datetime.Duration
 
 inductive Ext where
   | decimal (d : Decimal)
   | ipaddr (ip : IPAddr)
+  | datetime (dt : Datetime)
+  | duration (dur: Duration)
 
+-- Ordering on extension types: .decimal < .ipaddr < .datetime < .duration
```
shaobo-he-aws, line 38: Nit: Can we rewrite the comment as a doc comment?

cedar-lean/Cedar/Spec/Ext.lean
```diff
@@ -26,16 +27,25 @@ open Cedar.Spec.Ext
...
 def Ext.lt : Ext → Ext → Bool
   | .decimal d₁, .decimal d₂ => d₁ < d₂
   | .ipaddr ip₁, .ipaddr ip₂ => ip₁ < ip₂
-  | .decimal  _, .ipaddr _   => true
-  | .ipaddr   _, .decimal _  => false
+  | .datetime d₁, .datetime d₂ => d₁ < d₂
+  | .duration d₁, .duration d₂ => d₁ < d₂
+  | .decimal  _, _
```
shaobo-he-aws, line 44: Nit: IIUC, this looks cleaner and should also work,

```Lean
| .decimal _ , _ | .ipaddr _, _ | .datetime _, _  => true
| .duration _ => false
```

cedar-lean/Cedar/Spec/Ext/Datetime.lean
```diff
@@ -40,7 +38,30 @@ namespace Datetime
...
+  val : Int64
+
+def Datetime.lt (d₁ d₂ : Datetime) : Bool :=
+  d₁.val < d₂.val
+
+def Datetime.le (d₁ d₂ : Datetime) : Bool :=
+  d₁.val ≤ d₂.val
+
+instance : LT Datetime where
+  lt := fun d₁ d₂ => Datetime.lt d₁ d₂
```
shaobo-he-aws, line 51: Nit: it could be `(Datetime.lt · ·)`

cedar-lean/Cedar/Spec/Ext/Datetime.lean
```diff
@@ -94,18 +137,22 @@ def MILLISECONDS_PER_DAY: Int := 86400000
...
-abbrev duration? := Int64.ofInt?
+def duration? (i: Int) : Option Duration :=
+  Int64.ofInt? i |>.map Duration.mk
+
+def Duration.neg? (d : Duration) : Option Duration :=
+  d.val.neg? |>.map Duration.mk
 
-def durationUnits? (n: Nat) (suffix: String) : Option Duration :=
+def durationUnits? (n: Nat) (suffix: String) : Option Int :=
   match Int64.ofInt? n with
```
shaobo-he-aws, line 147: Nit: We can use `do` notation here if this function is not used in proofs. Like,

```Lean
do
  let i <- Int64.ofInt? n
  match suffix with ...
```

cedar-lean/Cedar/Thm/Validation/Typechecker/BinaryApp/Cmp.lean
```diff
@@ -32,23 +32,40 @@ theorem type_of_int_cmp_inversion {op₂ : BinaryOp} {x₁ x₂ : Expr} {c c' :
...
-    rename_i tc₁ tc₂ _ _ _ _ h₅ h₆
+    simp only [typeOfBinaryApp, ok, List.empty_eq, err] at h₂
+    split at h₂ <;> try {contradiction}
+    all_goals (
+      simp only [Except.ok.injEq, Prod.mk.injEq, List.nil_eq] at h₂
+      simp only [← h₂, TypedExpr.typeOf, true_and]
+      rename_i tc₁ tc₂ _ _ _ _ h₅ h₆
+    )
+  )
+  case' h_6 | h_9 =>
```
shaobo-he-aws, line 56: Nit: make them like `h₅ h₆`?

shaobo-he-aws (approved):
LGTM

cedar-lean/Cedar/Spec/Ext.lean
adpaco-aws: That definition wouldn't work because the asymmetric property (`a < b → ¬b < a`) cannot be proven with it. For example, `.decimal _, .ipaddr _` and `.ipaddr _, .decimal _` are both `true` in there.

cedar-lean/Cedar/Spec/Ext/Datetime.lean
adpaco-aws: I didn't know that, thanks! However, I think I'll keep like this since it's more beginner-friendly and we use the same style for IpAddr defintions.

cedar-lean/Cedar/Spec/Ext/Datetime.lean
adpaco-aws: Thanks, added!

cedar-lean/Cedar/Thm/Validation/Typechecker/BinaryApp/Cmp.lean
adpaco-aws: If I'm not mistaken, these names correspond to the cases themselves, not the hypotheses. I've added comments above each to make clear the cases that each arm handles, but please let me know if you've got other suggestions.

cedar-lean/Cedar/Spec/Expr.lean
```diff
@@ -142,20 +142,4 @@ end
 
 instance : DecidableEq Expr := decExpr
 
-def Value.asExpr : Value → Expr
-  | .prim p => .lit p
-  | .set vs => .set (vs.elts.map₁ λ ⟨v, _⟩ => v.asExpr)
-  | .record attrs => .record (attrs.kvs.attach₂.map λ ⟨(k, v), _⟩ => (k, v.asExpr))
-  | .ext (.decimal d) => .call ExtFun.decimal [.lit (.string (toString d))]
-  | .ext (.ipaddr ip) => .call ExtFun.ip [.lit (.string (toString ip))]
```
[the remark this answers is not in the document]
adpaco-aws: Yes, @emina noted it was unused and asked me to delete it.

adpaco-aws:
Thanks for the reviews! Ci was failing with errors like this:
```
error: ././././UnitTest/Datetime.lean:170:59: failed to synthesize
  Neg Duration
Additional diagnostic information may be available using the `set_option diagnostics true` command.
error: ././././UnitTest/Datetime.lean:171:62: failed to synthesize
  OfNat Duration 60000
numerals are polymorphic in Lean, but the numeral `60000` cannot be used in a context where the expected type is
  Duration
due to the absence of the instance above
Additional diagnostic information may be available using the `set_option diagnostics true` command.
```
when running the unit tests, so I added the instances which are infallible and I don't think they should be used for anything else other than the testing code (noted this in a comment).

I'm also planning on ignoring the DCO check because rebasing here could be complicated.

shaobo-he-aws (approved):

cedar-lean/Cedar/Spec/Ext/Datetime.lean
```diff
@@ -94,19 +154,21 @@ def MILLISECONDS_PER_DAY: Int := 86400000
...
-    | "ms" => duration? i
-    | "s" => duration? (i * MILLISECONDS_PER_SECOND)
-    | "m" => duration? (i * MILLISECONDS_PER_MINUTE)
-    | "h" => duration? (i * MILLISECONDS_PER_HOUR)
-    | "d" => duration? (i * MILLISECONDS_PER_DAY)
-    | _ => none
+def Duration.neg? (d : Duration) : Option Duration :=
+  d.val.neg?
+
+def durationUnits? (n: Nat) (suffix: String) : Option Int := do
```
emina, line 163: Can you help me understand why this change here is correct?  `durationUnits?` is now not checking that the result is a duration (and therefore in 64-bit range).  Is that checked elsewhere?

cedar-lean/Cedar/Thm/Validation/Typechecker/Types.lean
```diff
@@ -229,6 +231,30 @@ theorem instance_of_ipAddr_type_is_ipAddr {v₁ : Value} :
...
+  exists d
+
+theorem instance_of_duration_type_is_duration {v₁ : Value} :
+  InstanceOfType v₁ (CedarType.ext ExtType.duration) →
+  ∃ d, v₁ = .ext (.duration d)
+:= by
+  intro h₁
+  cases h₁
+  rename_i x h₁
+  simp [InstanceOfExtType] at h₁
```
emina, line 253: If a `simp` occurs in the middle of a proof, we want to replace with `simp only` for robustness in the face of updates. 

To figure out what needs to go into a `simp only` call, replace a `simp` with `simp?` in the IDE, and the Lean will give you a suggestion for how to replace that `simp` with `simp only`.

emina:
> when running the unit tests, so I added the instances which are infallible and I don't think they should be used for anything else other than the testing code (noted this in a comment).

Can you point me to where in the code these are?

cedar-lean/Cedar/Spec/Ext/Datetime.lean
adpaco-aws: `durationUnits?` is used by `parseUnit?` which in turn is used by `parseUnsignedDuration?`. This last function is where we create a checked duration in the statement
```lean
then duration? (days + hours + minutes + seconds + milliseconds)
```

Regardless, since we're using `do` notation here, the `let i ← Int64.ofInt? n` should result in `none` when the result wouldn't fit in a 64-bit integer.

cedar-lean/Cedar/Spec/Ext/Datetime.lean
```diff
@@ -85,7 +110,42 @@ def parse (str: String) : Option Datetime := do
...
+-- Note: The instances below are only used for tests.
+-- We might redefine them if they are need for something else.
+instance : OfNat Duration n where
+  ofNat := ⟨Int64.ofNat n⟩
+
+instance : ToString Duration where
+  toString d := toString d.val
+
+instance : Neg Duration where
+  neg d := ⟨-d.val⟩
```
adpaco-aws, line 148: These are the new instances required for `Duration` unit tests.

adpaco-aws (commented):
> Can you point me to where in the code these are?

cedar-lean/Cedar/Spec/Ext/Datetime.lean
emina: Just to be safe, can we move these instance declarations to the UnitTesting module that uses them?  This way, they will be available only in that module (and its importers).  

The `ToString` one seems benign, but `OfNat` and `Neg` may be tricky.

cedar-lean/Cedar/Spec/Ext/Datetime.lean
emina: The `do` notation won't help here though because the multiplications on lines 167-170 could overflow.

So, the output of this function could be an integer that doesn't fit into 64 bits.  This overflow would be propagated by `parseUnit?` too, because it also returns an Int.

We need to change this back to Int64, I think, because you want this overflow error detected as soon as it happens.  As far as I can tell, with this change, the overflow won't be detected until line 207, which will blame the overflow on the addition.

cedar-lean/Cedar/Spec/Ext/Datetime.lean
adpaco-aws: > but OfNat and Neg may be tricky.

Yeah, that's why I wanted to call it out. Putting them in the testing module makes more sense.

cedar-lean/Cedar/Spec/Ext/Datetime.lean
adpaco-aws: It's true that the overflow error would be delayed in this case. Changed it back.

cedar-lean/Cedar/Thm/Validation/Typechecker/Types.lean
adpaco-aws: I've added `only` to most of the `simp` occurring anywhere other than the "tail" of a proof.

There's a couple of `simp`s that cannot be converted to `simp only` without inducing errors. It's the only `simp` statement in the `type_of_call_offset_is_sound` and `type_of_call_durationSince_is_sound` theorems. I couldn't fix them - I suppose that's because `simp?` is not working well in those cases?

emina: You can leave them for now.  Usually, the tricky case for converting to `simp only` is when you're replacing a `simp` that is applied to multiple goals simultaneously.  In that case, you have to put the union of the suggestions into `simp only` by manually copying-pasting, which can get annoying.

emina (approved):
LGTM!

PR #560, opened 2025-03-01
Enable `datetime` generation

shaobo-he-aws:
This reverts commit 3681ffc8975e92aa645ee811c5eedb85ce9b5996.

*Issue #, if available:*

*Description of changes:*


cdisselkoen (approved):
Alternately, we could leave the flag and just set it to `true` in all targets.  No strong opinion from me.

Issue #543, opened 2025-02-17
Move remaining targets to using protobuf for Lean/Rust FFI


john-h-kastner-aws:
### Category

DRT target(s), DRT infrastructure

### Describe the feature you'd like to request

In #488 we moved authorization and validation targets to using protobuf instead of JSON. All other differential tests should also use protobuf. This should increase throughput allowing for more test, and allow us to remove these internal JSON formats from `cedar-policy-core`

### Describe alternatives you've considered

.

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

john-h-kastner-aws:
#562 did most of these, but we still need to update at least  validation-drt-type-directed and level-validation-drt

cdisselkoen:
I think #562 did accomplish this.  What validation-drt-type-directed and level-validation-drt are missing is #559, which missed these targets (well, missed validation-drt-type-directed, while level-validation-drt didn't exist yet).

PR #562, opened 2025-03-03
use Protobuf for additional DRT targets

cdisselkoen:
*Issue #, if available:*

*Description of changes:*

Uses Protobuf, rather than the undocumented internal JSON format, for the `eval-type-directed`, `entity-validation`, and `request-validation` DRT targets.  Also renames Protobuf-related structs in the DRT package to remove `Msg` suffix which is redundant.

Tested that these three targets successfully run for a least a few seconds on my machine.


john-h-kastner-aws:
linking #543

shaobo-he-aws (approved):
LGTM

shaobo-he-aws:
Does it mean that we can retire the JSON deserializer?

john-h-kastner-aws (approved):
looks good

PR #565, opened 2025-03-03
Add types and methods from `datetime` extension to parsers

adpaco-aws:
*Description of changes:* Adds the types and methods from the `datetime` extension to parsers. 

*Created as a draft because I haven't been able to test the changes locally yet*

### DIFF
--- a/cedar-lean/DiffTest/Parser.lean
+++ b/cedar-lean/DiffTest/Parser.lean
@@ -133,12 +133,6 @@ def jsonToExtFun (json : Lean.Json) : ParseResult ExtFun := do
   | "isLoopback" => .ok .isLoopback
   | "isMulticast" => .ok .isMulticast
   | "isInRange" => .ok .isInRange
-  | "datetime" => .ok .datetime
-  | "duration" => .ok .duration
-  | "offset" => .ok .offset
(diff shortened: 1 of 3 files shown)


cedar-lean/DiffTest/Parser.lean
```diff
@@ -133,6 +133,12 @@ def jsonToExtFun (json : Lean.Json) : ParseResult ExtFun := do
   | "isLoopback" => .ok .isLoopback
   | "isMulticast" => .ok .isMulticast
   | "isInRange" => .ok .isInRange
+  | "datetime" => .ok .datetime
```
cdisselkoen, line 136: The changes in this file conflict with https://github.com/cedar-policy/cedar-spec/pull/567, which deletes this file.  I guess it's a race to see which PR we merge first :) Either way, the correct end state is that this file is deleted

cdisselkoen (approved):

cedar-lean/DiffTest/Parser.lean
adpaco-aws: True, I've removed the changes in that file to avoid conflicts.

john-h-kastner-aws (approved):

PR #569, opened 2025-03-06
Extend typechecker inversion lemma for type annotation

john-h-kastner-aws:
Some changes pulled out of #533


cedar-lean/Cedar/Thm/Validation/Typechecker/BinaryApp.lean
```diff
@@ -35,6 +35,28 @@ open Cedar.Data
 open Cedar.Spec
 open Cedar.Validation
 
+theorem type_of_binaryApp_inversion {op₂ : BinaryOp} {x₁ x₂ : Expr} {c c' : Capabilities} {env : Environment} {tx : TypedExpr}
+  (htx : typeOf (Expr.binaryApp op₂ x₁ x₂) c env = Except.ok (tx, c')) :
+  ∃ tx₁ c₁ tx₂ c₂,
+    typeOf x₁ c env = Except.ok (tx₁, c₁) ∧
+    typeOf x₂ c env = Except.ok (tx₂, c₂) ∧
+    ∃ ty, tx = .binaryApp op₂ tx₁ tx₂ ty
```
shaobo-he-aws, line 43: Is it ok to combine the two existential quantifiers? Btw, there's no detailed specification of `ty`, like other inversion theorems. Is it intended?

john-h-kastner-aws: > Is it ok to combine the two existential quantifiers?

I don't see why not

> Btw, there's no detailed specification of ty

I don't actually care about `ty` for my proofs, so this is good enough. The inversion lemma for the specific binary operators do specify `ty`, but I wanted one inversion lemma for all binary operators to make my proofs simpler. We could extend this lemma to specify the types of `tx` `tx\1` and `tx\2` in a `match` on `op\2`. Then other inversion lemma could be done in terms of this one, but I didn't need to go that far.


> I realized that these inversion theorems essentially say that typedOf produces well-typed expressions

Yeah, there's probably some overlap here. I expect that TPE will need something stronger than what I needed for levels, but we'll probably be able to end up with both proof using the same inversion theorems.

shaobo-he-aws (approved):

shaobo-he-aws (approved):

emina (approved):

chaluli (approved):

PR #571, opened 2025-03-06
Add model for level checking and slicing

john-h-kastner-aws:
*Issue #, if available:*

Splits out the model implementations from #533 for review and merging separately

Note that level checking has diverged significantly what was included in the first draft PR. These changes should make level checking as implemented here more precise than in Rust implementation, which fell short of the behavior specified in the RFC. 

*Description of changes:*


emina (approved):
Looks great to me!

shaobo-he-aws (approved):
LGTM

shaobo-he-aws (approved):

PR #572, opened 2025-03-07
Restore `Serialize` for two fuzz targets

adpaco-aws:
*Description of changes:* Our fuzzers are built which `--features log` which in turn require the `Serialize` trait to be implemented for fuzz target inputs. https://github.com/cedar-policy/cedar/pull/1520 removed the `Serialize` implementations from many core/validator types, but CI didn't catch it because we don't test the build with `--features log`.

This PR restores `Serialize` for two fuzz targets:
 - `protobuf-roundtrip`: We use `#[serde(skip)]` for each field except `policy`. This is how it's done in other targets, and I assume `ValidatorSchema` is safe to skip because it's built from `Schema` which is skipped in other targets as well.
 - `wildcard-matching`: We implement a simple custom serializer to avoid restoring it in `cedar`.

### DIFF
--- a/cedar-drt/fuzz/fuzz_targets/protobuf-roundtrip.rs
+++ b/cedar-drt/fuzz/fuzz_targets/protobuf-roundtrip.rs
@@ -33,11 +33,16 @@ use cedar_policy_generators::{
     settings::ABACSettings,
 };
 
-#[derive(Debug, Clone)]
+use serde::Serialize;
+
+#[derive(Debug, Clone, Serialize)]
 struct FuzzTargetInput {
+    #[serde(skip)]
     request: ABACRequest,
     policy: ABACPolicy,
+    #[serde(skip)]
     entities: Entities,
+    #[serde(skip)]
     schema: cedar_policy_validator::ValidatorSchema,
 }
 
--- a/cedar-drt/fuzz/fuzz_targets/wildcard-matching.rs
+++ b/cedar-drt/fuzz/fuzz_targets/wildcard-matching.rs
@@ -20,6 +20,8 @@ use cedar_drt_inner::fuzz_target;
 use cedar_policy_core::ast::{Pattern, PatternElem};
 use libfuzzer_sys::arbitrary::{self, Arbitrary, Result, Unstructured};
 use regex::{escape, Regex};
+use serde::{Serialize, Serializer};
+use serde::ser::SerializeStruct;
 
 /// Input expected by this fuzz target:
 /// A pattern and a string that matches it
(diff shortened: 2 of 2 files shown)


john-h-kastner-aws (approved):
Looks fine. I still don't understand what we gain by building with `log`, so I wonder if we could get rid of it and then not need this serialize impl. It's not much overhead though, so I'm happy with this fix.

cedar-drt/fuzz/fuzz_targets/protobuf-roundtrip.rs
```diff
@@ -33,11 +33,16 @@ use cedar_policy_generators::{
...
+#[derive(Debug, Clone, Serialize)]
 struct FuzzTargetInput {
+    #[serde(skip)]
     request: ABACRequest,
     policy: ABACPolicy,
+    #[serde(skip)]
     entities: Entities,
+    #[serde(skip)]
     schema: cedar_policy_validator::ValidatorSchema,
 }
```
cdisselkoen, line 47: I'm still interested in what this `Serialize` is used for, and whether we could just remove uses of `Serialize` for this `FuzzTargetInput` instead of restoring `Serialize` here.

If we do go this route, I'm wondering whether this `Serialize` implementation which serializes only the `policy` is going to be helpful, or just cause even more confusion for whatever is consuming it.  For instance, if it's supposed to be a representation of the logged `FuzzTargetInput`, the fact that you only see the `policy` is probably going to be confusing or even deceiving while debugging a problem with a `FuzzTargetInput`.

katherine-hough (approved):

cedar-drt/fuzz/fuzz_targets/protobuf-roundtrip.rs
katherine-hough: I assume it's used [here](https://github.com/cedar-policy/cedar-spec/blob/c5e8ad2458dfbbc41d8af4615deb0d07090d8342/cedar-drt/fuzz/src/prt.rs#L147).

cdisselkoen: I think we can remove that code? Is it used by anything?

katherine-hough: Nothing if you don't use `--features log`.

cdisselkoen: I guess the question is whether `--features log` is useful if it only logs the `policy` and not the other inputs.

adpaco-aws: I doubt the logs are useful given that they ignore most of the input. I decided to open the PR because it was taking me too long to figure where `LOGFILE` is being set, and how it's connected to other components (i.e., are these files being expected somewhere else?).

katherine-hough: Probably not. I assume this is used to analyze a random sample of generated inputs after the campaign completes presumably to collect statistics about what was generated or debug generation. Since policy generation depends on the other generated inputs, I imagine just having the policies wouldn't help very much with either. Long term I think there is probably way to re-implement this that doesn't depend on the generated structures implementing Serialize.

cdisselkoen: I concur.

adpaco-aws: Update: I've found where it's used, and looks like at least a couple targets might be using it. Don't think anyone is looking at the logged information though.

adpaco-aws:
Closed in favor of #575

PR #573, opened 2025-03-10
Update level checking to allow access to literals equal to environment action

john-h-kastner-aws:
The change here is adding

```lean
  | action (ty : CedarType) (env : Environment) (n nmax : Nat) (path : List Attr) :
    EntityAccessAtLevel  (.lit (.entityUID env.reqty.action) ty) env n nmax path
```

but there are lot of other changes due to now needing `env` where previously this property did not.

I've also pushed these changes to #533 with accompanying proof updates, but I'm putting them up separately for easier review


cedar-lean/Cedar/Thm/Validation/Levels/CheckLevel.lean
```diff
@@ -33,101 +33,103 @@ def DereferencingBinaryOp : BinaryOp → Prop
...
-  | getTag (tx₁ tx₂ : TypedExpr) (ty : CedarType) (n : Nat)
-    (hl₁ : EntityAccessAtLevel tx₁ n (n + 1) [])
-    (hl₂ : AtLevel tx₂ (n + 1)) :
-    AtLevel (.binaryApp .getTag tx₁ tx₂ ty) (n + 1)
-  | hasTag (tx₁ tx₂ : TypedExpr) (ty : CedarType) (n : Nat)
-    (hl₁ : EntityAccessAtLevel tx₁ n (n + 1) [])
-    (hl₂ : AtLevel tx₂ (n + 1)) :
-    AtLevel (.binaryApp .hasTag tx₁ tx₂ ty) (n + 1)
-  | binaryApp (op : BinaryOp) (tx₁ tx₂ : TypedExpr) (ty : CedarType) (n : Nat)
+inductive TypedExpr.AtLevel : TypedExpr → Environment → Nat → Prop where
```
shaobo-he-aws, line 64: If `env` is a constant, you can make it as an explicit argument like,
```Lean
inductive TypedExpr.AtLevel (env : Environment) : TypedExpr → Nat → Prop
```

shaobo-he-aws: We have to pass `env` to every "call site" though.

shaobo-he-aws: I tried to make it an implicit argument somewhere else but had no luck.

john-h-kastner-aws: I started doing this, but I actually don't like that it means `TypedExpr` isn't the first argument at call sites. I think the properties read nicer when the expr I'm applying it to is the first thing I see, and it matches better with our `typeOf` function that takes an `Expr` argument first.

cdisselkoen: fun fact: as long as the function is called `TypedExpr.AtLevel`, you can use it as `my_typed_expr.AtLevel env` even if the `TypedExpr` is not the first argument.  Lean figures it out and makes the first `TypedExpr` argument the receiver argument

cdisselkoen: At least that works for functions, I assume also inductives

john-h-kastner-aws: I tried that, but couldn't get the method style call to work for the inductive even before moving the args around

emina: Ah, for this to work, `TypedExpr.atLevel` has to be in the `Cedar.Validation` namespace (where `TypedExpr` is originally declared).

john-h-kastner-aws: that's annoying. Definitely feels like this should live in the `Cedar.Thm` namespace IMO, but we could move it just to make this nicer. Any strong opinions?

emina: Ah, we can keep it in this file but put it in the `Cedar.Validation` namespace.  I agree that we should keep all the theorem-related stuff out of the model and in Cedar/Thm.

emina: It's fine to have multiple open/closed namespaces in the same file:
```
namespace Cedar.Validation
-- stuff
end Cedar.Validation

namespace Cedar.Thm
-- stuff
end Cedar.Thm
```

shaobo-he-aws (approved):

chaluli (approved):

PR #575, opened 2025-03-10
Remove the `log` feature

adpaco-aws:
*Description of changes:* Remove the `log` feature, and the `Serialize` implementations it required for all fuzz targets.

Context: We recently discovered that the `log` feature was being used in DRT, but it's mostly undocumented and we estimate it's not so useful since we don't log most of the input (see [this conversation](https://github.com/cedar-policy/cedar-spec/pull/572#discussion_r1985735186)). We've stopped using `--features log` in DRT so we can get rid of the feature in addition to the `Serialize` implementations we didn't want to restore in the Cedar API.


cdisselkoen (approved):

cedar-drt/fuzz/fuzz_targets/eval-type-directed.rs
```diff
@@ -15,7 +15,6 @@
  */
 
 #![no_main]
-use cedar_drt::utils::expr_to_est;
```
cdisselkoen, line 18: might be able to remove this utility function `expr_to_est` entirely

adpaco-aws: Good catch, it's removed now.

katherine-hough (approved):

john-h-kastner-aws (approved):

Issue #404, opened 2024-07-31
Extend Policy JSON roundtrip targets to PolicySet JSON


john-h-kastner-aws:
### Category

DRT target(s)

### Describe the feature you'd like to request

These targets only round trip the single policy format, but the EST JSON format now supports the multiple policies in a policy set, so we should test round tripping that as well.  

### Describe alternatives you've considered

.

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

PR #576, opened 2025-03-10
Updated parsing roundtrip targets to operate on policy sets

katherine-hough:
Note: this change requires https://github.com/cedar-policy/cedar/pull/1525 to be accepted in order to build.

*Issue #, if available:*
Resolves Issue #404 

*Description of changes:*
* Updated `convert_policy_cedar_to_json` and `convert_policy_json_to_cedar` fuzzing targets to convert policy sets instead of individual policies
* Changed corpus initialization script to copy over seeds from a seeds directory
* Added a seed file with a linked template for `convert-policy-json-to-cedar` fuzzing target
* Added `policy-set-roundtrip` fuzzing target to test similar roundtrip path to `roundtrip` but for policy sets instead policies
* Extracted function `arbitrary_abac_constraints` from `Schema::arbitrary_policy()`
* Extracted function `arbitrary_abac_constraints_size_hint` from `Schema::arbitrary_policy_size_hint()`
* Added generator for creating sets of ABAC templates and static policies
* Changed `GeneratedPolicy::arbitrary_for_hierarchy()` to optionally use a schema to generate actions

### DIFF
--- a/cedar-drt/fuzz/src/parsing_utils.rs
+++ b/cedar-drt/fuzz/src/parsing_utils.rs
@@ -65,6 +65,7 @@ pub fn check_for_internal_errors(errs: ParseErrors) {
 
 // Exposes only the aspects of a template that are preserved through conversions
 // to different representations.
+// In particular, policy ids and source locations are not exposed.
 #[derive(Eq, Debug)]
 struct PreservedTemplate<'a> {
     template: &'a Template,
@@ -107,6 +108,7 @@ impl<'a> From<&'a Template> for PreservedTemplate<'a> {
 
 // Checks that the static policies and templates in the specified
 // sets are equivalent, ignoring policy ids and source locations.
+// Ignores links in the policy sets when determining equivalence.
 // Panics if the two sets are not equivalent.
 pub fn check_policy_set_equivalence(set1: &PolicySet, set2: &PolicySet) {
     // Convert templates in each set to preserved templates
@@ -130,12 +132,13 @@ pub fn policy_set_to_text(policy_set: &cedar_policy_core::ast::PolicySet) -> Str
         to_cedar_text(template, &mut res).unwrap();
         for template in iter {
             to_cedar_text(template, &mut res).unwrap();
+            writeln!(&mut res).unwrap();
         }
     }
     res
 }
 
-fn to_cedar_text(template: &Template, f: &mut String) -> std::fmt::Result {
+fn to_cedar_text(template: &Template, f: &mut impl std::fmt::Write) -> std::fmt::Result {
     for (k, v) in template.annotations() {
         writeln!(f, "@{}(\"{}\")", k, v.val.escape_debug())?
     }
--- a/cedar-drt/initialize_corpus.sh
+++ b/cedar-drt/initialize_corpus.sh
@@ -45,9 +45,7 @@ initialize_corpus() {
   seed_dir="./fuzz/seeds/$1"
   if test -d $seed_dir; then
     echo "Copying seeds for fuzz target $1"
-    for f in "$seed_dir/"*; do
-      cp "$f" "$corpus_dir/$(uuidgen)-$(basename "$f")"
-    done
+    cp -r "$seed_dir"/. "$corpus_dir"
   fi
 
   case "$1" in


cedar-drt/fuzz/fuzz_targets/convert-policy-json-to-cedar.rs
```diff
@@ -15,40 +15,43 @@
...
 }
 
 fuzz_target!(|est_json_str: String| {
-    if let Ok(ast_from_est) = serde_json::from_str::<serde_json::Value>(&est_json_str)
-        .and_then(|val| serde_json::from_value::<cedar_policy_core::est::Policy>(val))
+    // JSON -> EST -> AST
+    // Attempt to deserialize an est policy set from the specified json string
+    // Then, convert the est policy set to an ast policy set
+    let ast_from_est_result = serde_json::from_str::<serde_json::Value>(&est_json_str)
+        .and_then(|val| serde_json::from_value::<cedar_policy_core::est::PolicySet>(val))
```
john-h-kastner-aws, line 36: IRRC, most of the seeds for this target are policy JSON rather than policy set JSON, they'll fail to parser here. Would it make sense to include a path on this target that tries parsing JSON->Single-Policy and then packaging that into a singleton policy set before continuing?

cedar-drt/fuzz/src/parsing_utils.rs
```diff
@@ -61,3 +62,90 @@ pub fn check_for_internal_errors(errs: ParseErrors) {
...
+    if let Some(template) = iter.next() {
+        to_cedar_text(template, &mut res).unwrap();
+        for template in iter {
+            to_cedar_text(template, &mut res).unwrap();
+        }
+    }
+    res
+}
+
+fn to_cedar_text(template: &Template, f: &mut String) -> std::fmt::Result {
```
john-h-kastner-aws, line 138: Could have this take a `f: &mut impl std::fmt::Write`  instead of a `&mut String`

cedar-drt/fuzz/src/parsing_utils.rs
```diff
@@ -61,3 +62,90 @@ pub fn check_for_internal_errors(errs: ParseErrors) {
...
+        .collect::<HashSet<PreservedTemplate>>();
+    // Check that the sets are the same size
+    assert_eq!(t1, t2);
+}
+
+/// Converts an ast policy set to Cedar text, removing linked policies.
+pub fn policy_set_to_text(policy_set: &cedar_policy_core::ast::PolicySet) -> String {
+    let mut res = String::new();
+    let mut iter = policy_set.all_templates();
+    if let Some(template) = iter.next() {
```
john-h-kastner-aws, line 129: Could you get rid of this outer `if`?

cedar-drt/fuzz/src/parsing_utils.rs
```diff
@@ -61,3 +62,90 @@ pub fn check_for_internal_errors(errs: ParseErrors) {
...
+    fn from(value: &'a Template) -> Self {
+        PreservedTemplate {
+            template: value,
+            annotations: value.annotations().map(|(k, v)| (k, &v.val)).collect(),
+        }
+    }
+}
+
+// Checks that the static policies and templates in the specified
+// sets are equivalent, ignoring policy ids and source locations.
```
john-h-kastner-aws, line 109: Also ignores links?

cedar-drt/fuzz/src/parsing_utils.rs
```diff
@@ -61,3 +62,90 @@ pub fn check_for_internal_errors(errs: ParseErrors) {
         miette::Report::new(errs).wrap_err("unexpected internal error")
     )
 }
+
+// Exposes only the aspects of a template that are preserved through conversions
+// to different representations.
```
john-h-kastner-aws, line 67: Can expand on this to mention what isn't preserved?

cedar-drt/initialize_corpus.sh
```diff
@@ -40,6 +40,16 @@ fi
 initialize_corpus() {
   corpus_dir="./fuzz/corpus/$1/"
   mkdir -p "$corpus_dir"
+
+  # Copy over seeds
+  seed_dir="./fuzz/seeds/$1"
+  if test -d $seed_dir; then
+    echo "Copying seeds for fuzz target $1"
+    for f in "$seed_dir/"*; do
+      cp "$f" "$corpus_dir/$(uuidgen)-$(basename "$f")"
```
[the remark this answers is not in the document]
katherine-hough: `uuidgen` is not needed to prevent collisions. I thought it would look more consistent this way, but you are right it probably makes more sense to just do a straight copy. I thought about storing the seeds in `fuzz/corpus` directly but ultimately decided that this made it more clear what the purpose of files is.

cedar-drt/fuzz/src/parsing_utils.rs
katherine-hough: What if `all_templates` is empty?

cedar-drt/fuzz/src/parsing_utils.rs
katherine-hough: Aren't links considered neither static policies nor templates?

john-h-kastner-aws: Sure, but for a function called `check_policy_set_equivalence` I expect those to be checked.

cedar-drt/fuzz/src/parsing_utils.rs
john-h-kastner-aws: loop zero times?

cedar-drt/initialize_corpus.sh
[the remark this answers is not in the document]
john-h-kastner-aws: Sounds good. I prefer the straight copy. No preference on whether we copy of commit `fuzz/corpus`

cedar-drt/fuzz/fuzz_targets/convert-policy-json-to-cedar.rs
katherine-hough: Five seeds are policy JSON. Three are policy sets, including the seed I added. Generally, I think it is better if an evolutionary fuzzer accepts a single input format when possible. Using multiple formats tends to negatively impact the effectiveness of crossover operators and one input format may be able to saturate shared feature probes before the other which can limit exploration of the other format.

cedar-drt/fuzz/src/parsing_utils.rs
katherine-hough: Oh, I was missing the newline between policies. That's why I was it separated out like this. It should be:
```  if let Some(template) = iter.next() {
        to_cedar_text(template, &mut res).unwrap();
        for template in iter {
            to_cedar_text(template, &mut res).unwrap();
            writeln!(&mut res).unwrap();
        }
    }
```

john-h-kastner-aws: ah makes sense

john-h-kastner-aws (approved):

PR #577, opened 2025-03-11
Fixes issue in Datetime.Duration.parse when parsing min duration

chaluli:
There is a bug in the current Duration.parse which prevented the successful parsing of durations that represented -9223372036854775808ms


adpaco-aws (approved):
Great catch, thanks!

cdisselkoen (approved):

PR #578, opened 2025-03-11
Lean: Definitions and theorems for `duration` converters

adpaco-aws:
*Description of changes:* #527 missed the converter "methods" associated to the `duration` type, as specified in [the RFC for the datetime extension](https://cedar-policy.github.io/rfcs/0080-datetime-extension.html):
 - `.toMilliseconds()` returns a `long` describing the number of milliseconds in this duration (the value as a long, itself)
 - `.toSeconds()` returns a `long` describing the number of seconds in this duration (`.toMilliseconds() / 1000`)
 - `.toMinutes()` returns a `long` describing the number of minutes in this duration (`.toSeconds() / 60`)
 - `.toHours()` returns a `long` describing the number of hours in this duration (`.toMinutes() / 60`)
 - `.toDays()` returns a `long` describing the number of days in this duration (`.toHours() / 24`)

We include unit tests for each method in addition to a type call theorem for all of them.

The changes also include fixes for an embarrassing mistake in the `MILLISECONDS_PER_HOUR` definition.

### DIFF
--- a/cedar-lean/Cedar/Spec/Ext/Datetime.lean
+++ b/cedar-lean/Cedar/Spec/Ext/Datetime.lean
@@ -138,7 +138,7 @@ instance : Coe Int64 Duration where
 
 def MILLISECONDS_PER_SECOND: Int := 1000
 def MILLISECONDS_PER_MINUTE: Int := 60000
-def MILLISECONDS_PER_HOUR: Int := 360000
+def MILLISECONDS_PER_HOUR: Int := 3600000
 def MILLISECONDS_PER_DAY: Int := 86400000
 
 ----- Definitions -----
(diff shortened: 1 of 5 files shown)


shaobo-he-aws (approved):

chaluli:
> I'm actually a little bit concerned about the rounding of `/`, particularly for negative numbers. But we have unit tests so it should be fine.

`Int64.div` will convert both `Int64`s to `BitVec`s then use `BitVec.sdiv` (signed division) which removes the signs, performs unsigned division `Bitvec.udiv` then restores the appropriate sign. So, `Int64.div` will round towards zero.

E.g., `duration("25h").toDays = 1` and `duration("-25h").toDays = -1`.

PR #579, opened 2025-03-13
Add protobuf parsing `offset` when used as a `datetime` constructor

john-h-kastner-aws:
This PR updates protobuf value parsing to handle `offset` as an extension value constructor. The evaluator will sometimes return a value as the result of a `datetime` expression using the `offset` function as a constructor.

For example:

```console
[jkastner@ cedar]$ cedar evaluate 'datetime("0000-01-01").offset(duration("-365d"))' --principal 'P::""' --action 'A::""' --resource 'R::""'

(datetime("1970-01-01")).offset(duration("-62198755200000ms"))
```

This is necessary because `datetime` will not parse strings with any year before year zero.

### DIFF
--- a/cedar-lean/CedarProto/Value.lean
+++ b/cedar-lean/CedarProto/Value.lean
@@ -49,6 +49,12 @@ private def extExprToValue (xfn : ExtFun) (args : List Expr) : Except String Val
   | .duration, [.lit (.string s)] => match Spec.Ext.Datetime.Duration.parse s with
     | .some v => .ok $ .ext (.duration v)
     | .none => .error s!"exprToValue: failed to parse duration {s}"
+  | .offset, [.call xfn₁ args₁, .call xfn₂ args₂] => do
+    let arg₁ ← extExprToValue xfn₁ args₁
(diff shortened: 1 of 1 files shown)


cedar-policy-generators/src/expr.rs
```diff
@@ -1849,6 +1844,60 @@ impl<'a> ExprGenerator<'a> {
         }
     }
 
+    /// Generate a call for an extension constructor producing `target_type`.
+    /// This functions is used to generically produce either `AttrValue`
+    /// extensions constructors `Expr` extension constructor by passing `mk_str`
+    /// and `mk_ext` function to build string literals and extension function
+    /// calls for these types.
```
cdisselkoen, line 1851:
```suggestion
    /// This functions is used to produce either `AttrValue`
    /// extension constructors, or `Expr` extension constructors, by passing `mk_str`
    /// and `mk_ext` functions that build string literals and extension function
    /// calls for these types.
```

cdisselkoen (approved):

shaobo-he-aws (approved):
LGTM

PR #580, opened 2025-03-13
Fix datetime parsing for extra digits and negative years

john-h-kastner-aws:
Fixes datetime parsing for extra digits in some positions (e.g., `00011-12-13`) and for negative years (`-0001-01-01`)

 We use the format `uuuu` specifier for the year which is documented as "Displays the year in a four-digit format (e.g., "2004" or "-1000")", but it accepts years with more digits. The behavior when parsing isn't documented as far as I can tell, but the observed behavior is likely unintended. The same issue exists for months, days, hours, minutes and second. I've fixed this by adding an extra check into the datetime parsing function. 

Fixing for negative years is simpler. We  used the `uuuu` format specifier which is documented as supporting negative years. I fixed this by switching to `yyyyy`. This passes existing tests, but I don't know why we're using `uuuu` in the first place, so maybe someone will want to tell me this is wrong. We could also implement this as an extra check looking for a `-` at the start of the string.


shaobo-he-aws:
Were they found by DRT?

shaobo-he-aws (approved):
LGTM

shaobo-he-aws:
Btw, we should probably want to report this issue to the Lean library maintainers.

cdisselkoen (approved):

cdisselkoen (approved):

PR #583, opened 2025-03-17
Let target `roundtrip-entities` fail fast when an attribute value is a `datetime`

shaobo-he-aws:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cedar-drt/fuzz/fuzz_targets/roundtrip-entities.rs
+++ b/cedar-drt/fuzz/fuzz_targets/roundtrip-entities.rs
@@ -97,7 +97,9 @@ fuzz_target!(|input: FuzzTargetInput| {
             // 1, which involves calls to `datetime` and `offset`.
             return;
         }
-        _ => panic!("Should be able to serialize entities to JSON"),
+        Err(err) => {
+            panic!("Should be able to serialize entities to JSON, instead got error: {err}")
+        }
     };
 
     let eparser: EntityJsonParser<'_, '_, NoEntitiesSchema> = EntityJsonParser::new(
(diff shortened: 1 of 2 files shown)


john-h-kastner-aws (commented):
we shouldn't need to disable `offset` entirely. These values work fine in the differential tests, it's just the one roundtrip target that can't handle them

john-h-kastner-aws (approved):
looks great

cedar-drt/fuzz/fuzz_targets/roundtrip-entities.rs
```diff
@@ -89,6 +89,14 @@ fuzz_target!(|input: FuzzTargetInput| {
...
         }
+        Err(EntitiesError::Serialization(JsonSerializationError::ExtnCall2OrMoreArguments(
+            err,
+        ))) if err.to_string().contains("offset") => {
+            // Serializing to JSON is expected to fail when there's a record
+            // attribute of type `datetime` and it represents a time before AD
+            // 1, which involves calls to `datetime` and `offset`.
+            return;
+        }
         _ => panic!("Should be able to serialize entities to JSON"),
```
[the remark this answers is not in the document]
shaobo-he-aws: I let the error get printed when the program panics.

cedar-drt/fuzz/fuzz_targets/roundtrip-entities.rs
chaluli, line 1: Why does this specific error mean that the time represented will be before 1 AD? Doesn't this only mean that it contains an offset?

If not, why is the error so nondescript? If yes, I think we need to update the comment and explain why that is okay.

john-h-kastner-aws: the actual problem isn't negative years. It's with extension functions taking two or more arguments, and the `offset` function happens to be the only such function we'll generate

chaluli: Okay, then I definitely propose the comment be updated to reflect this rather than the current comment which was confusing.

shaobo-he-aws: Updated comments.

chaluli (approved):

john-h-kastner-aws (approved):

PR #586, opened 2025-03-25
Don't generate extension types in schema when extensions are disabled

lucajone:
Currently, even if `ABACSettings::extensions_enabled` is set to `false`, this function can still return an extension type. This resulted in [the following assertion](https://github.com/cedar-policy/cedar-spec/blob/6c2f1ab367b87b11674daf187a63d039114dbb5c/cedar-policy-generators/src/expr.rs#L2177) being triggered:

```
thread '<redacted>' panicked at cedar-policy-generators/src/expr.rs:2174:17:
shouldn't have TypeVariant::Extension with extensions disabled
```

when generating a hierarchy without extensions enabled. In particular, schemas were being generated with extension tag types.

I'm not very familiar with this code and couldn't see an easy way to unit test this, but please do suggest one! :)

### DIFF
--- a/cedar-policy-generators/src/schema.rs
+++ b/cedar-policy-generators/src/schema.rs
@@ -91,23 +91,12 @@ fn arbitrary_attrspec<N: From<ast::Name>>(
             attributes: attr_names
                 .into_iter()
                 .map(|attr| {
-                    let mut ty = arbitrary_typeofattribute_with_bounded_depth::<N>(
+                    let ty = arbitrary_typeofattribute_with_bounded_depth::<N>(
                         settings,
                         entity_types,
                         settings.max_depth,
                         u,
                     )?;
-                    if !settings.enable_extensions {
-                        // can't have extension types. regenerate until morale improves
-                        while ty.ty.is_extension().expect("DRT does not generate schema type using type defs, so `is_extension` should be `Some`") {
-                            ty = arbitrary_typeofattribute_with_bounded_depth::<N>(
-                                settings,
-                                entity_types,
-                                settings.max_depth,
-                                u,
-                            )?;
-                        }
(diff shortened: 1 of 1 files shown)


cedar-policy-generators/src/schema.rs
john-h-kastner-aws, line 1: Looks reasonable, but this should already be handled by https://github.com/cedar-policy/cedar-spec/blob/6c2f1ab367b87b11674daf187a63d039114dbb5c/cedar-policy-generators/src/schema.rs#L102

lucajone: I think the problematic use is [here](https://github.com/cedar-policy/cedar-spec/blob/6c2f1ab367b87b11674daf187a63d039114dbb5c/cedar-policy-generators/src/schema.rs#L994)?

john-h-kastner-aws: That makes sense. Looks like we forgot to check this flag when adding entity tags.

I think we tend to prefer the current regenerate-in-loop strategy to returning an error since it doesn't abort the whole input generation attempt. We should probably move that logic into a wrapper around `arbitrary_schematype_with_bounded_depth` so both callers can use it

lucajone: Right, that seems reasonable. I'll have a shot at that!

lucajone: How do you feel about just making `arbitrary_schematype_with_bounded_depth` respect the value in `settings`, by handling the loop itself? I think it will make the recursive cases easier.

john-h-kastner-aws (approved):
looks great

cdisselkoen (approved):

PR #590, opened 2025-03-26
Add some more Map theorems 

john-h-kastner-aws:
A couple small theorems carved off #533

### DIFF
--- a/cedar-lean/Cedar/Thm/Data/Map.lean
+++ b/cedar-lean/Cedar/Thm/Data/Map.lean
@@ -371,6 +371,38 @@ theorem find?_none_all_absent [LT α] [DecidableLT α] [StrictLT α] [DecidableE
   specialize hf (k, v) hc
   simp only [not_true_eq_false] at hf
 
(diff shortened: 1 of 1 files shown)


emina (approved):

cedar-lean/Cedar/Thm/Data/Map.lean
```diff
@@ -371,6 +371,38 @@ theorem find?_none_all_absent [LT α] [DecidableLT α] [StrictLT α] [DecidableE
   specialize hf (k, v) hc
   simp only [not_true_eq_false] at hf
 
+theorem all_absent_find?_none [LT α] [DecidableLT α] [StrictLT α] [DecidableEq α] {m : Map α β} {k : α} :
```
shaobo-he-aws, line 374: QQ: is this theorem not "public API"? If it is, we should add doc comments.

Nit 1: IMO, using induction will make proofs a bit easier.
Nit 2: I wonder if it helps if we proof something like `=` as opposed to implies.

shaobo-he-aws: Nvm.

PR #591, opened 2025-03-26
avoiding generating invalid EIDs for enum entity types

cdisselkoen:
*Issue #, if available:*

Makes #1550 unnecessary, I hope.

*Description of changes:*

Avoids generating invalid EIDs for enum entity types.
1) When generating a new `Hierarchy`, when generating a new EID of an enum entity type, removed the 20% coinflip to generate an invalid EID.  Now always generates a valid EID if the entity type is an enum entity type.
2) For `arbitrary_uid_with_type()` (which is used for generating the principal/resource for requests), now takes a `schema` so it can determine if the requested type is an enum entity type, and reuses the above method. Preserves the previous behavior of (with some probability) generating a UID that doesn't exist in the hierarchy, but will no longer generate one that is invalid based on a schema enum-entity declaration.

### DIFF
--- a/cedar-policy-generators/src/hierarchy.rs
+++ b/cedar-policy-generators/src/hierarchy.rs
@@ -153,9 +153,13 @@ impl Hierarchy {
         )
     }
 
-    /// generate an arbitrary uid based on the hierarchy, with the given typename
+    /// generate an arbitrary uid based on the hierarchy and schema, with the given typename.
+    ///
+    /// Schema is used in order to ensure that if it's an enum entity type, we
+    /// pick one of the valid EIDs for that type.
     pub fn arbitrary_uid_with_type(
         &self,
+        schema: Option<&Schema>,
         typename: &ast::EntityType,
         u: &mut Unstructured<'_>,
     ) -> Result<EntityUID> {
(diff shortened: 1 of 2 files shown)


cedar-policy-generators/src/hierarchy.rs
```diff
@@ -171,11 +175,11 @@ impl Hierarchy {
...
             Ok(uid.clone())
         } else {
-            Ok(EntityUID::from_components(
-                typename.clone(),
-                u.arbitrary()?,
-                None,
-            ))
+            // generate a UID that (probably) doesn't exist, but, still is schema-compatible
+            let choices = schema
+                .map(|schema| schema.get_uid_choices(typename))
```
john-h-kastner-aws, line 180: Can you rename `get_uid_choices` to some thing like `get_uid_enum_choices`? It wasn't immediately clear to me where this change was actually handling enums until I looked at the definition of this function.

john-h-kastner-aws (approved):

cedar-policy-generators/src/hierarchy.rs
cdisselkoen: Done.

shaobo-he-aws:
QQ: would this affect DRT's capability of testing invalid EIDs of enumerated entity types?

shaobo-he-aws (approved):

shaobo-he-aws:
Makes sense. That being said, we should probably create an issue so that we keep the limitation in mind.

cdisselkoen:
Opened #593 for that followup.

PR #595, opened 2025-04-01
TPE soundness proofs

shaobo-he-aws:
*Issue #, if available:*

*Description of changes:*


cedar-lean/Cedar/Thm/Validation/Typechecker/Types.lean
```diff
@@ -105,7 +105,7 @@ def InstanceOfEntitySchema (entities : Entities) (ets: EntitySchema) : Prop :=
       IsValidEntityEID entry uid.eid ∧
       InstanceOfType data.attrs (.record entry.attrs) ∧
       (∀ ancestor, ancestor ∈ data.ancestors → ancestor.ty ∈ entry.ancestors) ∧
-      InstanceOfEntityTags data entry
+      InstanceOfEntityTags data.tags entry
```
emina, line 108: Why did we need this change?

cedar-lean/Cedar/Thm/Validation/Typechecker/Types.lean
```diff
@@ -81,10 +81,10 @@ def InstanceOfRequestType (request : Request) (reqty : RequestType) : Prop :=
   InstanceOfEntityType request.resource reqty.resource ∧
   InstanceOfType request.context (.record reqty.context)
 
-def InstanceOfEntityTags (data : EntityData) (entry : EntitySchemaEntry) : Prop :=
+def InstanceOfEntityTags (tags : Map Tag Value) (entry : EntitySchemaEntry) : Prop :=
```
emina, line 84: Why is this change needed (taking a Map rather EntityData as input)?

cedar-lean/Cedar/Thm/Data/Control.lean
```diff
@@ -61,3 +61,92 @@ theorem do_ok {res : Except ε α} {f : α → β} :
   (do let v ← res ; .ok (f v)) = .ok b ↔
   ∃ a, res = .ok a ∧ f a = b
 := by cases res <;> simp
+
+theorem do_ok' {res : Except ε α} {f : α → Except ε β} :
```
emina, line 65: What is the naming convention for primed vs subscripted names for the new theorems?

cedar-lean/Cedar/Thm/TPE.lean
```diff
@@ -0,0 +1,147 @@
...
+/-- The main lemma: Evaluating a residual derived from partially evaluating
+a well-typed expression is equivalent to that of evaluating the original
+expression, provided that requests and entities are consistent. The equivalency
+is defined using `Except.toOption`.
+-/
+theorem partial_evaluate_is_sound
+  {x : TypedExpr}
+  {req₁ : Request}
+  {es₁ : Entities}
+  {req₂ : PartialRequest}
```
emina, line 44: To make it easier to read the theorem, I suggest using `req` and `es` for concrete entities, and `preq` and `pes` (or something like that) for their partial variants.  Subscripts suggest that these are all values of the same type (which is not the case).  Using the `p` prefix suggests more strongly (at least to me :)) that these things are "partial" versions of their concrete countereparts.

cedar-lean/Cedar/Thm/TPE.lean
```diff
@@ -0,0 +1,147 @@
...
+evaluating the input policy, given valid and consistent requests and entities.
+The equivalence is w.r.t authorization results. That is, the evaluation results
+are strictly equal when they are `.ok` or both errors (captured by
+`Except.toOption`).
+-/
+theorem partial_evaluate_policy_is_sound
+  {schema : Schema}
+  {residual : Residual}
+  {policy : Policy}
+  {req₁ : Request}
```
emina, line 93: See comment above on re-naming `req` and `es` to distinguish more strongly between concrete and partial variants.

cedar-lean/Cedar/Thm/Validation/Typechecker.lean
```diff
@@ -98,3 +98,522 @@ theorem type_of_is_sound {e : Expr} {c₁ c₂ : Capabilities} {env : Environmen
...
+  List.mapM f xs =
+  List.mapM g ys
+:= by
+  intro h
+  cases h
+  case nil => simp only [List.mapM_nil]
+  case cons h₁ h₂ =>
+    simp only [List.mapM_cons, h₁, forall₂_impies_mapM_eq f g h₂, bind_pure_comp]
+
+/- TODO: make it a general theorem -/
```
emina, line 113: We might want to make these theorems `private` if they are specific to the `type_of_preserves_evaluation_results`.

cedar-lean/Cedar/Thm/TPE/Input.lean
```diff
@@ -0,0 +1,259 @@
...
+  PartialIsValid (· = req₁.resource) req₂.resource.asEntityUID  ∧
+  PartialIsValid (· = req₁.context) req₂.context
+
+def EntitiesAreConsistent (es₁ : Entities) (es₂ : PartialEntities) : Prop :=
+   ∀ a e₂, es₂.find? a = some e₂ → (∃ e₁, es₁.find? a = some e₁ ∧
+    PartialIsValid (· = e₁.attrs) e₂.attrs ∧
+    PartialIsValid (· = e₁.ancestors) e₂.ancestors  ∧
+    PartialIsValid (· = e₁.tags) e₂.tags)
+
+/- should have a better name like `abstracts`.
```
emina, line 82: I think we probably want something along the lines of `Refines` (i.e., the concrete input refines the partial input).

cedar-lean/Cedar/Thm/TPE/Input.lean
```diff
@@ -0,0 +1,259 @@
...
+  request.action = env.reqty.action ∧
+  PartialIsValid (InstanceOfEntityType · env.reqty.resource) request.resource.asEntityUID ∧
+  PartialIsValid (InstanceOfType · (.record env.reqty.context)) request.context
+
+def PartialEntitiesMatchEnvironment (env : Environment) (entities: PartialEntities) : Prop :=
+  ∀ uid data, entities.find? uid = some data →
+    ∃ entry, env.ets.find? uid.ty = some entry ∧
+      IsValidEntityEID entry uid.eid ∧
+      PartialIsValid (InstanceOfType · (.record entry.attrs)) data.attrs ∧
+      PartialIsValid (λ ancestors => ∀ ancestor, ancestor ∈ ancestors → ancestor.ty ∈ entry.ancestors) data.ancestors ∧
```
emina, line 64: This can be written more compactly as

```
PartialIsValid (λ ancestors => ∀ ancestor ∈ ancestors, ancestor.ty ∈ entry.ancestors) data.ancestors ∧
```

cedar-lean/Cedar/Thm/Validation/Typechecker.lean
```diff
@@ -98,3 +98,522 @@ theorem type_of_is_sound {e : Expr} {c₁ c₂ : Capabilities} {env : Environmen
...
+      simp only [TypedExpr.toExpr, evaluate, hᵢ₁, hᵢ₂]
+    case _ =>
+      rcases h₃₃ with ⟨h₃₃, _⟩
+      subst h₃₃
+      simp only [TypedExpr.toExpr, evaluate, hᵢ₁, hᵢ₂]
+    case _ =>
+      rcases h₃₃ with ⟨h₃₃, _⟩
+      subst h₃₃
+      simp only [TypedExpr.toExpr, evaluate, hᵢ₁, hᵢ₂]
+    case _ =>
```
emina, line 458: We should be able to collapse all of these repeated cases with 

```
any_goals {
      simp [do_ok] at h₃₃
      rcases h₃₃ with ⟨_, h₃₃₁, h₃₃₂⟩
      subst h₃₃₂
}
any_goals {
      rcases h₃₃ with ⟨h₃₃, _⟩
      subst h₃₃
      simp only [TypedExpr.toExpr, evaluate, hᵢ₁, hᵢ₂]
}
```

cedar-lean/Cedar/Thm/Validation/Typechecker/Types.lean
shaobo-he-aws: It makes it easier to specify `PartialRequestMatchesEnvironment`, which is actually unused. So I will restore the definition.

cedar-lean/Cedar/Thm/Data/Control.lean
shaobo-he-aws: Came up with a better name.

cedar-lean/Cedar/Thm/Validation/Typechecker.lean
shaobo-he-aws: TIL

emina (approved):
LGTM!

cedar-lean/Cedar/Thm/TPE.lean
```diff
@@ -0,0 +1,147 @@
...
+  rcases h₁ with ⟨_, ⟨_, h₁₁⟩, h₁₂⟩
+  simp [Except.mapError] at h₁₁
+  split at h₁₁ <;> try cases h₁₁
+  rename_i env heq₁ _ ty _ _ heq₂
+  simp [isValidAndConsistent] at h₂
+  split at h₂ <;> try cases h₂
+  rename_i heq₃
+  simp [heq₁] at heq₃
+  subst heq₃
+  rcases do_eq_ok₂ h₂ with ⟨h₂₁, h₂₂⟩
```
john-h-kastner-aws, line 116: Is `rcases` here the same as `have ⟨h₂₁, h₂₂⟩ := ...`? I feel like `have` is a little clearer.

cedar-lean/Cedar/Thm/Validation/Typechecker.lean
```diff
@@ -98,3 +98,345 @@ theorem type_of_is_sound {e : Expr} {c₁ c₂ : Capabilities} {env : Environmen
...
+evaluates to the same result as the input expression.
+-/
+theorem type_of_preserves_evaluation_results {e : Expr} {c₁ c₂ : Capabilities} {env : Environment} {ty : TypedExpr} {request : Request} {entities : Entities} :
+  CapabilitiesInvariant c₁ request entities →
+  RequestAndEntitiesMatchEnvironment env request entities →
+  typeOf e c₁ env = .ok (ty, c₂) →
+  evaluate e request entities = evaluate ty.toExpr request entities
+:= by
+  intro h₁ h₂ h₃
+  induction e, c₁, env using typeOf.induct generalizing ty c₂
```
john-h-kastner-aws, line 112: The existing type soundness proof is recursive instead of making recursive call which seems to avoid some of this complexity (at the cost of needing the termination proof). Any particular reason to use induction here instead?

cedar-lean/Cedar/Thm/TPE/Soundness.lean
john-h-kastner-aws, line 1: Probably worth pulling the cases out of this into individual files if it's anything like the other larger proofs we have

cedar-lean/Cedar/Thm/TPE/Soundness.lean
```diff
@@ -0,0 +1,1075 @@
...
+
+theorem partial_evaluate_is_sound_var
+{req : Request}
+{es : Entities}
+{preq : PartialRequest}
+{pes : PartialEntities}
+{v : Var}
+{ty : CedarType}
+(h₄ : RequestAndEntitiesRefine req es preq pes) :
+  Except.toOption (Spec.evaluate (TypedExpr.var v ty).toExpr req es) =
```
john-h-kastner-aws, line 157: Might be nice to lift this into a `def PartialEvaluateIsSound (tx : TypedExpr) := ...` to make the statement of this lemma and others cleaner

cedar-lean/Cedar/Thm/TPE.lean
```diff
@@ -0,0 +1,147 @@
...
+open Cedar.TPE
+open Cedar.Spec
+open Cedar.Validation
+open Cedar.Thm
+
+/-- The main lemma: Evaluating a residual derived from partially evaluating
+a well-typed expression is equivalent to that of evaluating the original
+expression, provided that requests and entities are consistent. The equivalency
+is defined using `Except.toOption`.
+-/
```
john-h-kastner-aws, line 39: > The equivalency is defined using `Except.toOption`

So, if they both error, the exact error doesn't have to be the same? Probably worth spelling that out in the comment

john-h-kastner-aws (approved):
Looks good, but I'd like a slightly clearer comment on the main theorem

cedar-lean/Cedar/Thm/Validation/Typechecker.lean
shaobo-he-aws: Avoiding termination proof is a huge advantage to me. Functional induction essentially gives us inductive hypothesis for free and avoid the need to have a property like `TypeOfIsSound` if we want to refer to the main theorem in the lemmas.

cedar-lean/Cedar/Thm/TPE/Soundness.lean
shaobo-he-aws: I think these proofs are not big compared to type checking soundness proofs so having multiple files potentially decreases readability and maintainability. In fact, proofs of well-typedness are also in a single file.

cedar-lean/Cedar/Thm/TPE/Soundness.lean
shaobo-he-aws: Good idea but I think that it's probably not worth it since it's a simple property. Adding this property will leads to a lot of unfolding in the proofs.

cedar-lean/Cedar/Thm/TPE.lean
shaobo-he-aws: Correct. Will update the comment.

cedar-lean/Cedar/Thm/TPE/Soundness.lean
john-h-kastner-aws: I personally expect that at 1000 lines editing this file will be noticeable slow at least depending on the hardware you're working on. But, this is an easy reorganization we can do later

john-h-kastner-aws: IMO it makes the statement of the IH in the set and record cases _much_ easier to follow, so it's worth a couple extra `unfolds` as necessary, but it's not a big deal

john-h-kastner-aws (approved):

PR #597, opened 2025-04-07
Set Rust toolchain to 1.85 on `fuzz` crate

adpaco-aws:
*Description of changes:* We are seeing linking failures when building targets that use the Lean models. This seems related to the Rust 1.86 release as stable. Pinning down the Rust toolchain to 1.85 seems to work for the moment.


chaluli (approved):
The change looks good. Not sure why the checks aren't passing. This change doesn't look like it would cause the checks to fail.

katherine-hough (approved):

katherine-hough (approved):

chaluli (approved):

PR #574, opened 2025-03-10
Add differential test target for Level Validation

john-h-kastner-aws:
This target is currently expected to fail due to multiple known issues

Issue #126, opened 2023-09-28
Resolve known mismatch in definitional validator


khieta:
### Category

Dafny formalization

### Describe the feature you'd like to request

The `cedar-policy` validator will always return an error for an unrecognized entity or action, even if that part of the expression should be excluded from typechecking (e.g., `true || Undefined::"foo"` should be well typed due to short-circuiting). The definitional validator currently does not have this behavior. We should need to decide which behavior we think is correct and then fix either the `cedar-policy` validator or the definitional validator. We then also need to remove [this piece of code](https://github.com/cedar-policy/cedar-spec/blob/fa6d9aadecaa7b4221d8cfa92351e81b97aea67f/cedar-drt/fuzz/src/lib.rs#L173C12-L177C58) that hides this difference during DRT.

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

john-h-kastner-aws:
Some known discrepancies:

* The Rust validator check for unknown actions and entity types before typechecking in a way that means short circuiting doesn't apply. `false && Unknown::""` or `false && Action::"Unknown"` are reported as errors by Rust, but not by Lean.
* The Rust validator only gives singleton type of `==` comparing entity literals. So `1 == 2` has type `False` in the Lean but `Bool` in Rust.
* The behavior if `in` is different. The Lean implementation is somewhat simpler, so the Rust should adopt it.

cdisselkoen:
For more on the second bullet, see https://github.com/cedar-policy/cedar/issues/104

PR #606, opened 2025-04-11
Remove stricter check for level validation drt

john-h-kastner-aws:
I added this stricter check in #574, but it doesn't hold due to the issues tracked in #126.

We still test that any error reported by Lean is also reported by Rust, preserving soundness.


katherine-hough (approved):

shaobo-he-aws (approved):
LGTM

PR #607, opened 2025-04-11
Deduplicate lemma about `typeOf .lit` added for levels and PE proofs

john-h-kastner-aws:
*Issue #, if available:*

*Description of changes:*


chaluli (approved):
Looks great! Happy to deduplicate proofs when possible.

cdisselkoen (approved):
Any reason why you chose to keep the name `type_of_lit_inversion` rather than `type_of_lit_is_lit`?

john-h-kastner-aws:
> Any reason why you chose to keep the name `type_of_lit_inversion` rather than `type_of_lit_is_lit`?

To match the other typechecking lemmas like `type_of_getAttr_inversion` and `type_of_set_inversion` which prove analogues (though somewhat more complex) propositions for the corresponding expressions.

shaobo-he-aws (approved):
Thank you for finding it out.

PR #609, opened 2025-04-21
Remove redundant typechecking from entity manifest target

john-h-kastner-aws:
Typechecking is[ done inside the `compute_entity_manifest` function](https://github.com/cedar-policy/cedar/blob/df11acefbfd7ce5c893acd9c4dbc139fa61bf6bc/cedar-policy-validator/src/entity_manifest.rs#L443), so we just need to check for the `EntityManifestError::Validation` error variant


shaobo-he-aws (approved):
LGTM

katherine-hough (approved):

Issue #610, opened 2025-04-22
Limit number of request environments in generated schema


john-h-kastner-aws:
### Category

DRT generator(s)

### Describe the feature you'd like to request

We should limit the maximum number of request environments that can be defined by a generated schema.

While it doesn't happen often, we can rarely generate long `appliesTo` lists for actions in the schema. A schema with large `appliesTo` lists can quickly generate a very large number of request environments because request environments are computed from the product of principal and resource applies to lists. This can lead to slow units when we need to validate a policy in every defined request environment.

We shouldn't set this limit too low since it is interesting to test at ~100 environments (maybe a couple actions each applying to ~10 principals and resources), but previous slow units were found in the rare case of a schema ~10000 request environments. 

### Describe alternatives you've considered

.

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

PR #613, opened 2025-04-28
Fail fast when the number of request environments hits a limit

shaobo-he-aws:
*Issue #, if available:*

#610 

*Description of changes:*

### DIFF
--- a/cedar-policy-generators/src/schema.rs
+++ b/cedar-policy-generators/src/schema.rs
@@ -954,7 +954,7 @@ impl Schema {
     // An upper bound on the number of request environment a schema can produce
     // See https://github.com/cedar-policy/cedar-spec/issues/610 for the
     // motivation why we want this limit
-    pub(crate) const REQUEST_ENV_LIMIT: usize = 128;
+    pub(crate) const PER_ACTION_REQUEST_ENV_LIMIT: usize = 128;
 
     /// Get an arbitrary `Schema`.
     pub fn arbitrary(settings: ABACSettings, u: &mut Unstructured<'_>) -> Result<Schema> {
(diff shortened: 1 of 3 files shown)


chaluli (approved):
Personally, I'm fine with this solution given the case it auto-fails is expected to be rare and this will just trigger a regeneration.

Though, my understanding we wanted to instead modify the random generation so that |principles| * |resources| would be below some given limit.

john-h-kastner-aws (commented):
I did have something in mind more like what Charlie suggests.

This approach is probably fine, but if we go with it we should at least adjust `pick_entity_types` so that the `principalTypes` and `resourceTypes` lists are always individually less than `REQUEST_ENV_LIMIT`

shaobo-he-aws:
> I did have something in mind more like what Charlie suggests.
> 
> This approach is probably fine, but if we go with it we should at least adjust `pick_entity_types` so that the `principalTypes` and `resourceTypes` lists are always individually less than `REQUEST_ENV_LIMIT`

Done.

cedar-policy-generators/src/schema.rs
```diff
@@ -1108,6 +1113,13 @@ impl Schema {
                             } else {
                                 principal_and_resource_types_exist = true;
                             }
+                            let req_env_num =
+                                picked_principal_types.len() * picked_resource_types.len();
+                            // Fail fast if the number of request environment
+                            // number is too large
+                            if req_env_num > Self::REQUEST_ENV_LIMIT {
+                                return Err(Error::TooManyReqEnvs(req_env_num));
```
katherine-hough, line 1121: Is there a way to rewrite the generator to ensure that the number of request environments is less than the threshold by construction?

cedar-policy-generators/src/schema.rs
```diff
@@ -951,6 +951,11 @@ impl Schema {
         }
     }
 
+    // An upper bound on the number of request environment a schema can produce
+    // See https://github.com/cedar-policy/cedar-spec/issues/610 for the
+    // motivation why we want this limit
+    pub(crate) const REQUEST_ENV_LIMIT: usize = 128;
```
john-h-kastner-aws, line 957: Rename to `PER_ACTION_REQUEST_ENV_LIMIT`

cedar-policy-generators/src/schema.rs
```diff
@@ -1063,7 +1068,7 @@ impl Schema {
                                  u: &mut Unstructured<'_>|
          -> Result<Vec<ast::InternalName>> {
             // Pre-select the number of entity types (minimum 1), then randomly select that many indices
-            let num = u.int_in_range(1..=entity_types.len()).unwrap();
+            let num = u.int_in_range(1..=entity_types.len()).unwrap() % Self::REQUEST_ENV_LIMIT;
```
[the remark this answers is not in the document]
shaobo-he-aws: Done.

cedar-policy-generators/src/schema.rs
shaobo-he-aws: Done.

PR #621, opened 2025-05-09
Share fuzzing code between the two validation-drt targets

john-h-kastner-aws:
We should be able to do the same with the `abac` and `validation-pbt` targets, but the fuzz target implementations there aren't exactly the same. It looks like the difference is only in logging code, but we'd want to make sure we don't lose anything important

### DIFF
--- a/cedar-drt/fuzz/fuzz_targets/validation-drt.rs
+++ b/cedar-drt/fuzz/fuzz_targets/validation-drt.rs
@@ -15,75 +15,8 @@
  */
 
 #![no_main]
-use cedar_drt::*;
-use cedar_drt_inner::*;
-use cedar_policy_core::ast;
-use cedar_policy_generators::{
-    abac::ABACPolicy, hierarchy::HierarchyGenerator, schema::Schema, settings::ABACSettings,
-};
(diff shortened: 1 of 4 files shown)


cedar-drt/fuzz/fuzz_targets/validation-drt.rs
```diff
@@ -15,75 +15,8 @@
...
-        policyset.add_static(policy).unwrap();
-        debug!("Policies: {policyset}");
-
-        // run the policy through both validators and compare the result
-        let (_, total_dur) =
-            time_function(|| run_val_test(&def_impl, schema, &policyset, ValidationMode::Strict));
-        info!("{}{}", TOTAL_MSG, total_dur.as_nanos());
-    }
-});
+fuzz_target!(|input: validation_drt::FuzzTargetInput<true>| validation_drt::fuzz_target(input));
```
cdisselkoen, line 22:
```suggestion
fuzz_target!(|input: validation_drt::FuzzTargetInput<false>| validation_drt::fuzz_target(input));
```

cdisselkoen (changes requested):

cdisselkoen (approved):

adpaco-aws (approved):

PR #624, opened 2025-05-30
Adds cedar-cli: a command-line-interface for using the lean formalization of Cedar.

chaluli:
Adds a rust based CLI that uses [`cedar`](https://github.com/cedar-policy/cedar) to parse policies, schemas, entities, requests, contexts, etc. and then uses the lean formalization of cedar (`cedar-lean`) to evaluate a users request.

### DIFF
--- a/cedar-cli/Cargo.toml
+++ b/cedar-cli/Cargo.toml
@@ -1,0 +1,22 @@
+[package]
+name = "cedar-cli"
+edition = "2021"
+version = "0.1.0"
+publish = false
+
+[dependencies]
+cedar-policy = { path="../cedar/cedar-policy/", version = "=4.4.0", features = ["protobufs", "permissive-validate"] }
+clap = { version = "4.5.36", features = ["derive"] }
+lean-sys = { version = "0.0.8", default-features = false }
+serde = "1"
+serde_json = "1.0"
+thiserror = "2.0"
+prost = "0.13"
+itertools = "0.14.0"
+miette = "7.6.0"
+prettytable-rs = "0.10"
+
+[build-dependencies]
+prost-build = "0.13"
+
+[profile.release]
--- a/cedar-cli/README.md
+++ b/cedar-cli/README.md
@@ -1,0 +1,227 @@
+# Cedar CLI
+
+This directory contains a command line interface (CLI) for interacting with the Lean formalization of Cedar. This CLI uses [`Cedar`](https://github.com/cedar-policy/cedar) to parse policies, schemas, entities, and requests and then uses the Lean formalization of Cedar to perform validation, evaluation, authorization, and analysis.
+
+## Build
+
+### Prerequisites:
+
+#### Install Lean
+For more details read about Lean's version manager [elan](https://github.com/leanprover/elan).
+
+```
+curl https://elan.lean-lang.org/elan-init.sh -sSf | sh
+```
+
+#### Install cvc5 (>= v1.0.5)
+The analysis features provided by this CLI requires a local install of [cvc5](https://github.com/cvc5/cvc5) with version [1.0.5](https://github.com/cvc5/cvc5/releases/tag/cvc5-1.0.5) or greater.
+
+Once you install cvc5. Export the following Environment variable which allows this CLI to know where cvc5 is installed. Consider adding the export to your `~/.bashrc` or `~/.profile` so that the environment variable is exported to all new terminal sessions.
+
+```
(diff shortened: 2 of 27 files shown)


lianah (approved):

lianah (approved):

cedar-lean/SymFFI/Main.lean
```diff
@@ -0,0 +1,502 @@
...
+  1.) .error err_message if there was in error in parsing or running the solver
+  2.) .ok true if the solver could prove `req` holds
+  3.) .ok false if the solver could prove `req` does not hold
+-/
+@[export runCheckNeverErrors] unsafe def runCheckNeverErrors (req : ByteArray) : String :=
+  let result : Except String Bool :=
+    match parseCheckPolicyReq req with
+    | .error s => .error s!"checkNeverErrors: {s}"
+    | .ok (policy, εnv) =>
+      match solve Solver.cvc5 (Cedar.SymCC.checkNeverErrors policy εnv) with
```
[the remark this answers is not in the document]
chaluli: Not sure how much lean would like this.

But perhaps we could export a function that creates a solver instance and returns a reference to the solver (thorugh the FFI). Then all of the current exported functions take both the request and a pointer to the solver instance to use.

Then when the pointer falls out of scope on the rust side it will deallocate the corresponding pointer within the lean code?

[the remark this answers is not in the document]
chaluli: Another option might be to set up some kind of statically allocated CVC5 solver instance that is used on every request. We might be able to do something smart like allocate upon first use.

cedar-cli/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "cedar-cli"
+edition = "2021"
+version = "0.1.0"
+publish = false
+
+[dependencies]
+cedar-policy = { path="../cedar/cedar-policy/", version = "=4.4.0", features = ["protobufs", "permissive-validate"] }
```
john-h-kastner-aws, line 8: Why enable `permissive-validate`? That validation mode isn't modeled in Lean

cedar-cli/src/lean_ffi.rs
john-h-kastner-aws, line 1: Is this duplicating a lot of code from the DRT crate?

cedar-cli/README.md
```diff
@@ -0,0 +1,227 @@
...
+Request Environment Options:
+      --principal-type <PRINCIPAL_TYPE_NAME>
+          Restrict Analysis to Request Environments for the given PrincipalType
+      --action-name <ACTION_NAME>
+          Restrict Analysis to Request Environmetns for the given Action
+      --resource-type <RESOURCE_TYPE_NAME>
+          Restrict Analysis to Request Environments for the given ResourceType
+```
+
+### Evaluation
```
john-h-kastner-aws, line 183: Ideally I'd like `authorize`, `evaluate` and `validate` to have the same interface as the Rust CLI. The `validate` API in particular stands out as different.

cedar-cli/build.rs
john-h-kastner-aws, line 1: Can we share build scripts with the DRT crate (to include the `set_env_vars.sh` and `build_lean_lib.sh` shell scripts if possible)?

cedar-cli/protobuf_schema/Messages.proto
john-h-kastner-aws, line 1: At least some of these definitions look like they're duplicated from DRT crate

cedar-cli/Cargo.toml
john-h-kastner-aws, line 1: Need to update CI to build/test/fmt this crate

cedar-cli/src/util.rs
```diff
@@ -0,0 +1,715 @@
...
+                Self::resource_type_fmt(rtype, f)
+            }
+        }
+    }
+}
+
+/// A simple wrapper around Cedar's RequestEnv type to allow for pretty-printing
+#[derive(Debug)]
+pub enum ReqEnv {
+    Env(RequestEnv),
```
john-h-kastner-aws, line 111: Any reason this is an `enum`?

john-h-kastner-aws (commented):
Mostly complaining code duplicated from the DRT crate (ofc I understand that this was developed outside this repository where sharing code would be inconvenient, but there should be room to refactor now)

john-h-kastner-aws:
Should also update the main and `cedar-lean`  `README.md` files to point to this crate

PR #628, opened 2025-06-05
Remove FFI interface from cedar-cli and create new create cedar-lean-ffi which can be used in both cedar-cli and cedar-drt.

chaluli:
Refactor `cedar-cli` to create a new crate`cedar-lean-ffi` which provides a reusable set of Rust-bindings to the Lean formalization of Cedar.

Additionally, fixes a bug in cedar-cli which caused the cli to parse json format schemas using the cedar-schema parser.

The new crate *should* be a superset of the needs for both `cedar-cli` and `cedar-drt`.

### DIFF
--- a/cedar-lean-ffi/src/lean_ffi.rs
+++ b/cedar-lean-ffi/src/lean_ffi.rs
@@ -167,7 +167,7 @@ impl AuthorizationResponse {
         let decision = match inner.decision.as_str() {
             "allow" => Decision::Allow,
             "deny" => Decision::Deny,
-            _ => return Err(FfiError::LeanDeserializationError),
+            _ => return Err(FfiError::LeanDeserializationError(inner.decision)),
         };
         let determining = inner
             .determining_policies
@@ -257,7 +257,7 @@ macro_rules! checkPolicy_func {
             match serde_json::from_str(&response) {
                 Ok(ResultDef::Ok(t)) => Ok(TimedResult::from_def(t)),
                 Ok(ResultDef::Error(s)) => Err(FfiError::LeanBackendError(s)),
-                Err(_) => Err(FfiError::LeanDeserializationError),
+                Err(_) => Err(FfiError::LeanDeserializationError(response)),
             }
         }
         pub fn $untimed_func_name(
@@ -290,7 +290,7 @@ macro_rules! checkPolicySet_func {
             match serde_json::from_str(&response) {
                 Ok(ResultDef::Ok(t)) => Ok(TimedResult::from_def(t)),
                 Ok(ResultDef::Error(s)) => Err(FfiError::LeanBackendError(s)),
-                Err(_) => Err(FfiError::LeanDeserializationError),
+                Err(_) => Err(FfiError::LeanDeserializationError(response)),
             }
         }
         pub fn $untimed_func_name(
@@ -329,7 +329,7 @@ macro_rules! comparePolicySet_func {
             match serde_json::from_str(&response) {
                 Ok(ResultDef::Ok(t)) => Ok(TimedResult::from_def(t)),
                 Ok(ResultDef::Error(s)) => Err(FfiError::LeanBackendError(s)),
-                Err(_) => Err(FfiError::LeanDeserializationError),
+                Err(_) => Err(FfiError::LeanDeserializationError(response)),
             }
         }
         pub fn $untimed_func_name(
@@ -526,7 +526,7 @@ impl CedarLeanFfi {
                 Ok(TimedResult::from_def(tdef))
             }
             Ok(ResultDef::Error(s)) => Err(FfiError::LeanBackendError(s)),
-            Err(_) => Err(FfiError::LeanDeserializationError),
+            Err(_) => Err(FfiError::LeanDeserializationError(response)),
         }
     }
     pub fn is_authorized(
@@ -555,7 +555,7 @@ impl CedarLeanFfi {
         match serde_json::from_str(&ret) {
             Ok(ResultDef::Ok(t)) => Ok(TimedResult::from_def(t)),
             Ok(ResultDef::Error(s)) => Err(FfiError::LeanBackendError(s)),
-            Err(_) => Err(FfiError::LeanDeserializationError),
+            Err(_) => Err(FfiError::LeanDeserializationError(ret)),
         }
     }
     pub fn print_evaluation(
@@ -591,7 +591,7 @@ impl CedarLeanFfi {
         match serde_json::from_str(&ret) {
             Ok(ResultDef::Ok(are_eq)) => Ok(TimedResult::from_def(are_eq)),
             Ok(ResultDef::Error(s)) => Err(FfiError::LeanBackendError(s)),
-            Err(_) => Err(FfiError::LeanDeserializationError),
+            Err(_) => Err(FfiError::LeanDeserializationError(ret)),
         }
     }
     pub fn check_evaluate(
@@ -621,7 +621,7 @@ impl CedarLeanFfi {
         match serde_json::from_str(&ret) {
             Ok(ResultDef::Ok(res)) => Ok(TimedResult::from_def(res)),
             Ok(ResultDef::Error(s)) => Err(FfiError::LeanBackendError(s)),
-            Err(_) => Err(FfiError::LeanDeserializationError),
+            Err(_) => Err(FfiError::LeanDeserializationError(ret)),
         }
     }
     pub fn validate(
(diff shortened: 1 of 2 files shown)


chaluli:
The goal of this refactor is to eliminate code dupilication between the `cedar-cli` and `cedar-drt` crates. This does the first part of this refactor by creating a new crate `cedar-lean-ffi` which provides re-usable Rust-bindings (in the form of a library) to the Lean formalization of Cedar.

I have refactored `cedar-cli` to use this new cate. In a future PR will update `cedar-drt` accordingly to do the same.

cedar-cli/src/validation.rs
```diff
@@ -34,15 +34,15 @@ pub fn validate(
             println!("Policyset failed to validate: {s}");
             Ok(())
         }
-        Err(e) => Err(e),
+        Err(e) => Err(e)?,
```
john-h-kastner-aws: `Err(e)?` looks funny to me. Should be able to put the `?` in the match like `lean_context.validate(policyset, schema, mode)?` and drop this arm. Same for the other changes in this file

cedar-lean-ffi/README.md
```diff
@@ -0,0 +1,59 @@
...
+Download and install Protobuf. The code has been tested with [Protobuf v29.3](https://github.com/protocolbuffers/protobuf/releases/tag/v29.3). If you are using Linux (on x86_64 arch) and wish to install protobuf to `/usr/local/` you can use the following instructions to install protobuf. For more information see the official [Protobuf installation page](https://github.com/protocolbuffers/protobuf#protobuf-compiler-installation).
+
+```
+curl -OL https://github.com/protocolbuffers/protobuf/releases/download/v29.3/protoc-29.3-linux-x86_64.zip
+unzip protoc-29.3-linux-x86_64.zip
+cp bin/protoc /usr/local/bin/
+cp -r include/google/ /usr/local/include/
+```
+
+### Build this CLI
```
john-h-kastner-aws:
```suggestion
### Build this Library
```

cedar-lean-ffi/build.rs
```diff
@@ -26,20 +48,23 @@ fn main() {
     if !Path::new(LEAN_BUILD_DIR).exists() {
         panic!("Lean build directory does not exist! Try running `( ./build_lean_lib.sh )`")
```
john-h-kastner-aws: Shouldn't need to run in a subshell since this isn't suggesting to `cd ../cedar-lean` anymore 

```suggestion
        panic!("Lean build directory does not exist! Try running `./build_lean_lib.sh`")
```

cedar-lean-ffi/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "cedar-lean-ffi"
+edition = "2021"
+version = "4.4.0"
+publish = false
+
+[dependencies]
+cedar-policy = { version = "4.4.0", features = ["protobufs"] }
```
john-h-kastner-aws, line 8: I like not always requiring a local clone, but what's the plan for local development and the Github actions build? As written, this will always pull from crates.io

chaluli: If I understand correctly, it won't pull from crates.io if the crate that imports this library uses a local clone of `cedar-policy`?

cedar-cli/src/validation.rs
chaluli: That's fair. The underlying err type changed and I was on auto-pilot to fix the errors. 😅

cedar-lean-ffi/Cargo.toml
john-h-kastner-aws: I guess the problem is if we add a new function to `cedar-policy` and want to call it in this crate. That will requiring depending on the local version for development and then the github version for CI until we do a new minor version release. Could be fine if you think this crate is small and stable enough for that not to come up.

john-h-kastner-aws: My biggest concern is that I want CI to be testing that this crate always build with the latest version of `cedar-policy` from `main`. If you think you have a way to set that up (or if building this into `cedar-drt` should do that automatically), then this isn't a problem

chaluli: I see two possible solutions.

1.) We add a `use-local` feature that will change the dependency to a local clone of cedar-policy.

by adding the following lines to `Cargo.toml`

```
[dependencies]
cedar-policy = { version = "4.4.0", features = ["protobufs"], optional = true }

[dependencies.cedar-policy-local]
package = "cedar-policy"
path = "../cedar/cedar-policy"
version = "4.4.0"
features = ["protobufs"]
optional = true

[features]
use-local = ["cedar-policy-local"]
default = ["cedar-policy"]
```

Normally, we build using crate.io (or whatever version the importing code uses). But allows us to do:

```
cargo build --features use-local --no-default-features
```

Note, as far as I can tell. I would need to modify every import of `cedar_policy::*` to use `cedar_policy_local::*` if the `use-local` feature is set.

2.) Require the CI or local developer to either manually edit the cargo or add a cargo patch that forces the use of a local clone of cedar-policy

```
# .cargo/config.toml
[patch.crates-io]
cedar-policy = { path = "../cedar/cedar-policy" }
```

Pros: No need to modify the actual Cargo.toml or rust code.
Cons: a bit clunkier to do during local development. Can be automated in CI by writing `.cargo/config.toml` in the CI actions as part of the setup.

cedar-cli/Cargo.toml
```diff
@@ -5,18 +5,14 @@ version = "0.1.0"
 publish = false
 
 [dependencies]
-cedar-policy = { path="../cedar/cedar-policy/", version = "=4.4.0", features = ["protobufs", "permissive-validate"] }
+cedar-policy = { version = "4.4.0", features = ["protobufs"] }
```
cdisselkoen, line 8: did we intentionally drop the `"=4.4.0"`?

cedar-cli/src/err.rs
```diff
@@ -97,8 +98,6 @@ pub enum ExecError {
         action_name: String,
         resource_type: String,
     },
-    #[error("Error deserializing Lean backend output")]
-    LeanDeserializationError,
-    #[error("Error occured in Lean backend : {0}")]
-    LeanBackendError(String),
+    #[error("{0}")]
```
cdisselkoen: This can be written as `#[error(transparent)]`

cedar-cli/src/evaluation.rs
```diff
@@ -70,9 +70,12 @@ pub fn evaluate(
                     }
                     Ok(())
                 }
-                Err(e) => Err(e),
+                Err(e) => Err(e)?,
```
cdisselkoen: Here's another place where we can move the `?` up, in this case to line 62 probably

cedar-lean-ffi/Cargo.toml
cdisselkoen: I think our CI for some repo in this org takes some variant of approach 2.  Probably makes sense to be consistent -- if we go with something else here, update our other CIs to match

cedar-lean-ffi/README.md
```diff
@@ -0,0 +1,59 @@
+# Cedar Lean FFI (Rust Bindings of Cedar Lean Formalization)
+
+This directory contains Rust bindings for interacting with the Lean formalization of Cedar. This FFI assumes all inputs for `Policy`s, `PolicySet`s, `Entities`, `Schema`s, `Request`s are represented using [`Cedar`](https://github.com/cedar-policy/cedar) public API types. This FFI makes use of Cedar's Protobuf feature to convert Cedar types in rust to the corresponding type within the Lean formalization.
```
cdisselkoen:
```suggestion
This directory contains Rust bindings for interacting with the Lean formalization of Cedar. This FFI assumes all inputs for `Policy`s, `PolicySet`s, `Entities`, `Schema`s, `Request`s are represented using [`Cedar`](https://github.com/cedar-policy/cedar) public API types. This FFI makes use of Cedar's Protobuf feature to convert Cedar types in Rust to the corresponding type within the Lean formalization.
```

cedar-lean-ffi/src/lean_ffi.rs
```diff
@@ -102,19 +110,102 @@ enum ResultDef<T> {
...
+    determining: HashSet<PolicyId>,
+    erroring: HashSet<PolicyId>,
+}
+
+impl AuthorizationResponse {
+    fn from_inner(inner: AuthorizationResponseInner) -> Result<Self, FfiError> {
+        let decision = match inner.decision.as_str() {
+            "allow" => Decision::Allow,
+            "deny" => Decision::Deny,
+            _ => return Err(FfiError::LeanDeserializationError),
```
cdisselkoen: Might be helpful to have this error carry the value that wasn't recognized

cedar-cli/Cargo.toml
chaluli: Yes. I am anticipating that the cli is compatible with any cedar-version 4.4.0 and up. That is currently true in that it works with 4.4.0 and 4.4.1.

cedar-lean-cli/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "cedar-lean-cli"
+edition = "2021"
+version = "0.1.0"
```
cdisselkoen, line 4: should this start at 4.4.0 as well?

cdisselkoen (approved):

PR #638, opened 2025-06-13
Adds examples and integration tests for cedar-lean-cli's analysis capabilities.

chaluli:
Adds examples to cedar-lean-cli for analysis. Uses examples to write integration tests. Add running integration tests to CI.

Updates analysis to output tabular results in a deterministic order (to make tests pass).

Updates cedar-lean-cli to change how vacuous policy findings are printed to make the text easier to understand. (applies to all requests instead of (permits) allows all / (forbids) denies all and applies to no requests instead of (permits) denies all / (forbids) allows all).


katherine-hough (approved):

cedar-lean-cli/src/util.rs
```diff
@@ -193,39 +191,31 @@ impl OpenRequestEnv {
...
-                                    );
-                                    req_envs.push(req_env)
+            if let Some(principal_types) = schema.principals_for_action(&action_id) {
+                for principal_type in principal_types {
+                    match &self.principal_type {
+                        Some(req_principal_type) if req_principal_type != principal_type => {
+                            continue
+                        }
+                        _ => (),
+                    };
```
adpaco, line 201:
```suggestion
if let Some(req_principal_type) = &self.principal_type && req_principal_type != principal_type {
    continue;
}
```

adpaco: Also, should we check and extract the value in `&self.principal_type` before going into this loop? Then you wouldn't need to have an `if let Some(...)` check in each iteration.

cedar-lean-cli/src/util.rs
```diff
@@ -193,39 +191,31 @@ impl OpenRequestEnv {
...
+                    if let Some(resource_types) = schema.resources_for_action(&action_id) {
+                        for resource_type in resource_types {
+                            match &self.resource_type {
+                                Some(req_resource_type) if req_resource_type != resource_type => {
+                                    continue
                                 }
```
adpaco, line 207: Approximately same comments apply here.

cedar-lean-cli/tests/integration_tests.rs
```diff
@@ -0,0 +1,932 @@
...
+fn test_analyze_policies_tabular_view_box_p1() {
+    let base_path = PathBuf::from("examples/analyze/view_box");
+
+    let output = Command::new("cedar-lean-cli")
+        .arg("analyze")
+        .arg("policies")
+        .arg(base_path.join("policies1.cedar"))
+        .arg(base_path.join("policies.cedarschema"))
+        .output()
+        .expect("Failed to run cedar-lean-cli");
```
adpaco, line 34: You could have a helper to build these commands so you only pass `policies1.cedar` and `policies.cedarschema` to it here, for example.

cedar-lean-cli/tests/integration_tests.rs
```diff
@@ -0,0 +1,932 @@
+use std::path::{Path, PathBuf};
+use std::process::{Command, Output};
+
+fn check_output<P: AsRef<Path>>(output: Output, expected_output_file: P, should_error: bool) {
```
adpaco, line 4: You don't have any test with `should_error` set to `true`, right?

adpaco-aws (approved):
Looks good, thanks!

I like output files to end with newlines and it looked like you were trimming in `check_output` so adding them shouldn't be an issue, right?

Issue #636, opened 2025-06-13
SMT encoding of strings


zhengyao-lin:
### Before opening, please confirm:

- [x] I have [searched for duplicate or closed issues](https://github.com/cedar-policy/cedar/issues?q=is%3Aissue+).
- [x] I have read the guide for [submitting bug reports](https://github.com/cedar-policy/cedar/blob/main/CONTRIBUTING.md#bug-reports).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### Bug Category

Other

### Describe the bug

There is a minor issue in the SMT encoding of strings in SymCC: https://github.com/cedar-policy/cedar-spec/blob/91b88c7a2414206e911a5c61ba21fdba2744dfe2/cedar-lean/Cedar/SymCC/Encoder.lean#L149
Per SMTLIB standard, a double quote character `"` in a string should be represented as `""`; so we should probably replace all `"` with `""` in `s` here
The [decoder](https://github.com/cedar-policy/cedar-spec/blob/91b88c7a2414206e911a5c61ba21fdba2744dfe2/cedar-lean/Cedar/SymCC/Decoder.lean#L67) seems to handle this correctly.

Otherwise this may cause an issue when encoding entity literal like `A::"\""`

### Expected behavior

Described above

### Reproduction steps

Described above

### Code Snippet

```rust
// Put your code below this line.

```

### Log output

```
// Put your output below this line

```

lianah:
This is indeed an issue in the SymCC encoding that we need to fix.

PR #640, opened 2025-06-16
Fix SMT encoding of string literals

zhengyao-lin:
*Issue #636*

*Description of changes:*

Escape character `"` correctly in `encodeString` (which is the only type of escape sequence per SMT-LIB [standard](https://smt-lib.org/papers/smt-lib-reference-v2.7-r2025-04-09.pdf)).

The added test would get stuck without this fix (not entirely sure why, but presumably due to CVC5 waiting more string contents since an invalid string literal `"""` is fed into it).

### DIFF
--- a/cedar-lean/Cedar/SymCC/Encoder.lean
+++ b/cedar-lean/Cedar/SymCC/Encoder.lean
@@ -147,7 +147,7 @@ https://smtlib.cs.uiowa.edu/theories-UnicodeStrings.shtml. Here,
 we're assuming ASCII strings for simplicity.
 -/
 def encodeString (s : String) : String :=
-  s!"\"{s}\""
+  s!"\"{s.replace "\"" "\"\""}\""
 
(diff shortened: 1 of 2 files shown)


emina (approved):

cedar-lean/Cedar/SymCC/Encoder.lean
```diff
@@ -147,7 +147,7 @@ https://smtlib.cs.uiowa.edu/theories-UnicodeStrings.shtml. Here,
 we're assuming ASCII strings for simplicity.
 -/
 def encodeString (s : String) : String :=
-  s!"\"{s}\""
+  s!"\"{s.replace "\"" "\"\""}\""
```
cdisselkoen, line 153: perhaps adding a comment here about `"` being the only character that needs escaping in SMTLib?

cdisselkoen (approved):

Issue #633, opened 2025-06-12
`cedar-lean-cli evaluate` errors even though `cedar authorize` allows, when attribute points to entity


luxas:
### Before opening, please confirm:

- [x] I have [searched for duplicate or closed issues](https://github.com/cedar-policy/cedar/issues?q=is%3Aissue+).
- [x] I have read the guide for [submitting bug reports](https://github.com/cedar-policy/cedar/blob/main/CONTRIBUTING.md#bug-reports).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### Bug Category

Policy Evaluation

### Describe the bug

`cedar-lean-cli evaluate` seems to have a hard time following entity references. Reproduction below, where following an entity reference errors, but authorizing a simple property works.

Cedarschema `foo.cedarschema`
```
entity User;
entity List = {
    org: Organization,
    name: String
};
entity Organization = {
    name: String
};
action get appliesTo {
    principal: User,
    resource: List
};
```
Policy `policy.cedar`
```
permit (principal == User::"id", action, resource)
when { resource.org.name == "acmeco" };

permit (principal == User::"alice", action, resource)
when { resource.name == "foo" };
```
Entities `repro.json`
```
[
    {
        "uid": {
            "type": "User",
            "id": "id"
        },
        "attrs": {},
        "parents": []
    },
    {
        "uid": {
            "type": "User",
            "id": "alice"
        },
        "attrs": {},
        "parents": []
    },
    {
        "uid": {
            "type": "List",
            "id": "id"
        },
        "attrs": {
            "org": {
                "type": "Organization",
                "id": "id"
            },
            "name": "foo"
        },
        "parents": []
    },
    {
        "uid": {
            "type": "Organization",
            "id": "id"
        },
        "attrs": {
            "name": "acmeco"
        },
        "parents": []
    }
]
```
`expr-ok`
```
resource.name == "foo"
```
`expr-error`
```
resource.org.name == "acmeco"
```

Output when trying to follow reference `resource.org.name`:
```console
$ cedar-lean-cli evaluate authorize --principal 'User::"id"' --action 'Action::"get"' --resource 'List::"id"' policy.cedar repro.json 
This request was implicitly denied as this request matched no policies
The following policies did not contribute to the decision as they errored during evaluation: policy0
$ cedar-lean-cli evaluate evaluate --principal 'User::"id"' --action 'Action::"get"' --resource 'List::"id"' expr-error repro.json
[...]

PR #645, opened 2025-06-24
Adds ability to parse entities using a provided schema during evaluation.

chaluli:
Addresses issue #633 which noticed the difference between entity parsing with and without schemas. This PR updates the cedar-lean-cli to allow a user to provide a schema while parsing entities.

### DIFF
--- a/cedar-lean-cli/src/evaluation.rs
+++ b/cedar-lean-cli/src/evaluation.rs
@@ -40,11 +40,14 @@ pub fn check_is_authorized(
         print!(" {policy}");
     }
     println!();
-    print!("The following policies did not contribute to the decision as they errored during evaluation:");
-    for policy in auth_response.erroring_policies() {
-        print!(" {policy}");
+    if !auth_response.erroring_policies().is_empty() {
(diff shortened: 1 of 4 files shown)


cedar-lean-cli/src/evaluation.rs
```diff
@@ -40,11 +40,14 @@ pub fn check_is_authorized(
...
-    print!("The following policies did not contribute to the decision as they errored during evaluation:");
-    for policy in auth_response.erroring_policies() {
-        print!(" {policy}");
+    if !auth_response.erroring_policies().is_empty() {
+        println!();
+        print!("The following policies did not contribute to the decision as they errored during evaluation:");
+        for policy in auth_response.erroring_policies() {
+            print!(" {policy}");
+        }
+        println!();
```
cdisselkoen: alternately:
```suggestion
        println!("The following policies did not contribute to the decision as they errored during evaluation: {}", auth_response.erroring_policies().join(" ");
```
requires `use itertools::Itertools`

PR #646, opened 2025-06-24
Prove that `compile` never fails on well-typed expressions

zhengyao-lin:
This PR adds a theorem stating that `compile` never fails and results in a `Term` with the corresponding type when the input is a `TypedExpr`.
This was not verified before, and there was in fact an issue causing SymCC to fail on validated policies.

The exact statement is
```
theorem compile_well_typed {Γ : Environment} {εnv : SymEnv} {tx : TypedExpr} :
  εnv = SymEnv.ofEnv Γ →
  TypedExpr.WellTyped Γ tx →
  εnv.WellFormedFor tx.toExpr →
  ∃ t : Term,
    compile tx.toExpr εnv = .ok t ∧
    t.typeOf = .option (TermType.ofType tx.typeOf)
```

Ideally `εnv.WellFormedFor tx.toExpr` should also be derived from `εnv = SymEnv.ofEnv Γ` and `TypedExpr.WellTyped Γ tx`, but I'm hoping to merge this version first and get some feedback.

Thanks in advance for any suggestions! and please be critical to my Lean proof style since I'm new to Lean :)

### DIFF
--- a/cedar-lean/Cedar/Thm/Data/List/Canonical.lean
+++ b/cedar-lean/Cedar/Thm/Data/List/Canonical.lean
@@ -66,22 +66,20 @@ theorem insertCanonical_new
   y ∈ List.insertCanonical f x xs
 := by
   induction xs
-
   case nil => simp at hmem
-
   case cons hd tl ih =>
     simp only [List.insertCanonical]
     split
-    simp [hmem]
+    simp only [mem_cons, hmem, or_true]
     split
-    · simp at hmem
+    · simp only [mem_cons] at hmem
       cases hmem
       case _ hy => simp [hy]
       case _ hy => simp [ih hy]
-    · simp at hmem
+    · simp only [mem_cons] at hmem
       cases hmem
       case _ hlt hgt hy =>
-        simp [← hy] at hlt hgt
+        simp only [← hy, gt_iff_lt] at hlt hgt
         have hltgt := StrictLT.connected (f x) (f y) hneq
         cases hltgt <;> contradiction
       case _ hy => simp [hy]
--- a/cedar-lean/Cedar/Thm/Data/List/Canonical.lean
+++ b/cedar-lean/Cedar/Thm/Data/List/Canonical.lean
@@ -46,7 +46,7 @@ theorem insertCanonical_not_nil [DecidableEq β] [LT β] [DecidableLT β] (f : �
     split at h <;> try trivial
     split at h <;> trivial
 
-theorem mem_insertCanonical [LT β] [Cedar.Data.StrictLT β] [DecidableLT β]
+theorem insertCanonical_mem [LT β] [Cedar.Data.StrictLT β] [DecidableLT β]
   {f : α → β} {xs : List α} (x : α) :
   x ∈ List.insertCanonical f x xs
 := by
@@ -58,7 +58,7 @@ theorem mem_insertCanonical [LT β] [Cedar.Data.StrictLT β] [DecidableLT β]
     split; any_goals simp
     apply Or.inr ih
 
-theorem insertCanonical_new
+theorem insertCanonical_preserves_non_duplicate_element
   [LT β] [Cedar.Data.StrictLT β] [DecidableLT β] [DecidableEq β]
   {f : α → β} {xs : List α} {x : α} {y : α}
   (hmem : y ∈ xs)
--- a/cedar-lean/Cedar/Thm/Data/List/Canonical.lean
+++ b/cedar-lean/Cedar/Thm/Data/List/Canonical.lean
@@ -46,14 +46,14 @@ theorem insertCanonical_not_nil [DecidableEq β] [LT β] [DecidableLT β] (f : �
     split at h <;> try trivial
     split at h <;> trivial
 
-theorem insertCanonical_exists [LT β] [Cedar.Data.StrictLT β] [DecidableLT β]
+theorem mem_insertCanonical [LT β] [Cedar.Data.StrictLT β] [DecidableLT β]
   {f : α → β} {xs : List α} (x : α) :
   x ∈ List.insertCanonical f x xs
 := by
   induction xs
   case nil => simp [List.insertCanonical]
   case cons head tail ih =>
-    simp [List.insertCanonical]
+    simp only [List.insertCanonical]
     split; any_goals simp
     split; any_goals simp
     apply Or.inr ih
@@ -61,25 +61,25 @@ theorem insertCanonical_exists [LT β] [Cedar.Data.StrictLT β] [DecidableLT β]
 theorem insertCanonical_new
   [LT β] [Cedar.Data.StrictLT β] [DecidableLT β] [DecidableEq β]
   {f : α → β} {xs : List α} {x : α} {y : α}
-  (hexists : y ∈ xs)
+  (hmem : y ∈ xs)
   (hneq : f x ≠ f y) :
   y ∈ List.insertCanonical f x xs
 := by
   induction xs
 
-  case nil => simp at hexists
+  case nil => simp at hmem
 
   case cons hd tl ih =>
-    simp [List.insertCanonical]
+    simp only [List.insertCanonical]
     split
-    simp [hexists]
+    simp [hmem]
     split
-    · simp at hexists
-      cases hexists
+    · simp at hmem
+      cases hmem
       case _ hy => simp [hy]
       case _ hy => simp [ih hy]
-    · simp at hexists
-      cases hexists
+    · simp at hmem
+      cases hmem
       case _ hlt hgt hy =>
         simp [← hy] at hlt hgt
         have hltgt := StrictLT.connected (f x) (f y) hneq
--- a/cedar-lean/Cedar/Thm/Data/List/Canonical.lean
+++ b/cedar-lean/Cedar/Thm/Data/List/Canonical.lean
@@ -46,6 +46,46 @@ theorem insertCanonical_not_nil [DecidableEq β] [LT β] [DecidableLT β] (f : �
     split at h <;> try trivial
     split at h <;> trivial
 
+theorem insertCanonical_exists [LT β] [Cedar.Data.StrictLT β] [DecidableLT β]
+  {f : α → β} {xs : List α} (x : α) :
+  x ∈ List.insertCanonical f x xs
+:= by
+  induction xs
+  case nil => simp [List.insertCanonical]
+  case cons head tail ih =>
+    simp [List.insertCanonical]
+    split; any_goals simp
+    split; any_goals simp
+    apply Or.inr ih
+
+theorem insertCanonical_new
+  [LT β] [Cedar.Data.StrictLT β] [DecidableLT β] [DecidableEq β]
+  {f : α → β} {xs : List α} {x : α} {y : α}
+  (hexists : y ∈ xs)
+  (hneq : f x ≠ f y) :
+  y ∈ List.insertCanonical f x xs
+:= by
+  induction xs
+
+  case nil => simp at hexists
+
+  case cons hd tl ih =>
+    simp [List.insertCanonical]
+    split
+    simp [hexists]
+    split
+    · simp at hexists
+      cases hexists
+      case _ hy => simp [hy]
+      case _ hy => simp [ih hy]
+    · simp at hexists
+      cases hexists
+      case _ hlt hgt hy =>
+        simp [← hy] at hlt hgt
+        have hltgt := StrictLT.connected (f x) (f y) hneq
+        cases hltgt <;> contradiction
+      case _ hy => simp [hy]
+
 theorem insertCanonical_sortedBy [LT β] [StrictLT β] [DecidableLT β] {f : α → β} {xs : List α} (x : α) :
   SortedBy f xs →
   SortedBy f (insertCanonical f x xs)
--- a/cedar-lean/Cedar/Thm/Data/List/Lemmas.lean
+++ b/cedar-lean/Cedar/Thm/Data/List/Lemmas.lean
@@ -920,16 +920,6 @@ theorem forall₂_implies_mapM_eq {α₁ α₂ β ε} {xs : List α₁} {ys : Li
   case cons h₁ h₂ =>
     simp only [List.mapM_cons, h₁, forall₂_implies_mapM_eq f g h₂, bind_pure_comp]
 
-theorem forall₂_implies_map_eq {α₁ α₂ β} {xs : List α₁} {ys : List α₂} (f : α₁ → β) (g : α₂ → β):
-  List.Forall₂ (fun x y => f x = g y) xs ys →
-  List.map f xs = List.map g ys
-:= by
-  intro h
-  cases h
-  case nil => simp only [List.map_nil]
-  case cons h₁ h₂ =>
-    simp only [List.map_cons, h₁, forall₂_implies_map_eq f g h₂, bind_pure_comp]
-
 /--
   our own variant of map_congr, for mapM'
 -/
--- a/cedar-lean/Cedar/Thm/Data/List/Lemmas.lean
+++ b/cedar-lean/Cedar/Thm/Data/List/Lemmas.lean
@@ -1074,6 +1074,36 @@ theorem find?_fst_map_implies_find? {α β γ} [BEq α] {f : β → γ} {xs : Li
       simp only [Prod.map, id_eq] at heq
       simp [find?_cons, heq, ih]
 
+theorem find?_implies_find?_fst_map
+  {α β γ} [BEq α] [ReflBEq α]
+  {l : List (α × β)}
+  {k : α} {v : β}
+  {f : α → β → γ}
+  (h : List.find? (λ x => x.fst == k) l = some (k, v)) :
+  List.find? (λ x => x.fst == k) (l.map λ (k, v) => (k, f k v)) = some (k, f k v)
+:= by
+  induction l
+  case nil => simp at h
+  case cons head tail ih =>
+    simp at h
+    cases h
+    case _ h => simp [h]
+    case _ h =>
+      have ih := ih h.2
+      simp only [List.map]
+      simp only [List.find?]
+      simp [ih]
+      simp [h]
+
+theorem find?_implies_append_find?
+  {a b : List α}
+  {v : α}
+  {f : α → Bool}
+  (h : List.find? f a = some v) :
+  List.find? f (a ++ b) = some v
+:= by
+  simp [List.find?_append, h]
+
 theorem not_find?_some_iff_find?_none {α} {p : α → Bool} {xs : List α} :
   (∀ x ∈ xs, ¬xs.find? p = .some x) ↔ xs.find? p = .none
 := by
--- a/cedar-lean/Cedar/Thm/Data/List/Lemmas.lean
+++ b/cedar-lean/Cedar/Thm/Data/List/Lemmas.lean
@@ -382,7 +382,7 @@ theorem mapM₁_eq_mapM [Monad m] [LawfulMonad m]
 := by
   simp only [mapM₁, attach_def, mapM_pmap_subtype]
 
-theorem mapM₂_eq_mapM [Monad m] [LawfulMonad m]
+theorem mapM₂_eq_mapM [Monad m] [LawfulMonad m] [SizeOf α] [SizeOf β]
   (f : (α × β) → m γ)
   (as : List (α × β)) :
   List.mapM₂ as (λ x : { x // sizeOf x.snd < 1 + sizeOf as } => f x.val) =
--- a/cedar-lean/Cedar/Thm/Data/List/Lemmas.lean
+++ b/cedar-lean/Cedar/Thm/Data/List/Lemmas.lean
@@ -1089,11 +1089,10 @@ theorem find?_implies_find?_fst_map
     cases h
     case _ h => simp [h]
     case _ h =>
-      have ih := ih h.2
+      specialize ih h.2
       simp only [List.map]
       simp only [List.find?]
-      simp [ih]
-      simp [h]
+      simp [ih, h]
 
 theorem find?_implies_append_find?
   {a b : List α}
--- a/cedar-lean/Cedar/Thm/Data/List/Lemmas.lean
+++ b/cedar-lean/Cedar/Thm/Data/List/Lemmas.lean
@@ -501,27 +501,24 @@ theorem mapM_implies_forall₂
 := by
   induction xs generalizing ys
   case nil =>
-    simp [mapM, mapM.loop, pure, Except.pure] at hmapM
+    simp only [mapM, mapM.loop, pure, Except.pure, reverse_nil, Except.ok.injEq, nil_eq] at hmapM
     simp [hmapM]
   case cons xhd xtl ih =>
-    simp [
-      pure, Except.pure,
-      Bind.bind, Except.bind,
-    ] at hmapM
+    simp only [mapM_cons, Bind.bind, Except.bind, pure, Except.pure] at hmapM
     split at hmapM
     contradiction
     split at hmapM
     contradiction
-    simp at hmapM
-    simp [← hmapM]
+    simp only [Except.ok.injEq] at hmapM
+    simp only [← hmapM, forall₂_cons]
     constructor
     · apply h
-      simp
+      simp only [mem_cons, true_or]
       assumption
     · apply ih
       intros
       apply h
-      simp [*]
+      simp only [mem_cons, or_true, *]
       assumption
       assumption
 
@@ -1163,9 +1160,9 @@ theorem find?_implies_find?_fst_map
   List.find? (λ x => x.fst == k) (l.map λ (k, v) => (k, f k v)) = some (k, f k v)
 := by
   induction l
-  case nil => simp at h
+  case nil => simp only [find?_nil, reduceCtorEq] at h
   case cons head tail ih =>
-    simp at h
+    simp only [find?_cons_eq_some, Bool.not_eq_eq_eq_not, Bool.not_true] at h
     cases h
     case _ h => simp [h]
     case _ h =>
--- a/cedar-lean/Cedar/Thm/Data/List/Lemmas.lean
+++ b/cedar-lean/Cedar/Thm/Data/List/Lemmas.lean
@@ -491,7 +491,7 @@ of a `f` through `List.mapM`. This is slightly stronger
 than the forward direction of `mapM_ok_iff_forall₂`
 since it allowed an extra `x ∈ xs` condition in `h`.
 -/
-theorem mapM_implies_attched_forall₂
+theorem mapM_implies_forall₂
   {f : α → Except ε β}
   {p : α → β → Prop}
   {xs : List α} {ys : List β}
--- a/cedar-lean/Cedar/Thm/Data/List/Lemmas.lean
+++ b/cedar-lean/Cedar/Thm/Data/List/Lemmas.lean
@@ -295,7 +295,6 @@ theorem forall₂_fun_subset_implies {R : α → β → Prop} {xs xs' : List α}
   subst hf
   exact hy'
 
-
 /--
 The converse (ys ≡ ys' → xs ≡ xs') doesn't hold without requiring R to be
 injective (as well as functional).
@@ -313,6 +312,37 @@ theorem forall₂_fun_equiv_implies {R : α → β → Prop} {xs xs' : List α} 
   constructor
   · exact forall₂_fun_subset_implies ha ha' hf heqv
   · exact forall₂_fun_subset_implies ha' ha hf heqv'
+
+theorem map_preserves_forall₂
+  {f : α → α'}
+  {g : β → β'}
+  {p₁ : α → β → Prop}
+  {p₂ : α' → β' → Prop}
+  {xs : List α} {ys : List β}
+  (h : ∀ x y, p₁ x y → p₂ (f x) (g y))
+  (hforall₂ : List.Forall₂ p₁ xs ys) :
+  List.Forall₂ p₂ (xs.map f) (ys.map g)
+:= by
+  induction hforall₂
+  case nil => simp
+  case cons =>
+    simp only [List.map]
+    constructor
+    apply h
+    assumption
+    assumption
+
+theorem forall₂_swap
+  {R : α → β → Prop} {xs : List α} {ys : List β}
+  (hforall₂ : List.Forall₂ (λ y x => R x y) ys xs) :
+  List.Forall₂ R xs ys
+:= by
+  induction hforall₂
+  case nil => simp
+  case cons =>
+    constructor
+    assumption
+    assumption
 
 /-! ### mapM, mapM', mapM₁, and mapM₂ -/
 
@@ -454,6 +484,46 @@ theorem mapM_ok_iff_forall₂ {α β γ} {f : α → Except γ β} {xs : List α
 := by
   rw [← List.mapM'_eq_mapM]
   exact mapM'_ok_iff_forall₂
+
+/--
+Introduces `forall₂` through the input output relation
+of a `f` through `List.mapM`. This is slightly stronger
+than the forward direction of `mapM_ok_iff_forall₂`
+since it allowed an extra `x ∈ xs` condition in `h`.
+-/
+theorem mapM_implies_attched_forall₂
+  {f : α → Except ε β}
+  {p : α → β → Prop}
+  {xs : List α} {ys : List β}
+  (h : ∀ x y, x ∈ xs → f x = .ok y → p x y)
+  (hmapM : List.mapM f xs = .ok ys) :
+  List.Forall₂ p xs ys
+:= by
+  induction xs generalizing ys
+  case nil =>
+    simp [mapM, mapM.loop, pure, Except.pure] at hmapM
+    simp [hmapM]
+  case cons xhd xtl ih =>
+    simp [
+      pure, Except.pure,
+      Bind.bind, Except.bind,
+    ] at hmapM
+    split at hmapM
+    contradiction
+    split at hmapM
+    contradiction
+    simp at hmapM
+    simp [← hmapM]
+    constructor
+    · apply h
+      simp
+      assumption
+    · apply ih
+      intros
+      apply h
+      simp [*]
+      assumption
+      assumption
 
 /--
   Note that the converse is not true:
@@ -852,6 +922,16 @@ theorem forall₂_implies_mapM_eq {α₁ α₂ β ε} {xs : List α₁} {ys : Li
   case nil => simp only [List.mapM_nil]
   case cons h₁ h₂ =>
     simp only [List.mapM_cons, h₁, forall₂_implies_mapM_eq f g h₂, bind_pure_comp]
+
+theorem forall₂_implies_map_eq {α₁ α₂ β} {xs : List α₁} {ys : List α₂} (f : α₁ → β) (g : α₂ → β):
+  List.Forall₂ (fun x y => f x = g y) xs ys →
+  List.map f xs = List.map g ys
+:= by
+  intro h
+  cases h
+  case nil => simp only [List.map_nil]
+  case cons h₁ h₂ =>
+    simp only [List.map_cons, h₁, forall₂_implies_map_eq f g h₂, bind_pure_comp]
 
 /--
   our own variant of map_congr, for mapM'
--- a/cedar-lean/Cedar/Thm/Data/Map.lean
+++ b/cedar-lean/Cedar/Thm/Data/Map.lean
@@ -403,6 +403,41 @@ theorem find?_none_iff_all_absent [LT α] [DecidableLT α] [StrictLT α] [Decida
   m.find? k = none ↔ ∀ v, ¬ (k, v) ∈ m.kvs
 := Iff.intro find?_none_all_absent all_absent_find?_none
 
+/--
+  If the key-value pair exists in l, then it exists in `Data.Map.make l`
+-/
+theorem find?_implies_make_find?
+  [DecidableEq α] [LT α] [DecidableLT α]
+  [Cedar.Data.StrictLT α]
+  {l : List (α × β)}
+  {k : α} {v : β}
+  (h : List.find? (λ x => x.fst == k) l = some (k, v)) :
+  (Data.Map.make l).find? k = some v
+:= by
+  apply (in_list_iff_find?_some (Data.Map.make_wf l)).mp
+  simp [make, Data.Map.kvs]
+
+  induction l
+
+  case nil => simp at h
+  case cons head tail ih =>
+    simp [List.find?] at h
+    split at h
+
+    case _ heq =>
+      simp at h
+      simp [List.canonicalize, h]
+      apply List.insertCanonical_exists
+
+    case _ hneq =>
+      have ih := ih h
+      simp at hneq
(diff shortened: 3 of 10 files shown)


cedar-lean/Cedar/Thm/Data/List/Canonical.lean
```diff
@@ -46,6 +46,46 @@ theorem insertCanonical_not_nil [DecidableEq β] [LT β] [DecidableLT β] (f : 
     split at h <;> try trivial
     split at h <;> trivial
 
+theorem insertCanonical_exists [LT β] [Cedar.Data.StrictLT β] [DecidableLT β]
```
cdisselkoen: maybe this should be called `mem_insertCanonical`? @emina what do you think, I think you're more familiar than me with Lean lemma naming conventions?

cedar-lean/Cedar/Thm/Data/List/Lemmas.lean
```diff
@@ -1074,6 +1074,36 @@ theorem find?_fst_map_implies_find? {α β γ} [BEq α] {f : β → γ} {xs : Li
...
+  List.find? (λ x => x.fst == k) (l.map λ (k, v) => (k, f k v)) = some (k, f k v)
+:= by
+  induction l
+  case nil => simp at h
+  case cons head tail ih =>
+    simp at h
+    cases h
+    case _ h => simp [h]
+    case _ h =>
+      have ih := ih h.2
```
cdisselkoen: This can be written `specialize ih h.2` :)

cedar-lean/Cedar/Thm/Data/List/Lemmas.lean
```diff
@@ -1074,6 +1074,36 @@ theorem find?_fst_map_implies_find? {α β γ} [BEq α] {f : β → γ} {xs : Li
...
+      simp only [List.map]
+      simp only [List.find?]
+      simp [ih]
+      simp [h]
```
cdisselkoen: can this be combined into a single `simp [ih, h]`?

cedar-lean/Cedar/Thm/SymCC/Compiler/WellTyped.lean
```diff
@@ -0,0 +1,2231 @@
...
+/--
+A wrapper around compile_wf for convenience
+-/
+theorem wt_cond_implies_compile_wf
+  {tx : TypedExpr} {Γ : Environment} {εnv : SymEnv} {t : Term} :
+  CompileWellTypedCondition tx Γ εnv →
+  compile tx.toExpr εnv = .ok t →
+  t.WellFormed εnv.entities
+:= by
+  intros h hcomp; rcases h with ⟨_, _, _, hwf⟩
```
cdisselkoen: When `h` is just a big conjunction, we'd typically use `have ⟨_, _, _, hwf⟩ := h` rather than `rcases`

cedar-lean/Cedar/Thm/SymCC/Compiler/WellTyped.lean
```diff
@@ -0,0 +1,2231 @@
...
+  have htwf := compile_wf hwf hcomp
+  simp [htwf]
```
cdisselkoen: this could be written `simp [compile_wf hwf hcomp]`

cedar-lean/Cedar/Thm/SymCC/Compiler/WellTyped.lean
```diff
@@ -0,0 +1,2231 @@
...
+
+  any_goals simp
+  any_goals contradiction
+  all_goals simp [TermType.cedarType?] at *
+
+  case prim _ h _ _ =>
+    unfold TermType.cedarType? at *
+    split at h
+    any_goals simp at h
+    any_goals contradiction
```
cdisselkoen: here and in some other spots, I think it would be preferable to use `·` or explicit case names rather than `any_goals`, particularly if the tactic you're using only applies to one goal anyway.  If something changes in our definitions, it will be easier to figure out what went wrong if the proof contains an explicit record of which tactic was "supposed" to apply to which goal.

cedar-lean/Cedar/Thm/SymCC/Compiler/WellTyped.lean
```diff
@@ -0,0 +1,2231 @@
...
+
+  case set ty =>
+    cases e : ty.cedarType?
+    all_goals rw [e] at hty
+    all_goals simp at hty
```
cdisselkoen: can this be combined into `all_goals simp [e] at hty`?

cedar-lean/Cedar/Thm/SymCC/Compiler/WellTyped.lean
```diff
@@ -0,0 +1,2231 @@
...
+theorem compile_well_typed_var {v : Var} {ty : CedarType} {Γ : Environment} {εnv : SymEnv} :
+  CompileWellTypedCondition (.var v ty) Γ εnv →
+  CompileWellTypedForExpr (.var v ty) εnv
+:= by
+  intros h; rcases h with ⟨_, hεnv, hwt, hwf⟩
+  simp [TypedExpr.typeOf, TypedExpr.toExpr, CompileWellTypedForExpr] at *
+
+  unfold SymEnv.WellFormedFor SymEnv.WellFormed SymRequest.WellFormed at hwf
+
+  rcases hwf with ⟨⟨⟨_, hprincipal, _, haction, _, hresource, _, hcontext⟩, _⟩, _⟩
```
cdisselkoen: here's another case where I'd use `have` instead of `rcases`

cedar-lean/Cedar/Thm/SymCC/Compiler/WellTyped.lean
```diff
@@ -0,0 +1,2231 @@
...
+  case _ =>
```
cdisselkoen: is this necessary?

cedar-lean/Cedar/Thm/SymCC/Compiler/WellTyped.lean
```diff
@@ -0,0 +1,2231 @@
...
+  (ihb : CompileWellTypedForExpr b εnv) :
+
+  (CompileWellTypedCondition (.or a b ty) Γ εnv →
+    CompileWellTypedForExpr (.or a b ty) εnv) ∧
+
+  (CompileWellTypedCondition (.and a b ty) Γ εnv →
+    CompileWellTypedForExpr (.and a b ty) εnv)
+:= by
+  constructor
+  all_goals
```
cdisselkoen, line 287: I actually don't know what Lean's rules are for where the scope of this `all_goals` ends. Can we use indentation? or `{ }`?

cedar-lean/Cedar/Thm/SymCC/Compiler/WellTyped.lean
```diff
@@ -0,0 +1,2231 @@
...
+  simp [h] at hfound; clear h
+
+  apply Data.Map.find?_implies_make_find?
+  apply List.find?_implies_append_find?
+  apply List.find?_implies_find?_fst_map
+  assumption
+
+/--
+Lemma that if a concrete `Γ : Environment` has tags for
+a particular entity type, when `SymEnv.ofEnv Γ` must also
```
cdisselkoen:
```suggestion
a particular entity type, then `SymEnv.ofEnv Γ` must also
```

cedar-lean/Cedar/Thm/SymCC/Compiler/WellTyped.lean
```diff
@@ -0,0 +1,2231 @@
...
+have tags for it
+-/
+theorem SymEnv_of_preserves_tags
+  {Γ : Environment} {ety : EntityType} {ty : CedarType}
+  (h : Γ.ets.tags? ety = some (some ty)) :
+  ∃ τags : SymTags,
+    (SymEnv.ofEnv Γ).entities.tags ety = τags ∧
+    τags.vals.outType = TermType.ofType ty
```
cdisselkoen: Does this theorem need to be stated as an existential, or could we instead have `τags` as an implicit parameter and `(SymEnv.ofEnv Γ).entities.tags ety = τags` be a hypothesis?  The theorem would be stronger if stated that way, but maybe it wouldn't hold?

cedar-lean/Cedar/Thm/SymCC/Compiler/WellTyped.lean
```diff
@@ -0,0 +1,2231 @@
...
+        simp
+
+-- Given that
+--   Γ.ets.attrs? ety = some a
+--   a.liftBoolTypes = rty
+-- Show that
+--   (SymEnv.ofEnv Γ).entities.attrs ety = .some attrs
+--   UnaryFunction.WellFormed εnv.entities attrs
+--   attrs.argType = CedarType.entity ety
+--   attrs.outType = .record (Data.Map.mk (TermType.ofRecordType rty.1))
```
cdisselkoen: Comment can probably be removed now

cedar-lean/Cedar/Thm/SymCC/Compiler/WellTyped.lean
```diff
@@ -0,0 +1,2231 @@
...
+    simp [Data.Map.find?, hfound]
+
+  have ⟨attrs, hattrs_exists2⟩ :
+    ∃ attrs : UnaryFunction, εnv.entities.attrs ety = .some attrs
+  := by
+    simp [SymEntities.attrs, hety_exists]
+
+  have ⟨hwf_attrs, hty_arg_attrs⟩ := env_wf_implies_attrs_wf hwf hattrs_exists2
+
+  apply Exists.intro attrs
```
cdisselkoen: I think you can write this as `exists attrs`

cedar-lean/Cedar/Thm/SymCC/Compiler/WellTyped.lean
```diff
@@ -0,0 +1,2231 @@
...
+        apply Exists.intro x
+        simp [hx]
+
+theorem compile_well_typed_set
+  {xs : List TypedExpr} {ty : CedarType}
+  {Γ : Environment} {εnv : SymEnv}
+  (ihxs : ∀ x, x ∈ xs → CompileWellTypedForExpr x εnv)
+  (hcond : CompileWellTypedCondition (.set xs ty) Γ εnv) :
+  CompileWellTypedForExpr (.set xs ty) εnv
+:= by
```
cdisselkoen, line 1112: This proof is long, maybe we could pull some parts out into helper lemmas?

cdisselkoen (approved):
Very nice result, only minor comments

cedar-lean/Cedar/Thm/SymCC/Compiler/WellTyped.lean
zhengyao-lin: I think that would still hold. Although in the way it's used, we also need the fact that `Γ.ets.tags? ety = some (some ty)` implies `(SymEnv.ofEnv Γ).entities.tags ety = some (some τags)` for some `τags`; but I guess I can prove that in a separate theorem.

cedar-lean/Cedar/Thm/Data/List/Canonical.lean
emina: Yes, this should be called `mem_insertCanonical`.

cedar-lean/Cedar/Thm/Data/List/Canonical.lean
```diff
@@ -46,6 +46,46 @@ theorem insertCanonical_not_nil [DecidableEq β] [LT β] [DecidableLT β] (f : 
...
+  {f : α → β} {xs : List α} (x : α) :
+  x ∈ List.insertCanonical f x xs
+:= by
+  induction xs
+  case nil => simp [List.insertCanonical]
+  case cons head tail ih =>
+    simp [List.insertCanonical]
```
emina: In this proof and all other proofs, let's not use `simp [...]` _unless_ it closes the goal.  If `simp` is not closing a goal (like here), we should write `simp only [...]` instead, which specifies the exact lemmas needed to proceed.  

You can ask Lean to tell you which lemmas it used for a `simp` statement by replacing it with `simp?`.  The IDE will then suggest a replacement expressed in terms of `simp only`.

This is important to ensure stability of proofs across Lean upgrades.  Proofs that use plain `simp` are more likely to break (as the updates can change the default set of available `simp` lemmas).

cedar-lean/Cedar/Thm/Data/List/Canonical.lean
```diff
@@ -46,6 +46,46 @@ theorem insertCanonical_not_nil [DecidableEq β] [LT β] [DecidableLT β] (f : 
...
+    split; any_goals simp
+    split; any_goals simp
+    apply Or.inr ih
+
+theorem insertCanonical_new
+  [LT β] [Cedar.Data.StrictLT β] [DecidableLT β] [DecidableEq β]
+  {f : α → β} {xs : List α} {x : α} {y : α}
+  (hexists : y ∈ xs)
```
emina: Let's replace `hexists` with `hmem` for consistency.

cedar-lean/Cedar/Thm/Data/List/Canonical.lean
```diff
@@ -46,6 +46,46 @@ theorem insertCanonical_not_nil [DecidableEq β] [LT β] [DecidableLT β] (f : 
...
+  (hneq : f x ≠ f y) :
+  y ∈ List.insertCanonical f x xs
+:= by
+  induction xs
+
+  case nil => simp at hexists
+
+  case cons hd tl ih =>
+    simp [List.insertCanonical]
```
emina: Let's replace plain `simp [ ... ]` with `simp [only]` (see above comment).

cedar-lean/Cedar/Thm/Data/Map.lean
```diff
@@ -403,6 +403,41 @@ theorem find?_none_iff_all_absent [LT α] [DecidableLT α] [StrictLT α] [Decida
...
+  [DecidableEq α] [LT α] [DecidableLT α]
+  [Cedar.Data.StrictLT α]
+  {l : List (α × β)}
+  {k : α} {v : β}
+  (h : List.find? (λ x => x.fst == k) l = some (k, v)) :
+  (Data.Map.make l).find? k = some v
+:= by
+  apply (in_list_iff_find?_some (Data.Map.make_wf l)).mp
+  simp [make, Data.Map.kvs]
+
```
emina: I suggest deleting the extra newlines inside of proofs.

cedar-lean/Cedar/Thm/Data/Map.lean
```diff
@@ -403,6 +403,41 @@ theorem find?_none_iff_all_absent [LT α] [DecidableLT α] [StrictLT α] [Decida
...
+  induction l
+
+  case nil => simp at h
+  case cons head tail ih =>
+    simp [List.find?] at h
+    split at h
+
+    case _ heq =>
+      simp at h
```
emina: See earlier comments about `simp` vs `simp only`.

cedar-lean/Cedar/Thm/SymCC/Compiler.lean
```diff
@@ -24,6 +24,7 @@ import Cedar.Thm.SymCC.Compiler.LitVar
 import Cedar.Thm.SymCC.Compiler.Record
 import Cedar.Thm.SymCC.Compiler.Set
 import Cedar.Thm.SymCC.Compiler.Unary
+import Cedar.Thm.SymCC.Compiler.WellTyped
```
emina: Why is this imported needed?  

Nothing in this file depends on the WellTyped lemmas, so we should remove this import.

cedar-lean/Cedar/Thm/SymCC/Compiler/WellTyped.lean
```diff
@@ -0,0 +1,2209 @@
...
+/--
+States the symbolic compiler succeeds on a typed expression `tx` and
+produces a term `t` with the corresponding type.
+-/
+def CompileWellTypedForExpr (tx : TypedExpr) (εnv : SymEnv) : Prop :=
+  ∃ t : Term,
+    compile tx.toExpr εnv = .ok t ∧
+    t.typeOf = .option (TermType.ofType tx.typeOf)
+
+/--
```
emina, line 27: Let's remove the placeholder.  

I would suggest not modifying this proof later. 

Instead, prove the following theorem separately:

```
theorem wf_type_env_implies_ofEnv_wf {Γ : Environment} {tx : TypedExpr }: 
  Γ.WellFormed → 
  tx.WellTyped Γ → 
  (SymEnv.ofEnv Γ).WellFormedFor tx.toExpr
```

Then, combine the main theorem in this file and `wf_type_env_implies_ofEnv_wf` to obtain a top-level theorem, directly in the `Cedar.Thm` folder that looks like this:


```
theorem compile_well_typed_ok {Γ : Environment} {tx : TypedExpr }: 
  Γ.WellFormed → 
  tx.WellTyped Γ → 
  ∃ t : Term,
    compile tx.toExpr (SymEnv.of Γ) = .ok t ∧
    t.typeOf = .option (TermType.ofType tx.typeOf)
```

cedar-lean/Cedar/Thm/SymCC/Compiler/WellTyped.lean
```diff
@@ -0,0 +1,2209 @@
...
+-/
+theorem compile_well_typed_lit {p : Prim} {tx : CedarType} {Γ : Environment} {εnv : SymEnv} :
+  CompileWellTypedCondition (.lit p tx) Γ εnv →
+  CompileWellTypedForExpr (.lit p tx) εnv
+:= by
+  intros h; rcases h with ⟨_, hεnv, hwt, hwf⟩
+  simp [TypedExpr.typeOf, TypedExpr.toExpr, CompileWellTypedForExpr] at *
+
+  unfold SymEnv.WellFormedFor at hwf
+  rcases hwf with ⟨_, ⟨hrefs⟩⟩
```
emina: Let's replace all occurrences of `rcases` with `have` unless `rcases` is being using to decompose a disjunction.

cedar-lean/Cedar/Thm/SymCC/Compiler/WellTyped.lean
```diff
@@ -0,0 +1,2209 @@
...
+  rcases hwt with ⟨_, ⟨⟨hwt⟩⟩⟩
+
+  all_goals
+    simp [compile, compilePrim, Factory.someOf]
```
emina: See previous comments about using `simp` vs `simp only` for proof stability.

cedar-lean/Cedar/Thm/SymCC/Compiler/WellTyped.lean
```diff
@@ -0,0 +1,2209 @@
...
+    cases qty
+    all_goals
+      simp [
+        TermType.ofQualifiedType,
+        CedarType.liftBoolTypes,
+        QualifiedType.liftBoolTypes,
+      ]
+      apply ofType_ignores_liftBool
+
+  private theorem ofRecordType_ignores_liftBool:
```
emina: I find this proof style hard to read because there is no clear visual separation of the theorem statement and proof, via `:= by`.  I suggest rewriting this to use `:= by` with `cases` or `match`.

cedar-lean/Cedar/Thm/SymCC/Compiler/WellTyped.lean
```diff
@@ -0,0 +1,2209 @@
...
+  TermType.isRecordType ty
+:= by
+  intros hty
+  simp [
+    TermType.isCedarRecordType,
+    TermType.isRecordType,
+    TermType.cedarType?,
+  ] at *
+  cases ty
+
```
emina: Remove extra newlines from proofs.

cedar-lean/Cedar/Thm/SymCC/Compiler/WellTyped.lean
emina: +1 to Craig's comment.

cedar-lean/Cedar/Thm/SymCC/Compiler/WellTyped.lean
```diff
@@ -0,0 +1,2209 @@
...
+      TermType.ofType,
+      Term.typeOf,
+      Factory.ifFalse,
+      Factory.noneOf,
+      Factory.someOf,
+      TermPrim.typeOf,
+    ]
+
+  case isEmpty elem_ty hty_expr =>
+    simp [hty_expr, TermType.ofType] at hty_comp_expr hty_get_comp_expr
```
emina: See previous comments about using `simp` vs `simp only`.

cedar-lean/Cedar/Thm/SymCC/Compiler/WellTyped.lean
```diff
@@ -0,0 +1,2209 @@
...
+    · constructor
+      · intros; contradiction
+      · intros; contradiction
+      · have h : TermType.WellFormed εnv.entities (.set (TermType.ofType elem_ty)) := by
+          simp [←hty_get_comp_expr]
+          apply typeOf_wf_term_is_wf
+          assumption
+        cases h; assumption
+      · constructor
+
```
emina: Remove extra newlines.

cedar-lean/Cedar/Thm/SymCC/Compiler/WellTyped.lean
```diff
@@ -0,0 +1,2209 @@
...
+    ]
+    exact (wf_string_like (εs := εnv.entities)
+      hwf_get_comp_expr
+      hty_get_comp_expr).right
+
+/--
+CompileWellTypedCondition decomposes for binaryApp
+TODO: merge this with other eliminate_wt_cond_*
+-/
+theorem CompileWellTypedCondition.eliminate_binaryApp
```
emina: Let's mark all theorems that are not going to be used outside of this file as `private`.

cedar-lean/Cedar/Thm/SymCC/Compiler/WellTyped.lean
```diff
@@ -0,0 +1,2209 @@
...
+  (hattrs_exists : εnv.entities.attrs ety = .some attrs) :
+  UnaryFunction.WellFormed εnv.entities attrs ∧
+  attrs.argType = .entity ety ∧
+  attrs.outType.isCedarRecordType
+:= by
+  have ⟨_, _, hwf_entities⟩ := hwf
+  simp [SymEntities.attrs] at hattrs_exists
+  simp_do_let (Data.Map.find? εnv.entities ety) at hattrs_exists
+  contradiction
+
```
emina: Remove extra new lines.

cedar-lean/Cedar/Thm/SymCC/Compiler/WellTyped.lean
```diff
@@ -0,0 +1,2209 @@
...
+      hty_expr,
+    ]
+
+    simp at hent_attrs_exists
+    have ⟨rty2, hrty2, hrty_rty2⟩ := hent_attrs_exists
+
+    simp at henv_attr_lookup
+    have ⟨attr_ty, henv_attr_lookup, hty_env_attr⟩ := henv_attr_lookup
+
+    -- Show some facts about `εnv` from the hypotheses in `getAttr_entity`
```
emina, line 862: The body of this proof is very long.  The general rule of thumb is that "if it's too long to be a function, it's too long to be a proof" :).  Let's pull out some of these auxiliary proofs (like the `have` below) into their own `private` lemmas:  `private theorem compile_well_typed_getAttr_<name>`.

cedar-lean/Cedar/Thm/SymCC/Compiler/WellTyped.lean
emina: +1.  See the comment above about the rule of thumb for proof length.

cedar-lean/Cedar/Thm/SymCC/Compiler/WellTyped.lean
```diff
@@ -0,0 +1,2209 @@
...
+      · simp at hrefs
+        apply hrefs a x.toExpr a x
+        simp [List.attach₂, hx]
+        any_goals rfl
+        have e : x = (a, x).snd := by rfl
+        rw [e]
+        apply List.sizeOf_snd_lt_sizeOf_list
+        exact hx
+
+/--
```
emina: We already have a mechanism to do this.  See https://github.com/cedar-policy/cedar-spec/blob/2b113dd16e4cd7d1696ec950c9a94ce0edc14a16/cedar-lean/Cedar/Thm/Data/List/Basic.lean#L468

Let's get rid of this inductive predicate and use `Forallᵥ` instead.

cedar-lean/Cedar/Thm/SymCC/Compiler/WellTyped.lean
```diff
@@ -0,0 +1,2209 @@
...
+-/
+inductive MapListValueRelation {α β κ} (p : α → β → Prop) : List (κ × α) → List (κ × β) → Prop
+  | nil : MapListValueRelation p [] []
+  | cons {k₁ k₂ : κ} {x : α} {y : β} {xs : List (κ × α)} {ys : List (κ × β)} :
+    p x y →
+    k₁ = k₂ →
+    MapListValueRelation p xs ys →
+    MapListValueRelation p ((k₁, x) :: xs) ((k₂, y) :: ys)
+
+/--
```
emina: You can delete these two theorems and use this one instead: https://github.com/cedar-policy/cedar-spec/blob/2b113dd16e4cd7d1696ec950c9a94ce0edc14a16/cedar-lean/Cedar/Thm/Data/List/Canonical.lean#L400

cedar-lean/Cedar/Thm/SymCC/Compiler/WellTyped.lean
```diff
@@ -0,0 +1,2209 @@
...
+
+        constructor; assumption; simp
+        assumption
+
+/--
+If `p x y` implies `f x = g y`,
+then `MapListValueRelation p xs ys` implies
+`List.map (Prod.map id f) xs = List.map (Prod.map id g) ys`
+-/
+theorem MapListValueRelation.implies_map_eq_if_p_implies_eq
```
emina: I think you can replace or simplify several of these lemmas with some of the existing ones you'll find here: https://github.com/cedar-policy/cedar-spec/blob/main/cedar-lean/Cedar/Thm/Data/List/Lemmas.lean

Grep for `forall₂` or `Forall₂`.  Note that `Forallᵥ ` is defined in terms of `Forall₂`.

cedar-lean/Cedar/Thm/SymCC/Compiler/WellTyped.lean
```diff
@@ -0,0 +1,2209 @@
...
+        xs = Except.ok ats
+  := by
+    apply List.all_ok_implies_mapM_ok
+    intros p hx
+    have ⟨k, x⟩ := p
+    have ⟨_, hcomp_x, _⟩ := ihxs k x hx
+    simp [hcomp_x]
+
+  -- Compilation of all fields succeeds (the exact version)
+  have hcomp_xs :
```
emina: You should not need to do this, I think.  Grep for `simp_do_let` and see if that does the right thing.

cedar-lean/Cedar/Thm/SymCC/Term/TypeOf.lean
```diff
@@ -415,4 +415,22 @@ theorem typeOf_ifAllSome_option_type {gs : List Term} {t : Term} {ty : TermType}
   simp only [ifAllSome, hty, Factory.ite, noneOf]
   exact typeOf_ite_simplify_option hty
 
+theorem typeOf_not {t : Term} {entities : SymEntities} :
```
emina: We've proved a stronger theorem here:  https://github.com/cedar-policy/cedar-spec/blob/2b113dd16e4cd7d1696ec950c9a94ce0edc14a16/cedar-lean/Cedar/Thm/SymCC/Term/WF.lean#L644

Let's delete this theorem and use the stronger one instead.

emina (changes requested):
Nice work!

I've left comments to help iterate on the proof :)

The high-level bits are:

* Use `simp only` instead of `simp` unless you're closing a goal.  This helps with proof stability.  (Lean can help you make this change; see comments.)

* Use `rcases` only for desctructuring disjunctions and `have` for everything else.

* Break up long proofs into smaller lemmas.  The rule of thumb is that individual proofs shouldn't be longer than, say, 200 lines of code or so (similar to function bodies).

* The proof has re-proved some lemmas we already have.  Let's delete those and replace with existing lemmas (see comments).

cedar-lean/Cedar/Thm/SymCC/Compiler/WellTyped.lean
zhengyao-lin: oh I totally missed this. It's interesting how they are exactly the same lol

cedar-lean/Cedar/Thm/SymCC/Compiler/WellTyped.lean
zhengyao-lin: That didn't seem to work here. I tried `simp_do_let (List.mapM ...)`, and it basically splitted into two goals (i.e. whether `(List.mapM ...)` is `err` or `ok`).

I was trying to apply `List.mapM_pmap_subtype` here to get rid of the `attach`, but Lean is not able to unify
```
(fun x => do
  let __do_lift ← compile x.1.snd εnv
  Except.ok (x.1.fst, __do_lift))
```
with
```
(fun x => ?_ x.1)
```
(i.e. factoring the term `x.1` out)
Is there a more concise way to handle this situation?

cdisselkoen: You shouldn't need to use `List.mapM_pmap_subtype` directly in most cases.  (Maybe we should even make it `private`?)  I'd recommend getting rid of `List.mapM₂` directly at the top of this proof, not by `simp` like you are now, but with `List.mapM₂_eq_mapM`.

zhengyao-lin:
@cdisselkoen @emina Thank you so much for your feedback!

I have gone through the entire proof again and addressed most of your suggestions.
In particular,
- I've replaced all non-final `simp` with `simp only`
- Refactored the proof a bit to reduce proof size; I think all theorems are less than ~150 lines now
- I tried expanding some of the `any_goals` into cases (except for some simple goals such as for `CompileWellTypedCondition.eliminate_*`, or some cases where `any_goals assumption` helps with unification of lemma application)
- All empty lines in proofs are removed
- `MapListValueRelation` and related lemmas are replaced with `Forallᵥ` or `Forall₂`
- `rcases => have`

A few general Lean questions:
- Why is `simp` ok as the final tactic? Does Lean guarantee that `simp` only gets more/stronger rewrites over updates?
- I'm a bit new to the Lean build system, how do I guarantee that `Cedar.Thm.SymCC.Compiler.WellTyped` gets verified with `lake build`? Right now it seems to skip it unless some module imports it

cedar-lean/Cedar/Thm/SymCC/Compiler/WellTyped.lean
zhengyao-lin: I figured out the rewriting issue with `List.mapM₂_eq_mapM` and the proof is now simplified.

It seems to be due to `List.mapM₂_eq_mapM` not general enough wrt the `SizeOf` instances of `α` and `β`. The original statement says
```
theorem mapM₂_eq_mapM [Monad m] [LawfulMonad m]
  (f : (α × β) → m γ)
  (as : List (α × β)) :
  List.mapM₂ as (λ x : { x // sizeOf x.snd < 1 + sizeOf as } => f x.val) =
  List.mapM f as
```
which uses the default `sizeOf` instance.

But the `List.mapM₂` calls in `compile` of `record` seems to use another instance for `Attr` and `Expr`, which causes Lean fail to unify.

In any case, generalizing the theorem to
```
theorem mapM₂_eq_mapM [Monad m] [LawfulMonad m] [SizeOf α] [SizeOf β]
  ...
```
fixed it.

I might be wrong though since I'm not familiar with how typeclasses work in Lean.

cdisselkoen: Nice find!

cedar-lean/Cedar/Thm/Data/List/Lemmas.lean
```diff
@@ -853,6 +920,16 @@ theorem forall₂_implies_mapM_eq {α₁ α₂ β ε} {xs : List α₁} {ys : Li
   case cons h₁ h₂ =>
     simp only [List.mapM_cons, h₁, forall₂_implies_mapM_eq f g h₂, bind_pure_comp]
 
+theorem forall₂_implies_map_eq {α₁ α₂ β} {xs : List α₁} {ys : List α₂} (f : α₁ → β) (g : α₂ → β):
```
emina: We don't need this theorem.  A stronger one is already proven: grep for `forall₂_iff_map_eq`.

cedar-lean/Cedar/Thm/SymCC/Compiler/WellTyped.lean
```diff
@@ -0,0 +1,1789 @@
...
+  t.WellFormed εnv.entities
+:= by
+  have ⟨_, _, hwf⟩ := h
+  simp [compile_wf hwf hcomp]
+
+/--
+If some entity exists in `Γ`, then it must
+also exists in `SymEnv.ofEnv Γ` with the corresponding `SymEntityData`
+-/
+theorem ofEnv_preserves_entity
```
emina: Let's move this theorem, and all theorems related to `SymEnv.ofEnv`, to a new file:  `Cedar.Thm.SymCC.Env.OfEnv`. 

This follows the pattern of putting reusable `Env`-related lemmas into `Cedar.Thm.SymCC.Env`, organized by functionality.  For example, all lemmas related to interpretation of SymEnvs are in `Cedar.Thm.SymCC.Env.Interpet`.   This makes things easier to find later :)

cedar-lean/Cedar/Thm/SymCC/Compiler/WellTyped.lean
```diff
@@ -0,0 +1,1789 @@
...
+    | ext xty => simp [TermType.ofType, CedarType.liftBoolTypes]
+    | set ty =>
+      simp [TermType.ofType, CedarType.liftBoolTypes]
+      apply ofType_eq_ofType_liftBool ty
+    | record rty =>
+      simp [TermType.ofType, CedarType.liftBoolTypes, RecordType.liftBoolTypes]
+      apply ofRecordType_eq_ofRecordType_liftBool
+end
+
+theorem isCedarRecordType_implies_isRecordType
```
emina, line 173: The proof of this theorem can be much shorter:  grep for `isCedarRecordType_implies_term_record_type`.

cedar-lean/Cedar/Thm/SymCC/Compiler/WellTyped.lean
```diff
@@ -0,0 +1,1789 @@
...
+    | bitvec n =>
+      if h : n = 64 then
+        simp [h, TermType.cedarType?] at hty
+      else
+        simp [h, TermType.cedarType?] at hty
+
+/--
+A variant of `wf_ite` to simplify things in this proof
+-/
+theorem wf_typeOf_ite {g t₁ t₂ : Term} {ty : TermType} {entities : SymEntities}
```
emina: Should this be `private`?

emina (changes requested):

cedar-lean/Cedar/Thm/Data/List/Canonical.lean
```diff
@@ -46,6 +46,44 @@ theorem insertCanonical_not_nil [DecidableEq β] [LT β] [DecidableLT β] (f : 
     split at h <;> try trivial
     split at h <;> trivial
 
+theorem mem_insertCanonical [LT β] [Cedar.Data.StrictLT β] [DecidableLT β]
```
shaobo-he-aws: Nit: rename it to `insertCanonical_mem`.

cedar-lean/Cedar/Thm/Data/List/Canonical.lean
```diff
@@ -46,6 +46,44 @@ theorem insertCanonical_not_nil [DecidableEq β] [LT β] [DecidableLT β] (f : 
...
+:= by
+  induction xs
+  case nil => simp [List.insertCanonical]
+  case cons head tail ih =>
+    simp only [List.insertCanonical]
+    split; any_goals simp
+    split; any_goals simp
+    apply Or.inr ih
+
+theorem insertCanonical_new
```
shaobo-he-aws: We should use a more descriptive name here.

cedar-lean/Cedar/Thm/SymCC/Env/WF.lean
```diff
@@ -643,5 +643,27 @@ theorem expr_valid_refs_same_domain {εs₁ εs₂ : SymEntities} {x : Expr} :
   case call_valid ih₁ =>
     exact Expr.ValidRefs.call_valid ih₁
 
+/--
+`SymEnv` being well-formed implies that any
+attribute function is well-formed
+-/
+theorem wf_env_implies_attrs_wf
```
emina: Nit:  this should be called `wf_εnv_implies_attrs_wf`.  (We use `εnv` for symbolic environments and `env` for concrete ones.)

cedar-lean/Cedar/Thm/Data/List/Canonical.lean
zhengyao-lin: how about `insertCanonical_new_element`?

shaobo-he-aws: It's kind of tricky. I was thinking along the line of something like `insertCanonical_preverses_non_duplicate_element`. Feel free to ignore it if you feel it's too complicated.

zhengyao-lin: Sure! I don't have a strong opinion on naming

emina (approved):

cedar-lean/Cedar/Thm/SymCC/Compiler/WellTyped.lean
```diff
@@ -0,0 +1,1628 @@
...
+    compile tx.toExpr εnv = .ok t ∧
+    t.typeOf = .option (TermType.ofType tx.typeOf)
+
+/--
+A sufficient condition for `CompileWellTyped` to hold.
+-/
+def CompileWellTypedCondition (tx : TypedExpr) (Γ : Environment) (εnv : SymEnv) : Prop :=
+  εnv = SymEnv.ofEnv Γ ∧
+  TypedExpr.WellTyped Γ tx ∧
+  εnv.WellFormedFor tx.toExpr
```
cdisselkoen, line 33: Could this be refactored to not require `εnv` as a parameter, and replace all usages of it with `SymEnv.ofEnv Γ`?  This would improve ergonomics and I think clarity?  (Open to arguments why this would not improve clarity)

zhengyao-lin: I was trying this change, but I realized that a lot of the theorems have the `εnv` parameter and rephrasing them in term of `SymEnv.ofEnv Γ` might make them a bit complicated.

Also this definition is likely subject to change once we prove `WellFormedFor` from the first two conditions. So maybe we can simplify things then?

cdisselkoen: Sounds good to me.

Issue #75, opened 2023-07-18
DRT should use `cedar-policy::is_authorized` and `cedar-policy::validate`


khieta:
### Category

DRT infrastructure

### Describe the feature you'd like to request

The DRT infrastructure currently tests the `is_authorized` function in `cedar-policy-core` and the `validate` function in `cedar-policy-validator`. I think it would be cleaner to use the APIs available in the public `cedar-policy` crate instead. This gives a more "end-to-end" feel to our tests and allows us to use the interfaces in [cedar-policy/src/frontend](https://github.com/cedar-policy/cedar/tree/main/cedar-policy/src/frontend) for more consistent input/output formats.

### Describe the solution you'd like

N/A

### Describe alternatives you've considered

N/A

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

cdisselkoen:
I think this is a good idea, but pointing out that one difficulty with implementing this issue is that the DRT currently generates (internal) ASTs for policies, which are not exposed from `cedar-policy`.  My opinion is that the best way to proceed would be for DRT to instead generate ESTs and send them into `cedar-policy` via `Policy::from_json()`.

khieta:
Opened #76. @cdisselkoen feel free to edit the issue description to add additional details/motivation.

PR #647, opened 2025-06-25
[cedar-policy-generators] Add support for generating cedar-policy types

cdisselkoen:
Via optional Cargo feature `cedar-policy`, so that folks who don't need this functionality don't need to bring in the `cedar-policy` dependency.  (No strong reason to do this behind a feature, but it's very easy to do in the implementation so why not.)


john-h-kastner-aws (approved):
seems legit. possibly a step towards https://github.com/cedar-policy/cedar-spec/issues/75 ?

cdisselkoen:
Yes, that is the intent :)

chaluli:
I plan on making this change as I refactor cedar-drt to use the `cedar-lean-ffi` rust bindings of the `cedar-lean` formalization.

PR #648, opened 2025-06-27
Fix validator soundness when `updateSchema` is not used

zhengyao-lin:
**Current status**
The soundness of validator/type checking depends on [this precondition](https://github.com/cedar-policy/cedar-spec/blob/f39f21acdddff3b7bed6f507b2812f638b886f93/cedar-lean/Cedar/Thm/Validation/Typechecker/Types.lean#L102-L108) on the schema:
```
def InstanceOfEntitySchema (entities : Entities) (ets: EntitySchema) : Prop :=
  ∀ uid data, entities.find? uid = some data →
    ∃ entry, ets.find? uid.ty = some entry ∧
      InstanceOfType data.attrs (.record entry.attrs) ∧
      InstanceOfEntityTags data entry
```
In particular, for actions, this requires that all action entity types to be in `EntitySchema`

However, this condition is not satisfied by the Rust version: [`ValidationSchema`](https://github.com/cedar-policy/cedar/blob/b18b7b0a092a7e264548ffcb789e46cfa891371a/cedar-policy-core/src/validator/schema.rs#L231-L249)
```
pub struct ValidatorSchema {
    entity_types: HashMap<EntityType, ValidatorEntityType>,
    actions: HashMap<EntityUID, Arc<Entity>>,
    ...
}
```
where entity types in `actions` and `entity_types` are *disjoint*.

This leads to a precondition violation when Rust `ValidatorSchema` are passed to the Lean version for validation

The current fix for this in DRT is to have a function called [`updateSchema`](https://github.com/cedar-policy/cedar-spec/blob/f39f21acdddff3b7bed6f507b2812f638b886f93/cedar-lean/Cedar/Validation/RequestEntityValidator.lean#L159), which is called [here](https://github.com/cedar-policy/cedar-spec/blob/f39f21acdddff3b7bed6f507b2812f638b886f93/cedar-lean/DiffTest/Main.lean#L146), that adds action entity types to `ets` in the schema (which corresponds to `entity_types`, I think?)
```
structure Schema where
  ets : EntitySchema
  acts : ActionSchema
```

**However, this leads to two issues:**
- In some cases such as `cedar-lean-cli`, `updateSchema` is not called on schema parsed in Rust, leading to a *soundness issue*; for example, when validating [this example Cedar policy by @emina and @chaluli](https://gist.github.com/zhengyao-lin/33666aee8e9694ebed7c9ed7c800e7d6):
```
Rust version:
  × policy set validation failed
  ╰─▶ the types Long and String are not compatible
   ╭─[4:8]
 3 │   if principal.a in N::Action::"bad"
 4 │   then 1 == "2"
   ·        ────────
 5 │   else true
   ╰────

Lean version (via cedar-lean-cli validate policy-set)
Policyset successfully validated
^ this is unsound!
```
Note that you'd have to update `cedar-lean-ffi/Cargo.toml` and `cedar-lean-cli/Cargo.toml` to the latest `cedar-policy` Rust crate, otherwise `cedar-lean-cli` fails due to another front-end issue in Rust that was [fixed](https://github.com/cedar-policy/cedar/pull/1652) recently.

- To make things a bit more tricky, `updateSchema` conflicts with the definition of [`SymEntities.ofSchema`](https://github.com/cedar-policy/cedar-spec/blob/f39f21acdddff3b7bed6f507b2812f638b886f93/cedar-lean/Cedar/SymCC/Env.lean#L199) in SymCC:
```
def SymEntities.ofSchema (ets : EntitySchema) (acts : ActionSchema) : SymEntities :=
  let eData := ets.toList.map λ (ety, sch) => (ety, SymEntityData.ofEntityType ety sch)
  let actTys := (acts.toList.map λ (act, _) => act.ty).eraseDups
  let aData := actTys.map λ actTy => (actTy, SymEntityData.ofActionType actTy actTys acts)
  Map.make (eData ++ aData)
```
Because `updateSchema` already puts all action entity types to `eData`, incorrect symbolic representations of action entity types are used, overriding the correct ones in `aData`.

**This PR**

This PR fixes the validator and its soundness proof even in the absence of `updateSchema`.
This way, we also avoid changing SymCC's `SymEntities.ofSchema` and any related proofs.

This primarily includes these precondition and executable code changes:
- `InstanceOfEntitySchema` is updated to allow action entity types to not occur in `EntitySchema` but only in `ActionSchema`. This is still compatible with `updateSchema`, so it should not break DRT.
- `EntitySchema.descendentOf` is modified to `Environment.descendentOf` to also consider action entities when the entity type is not in `ets`
- `instanceOfEntitySchema` is updated to check the new precondition
- Soundness proofs are fixed

### DIFF
--- a/cedar-lean/Cedar/Thm/Validation/RequestEntityValidation.lean
+++ b/cedar-lean/Cedar/Thm/Validation/RequestEntityValidation.lean
@@ -217,7 +217,7 @@ theorem instance_of_entity_schema_refl {entities : Entities} {ets : EntitySchema
     split at h₀
     any_goals contradiction
     case _ h₃ h₄ h₅ =>
-    simp [ValidActionEntityEntry]
+    simp [WellFormedActionData]
     and_intros
     · simp [h₃]
     · simp [h₂]
--- a/cedar-lean/Cedar/Thm/Validation/RequestEntityValidation.lean
+++ b/cedar-lean/Cedar/Thm/Validation/RequestEntityValidation.lean
@@ -159,10 +159,10 @@ theorem instance_of_request_type_refl {request : Request} {reqty : RequestType} 
   · exact instance_of_entity_type_refl h₃
   · exact instance_of_type_refl h₄
 
-theorem instance_of_schema_refl {entities : Entities} {ets : EntitySchema} {acts : ActionSchema} :
-  instanceOfSchema entities ets acts = .ok () →
-    InstanceOfEntitySchema entities ets acts ∧
-    InstanceOfActionSchema entities acts
+theorem instance_of_schema_refl {entities : Entities} {env : Environment} :
+  instanceOfSchema entities env = .ok () →
+    InstanceOfEntitySchema entities env ∧
+    InstanceOfActionSchema entities env.acts
 := by
   intro h₀
   simp only [InstanceOfEntitySchema]
@@ -170,7 +170,7 @@ theorem instance_of_schema_refl {entities : Entities} {ets : EntitySchema} {acts
   split at h₀
   case h_1 => contradiction
   case h_2 h₀₁ =>
-  generalize h₁ : (λ x : EntityUID × EntityData => instanceOfSchema.instanceOfEntityData ets acts x.fst x.snd) = f
+  generalize h₁ : (λ x : EntityUID × EntityData => instanceOfSchema.instanceOfEntityData env x.fst x.snd) = f
   rw [h₁] at h₀₁
   constructor
   · intro uid data h₂
@@ -178,7 +178,7 @@ theorem instance_of_schema_refl {entities : Entities} {ets : EntitySchema} {acts
     replace h₀ := h₀ (Map.find?_mem_toList h₂)
     rw [← h₁] at h₀
     simp only [instanceOfSchema.instanceOfEntityData] at h₀
-    cases h₂ : Map.find? ets uid.ty <;> simp [h₂] at h₀
+    cases h₂ : Map.find? env.ets uid.ty <;> simp [h₂] at h₀
     case some entry =>
       apply Or.inl
       exists entry
@@ -225,14 +225,14 @@ theorem instance_of_schema_refl {entities : Entities} {ets : EntitySchema} {acts
       case _ h₃ h₄ h₅ =>
       simp [WellFormedActionData]
       and_intros
-      · simp [h₃]
       · simp [h₂]
       · simp [h₄]
       · simp [h₅]
+      · simp [h₃]
   · generalize h₁ : (fun x : EntityUID × ActionSchemaEntry => instanceOfSchema.instanceOfActionSchemaData entities x.fst x.snd) = f
     rw [h₁] at h₀
     intro uid entry h₂
-    replace h₀ := List.forM_ok_implies_all_ok (Map.toList acts) f h₀ (uid, entry)
+    replace h₀ := List.forM_ok_implies_all_ok (Map.toList env.acts) f h₀ (uid, entry)
     replace h₀ := h₀ (Map.find?_mem_toList h₂)
     rw [← h₁] at h₀
     simp only [instanceOfSchema.instanceOfActionSchemaData, beq_iff_eq] at h₀
@@ -253,7 +253,7 @@ theorem request_and_entities_match_env {env : Environment} {request : Request} {
   simp only [entitiesMatchEnvironment] at h₁
   constructor
   exact instance_of_request_type_refl h₀
-  cases h₂ : instanceOfSchema entities env.ets env.acts <;> simp only [h₂, Except.bind_err, Except.bind_ok, reduceCtorEq] at h₁
+  cases h₂ : instanceOfSchema entities env <;> simp only [h₂, Except.bind_err, Except.bind_ok, reduceCtorEq] at h₁
   exact instance_of_schema_refl h₂
 
 theorem request_and_entities_validate_implies_match_schema (schema : Schema) (request : Request) (entities : Entities) :
--- a/cedar-lean/Cedar/Thm/Validation/RequestEntityValidation.lean
+++ b/cedar-lean/Cedar/Thm/Validation/RequestEntityValidation.lean
@@ -159,90 +159,88 @@ theorem instance_of_request_type_refl {request : Request} {reqty : RequestType} 
   · exact instance_of_entity_type_refl h₃
   · exact instance_of_type_refl h₄
 
-theorem instance_of_entity_schema_refl {entities : Entities} {ets : EntitySchema} {acts : ActionSchema} :
-  instanceOfEntitySchema entities ets acts = .ok () → InstanceOfEntitySchema entities ets acts
+theorem instance_of_schema_refl {entities : Entities} {ets : EntitySchema} {acts : ActionSchema} :
+  instanceOfSchema entities ets acts = .ok () →
+    InstanceOfEntitySchema entities ets acts ∧
+    InstanceOfActionSchema entities acts
 := by
   intro h₀
   simp only [InstanceOfEntitySchema]
-  simp only [instanceOfEntitySchema] at h₀
-  generalize h₁ : (λ x : EntityUID × EntityData => instanceOfEntitySchema.instanceOfEntityData ets acts x.fst x.snd) = f
-  rw [h₁] at h₀
-  intro uid data h₂
-  have h₀ := List.forM_ok_implies_all_ok (Map.toList entities) f h₀ (uid, data)
-  replace h₀ := h₀ (Map.find?_mem_toList h₂)
-  rw [← h₁] at h₀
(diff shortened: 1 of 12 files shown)


cedar-lean/Cedar/Validation/RequestEntityValidator.lean
```diff
@@ -99,7 +99,7 @@ For every entity in the store,
    in the type store.
 4. The entity's tags' types are consistent with the tags information in the type store.
 -/
-def instanceOfEntitySchema (entities : Entities) (ets : EntitySchema) : EntityValidationResult :=
+def instanceOfEntitySchema (entities : Entities) (ets : EntitySchema) (as : ActionSchema) : EntityValidationResult :=
```
emina: Let's not separate EntitySchema and ActionSchema.  They are part of a single schema we can pass together: 

```
def instanceOfEntitySchema (entities : Entities) (sch : Schema) : EntityValidationResult :=
```

cedar-lean/Cedar/Validation/Types.lean
```diff
@@ -199,6 +195,16 @@ structure Environment where
   acts : ActionSchema
   reqty : RequestType
 
+def Environment.descendentOf (env : Environment) (ety₁ ety₂ : EntityType) : Bool :=
```
emina, line 201: The check for actions is a bit too imprecise.  Let's try a more precise one:

```
def ActionSchema.maybeDescendentOf (as : ActionSchema) (ety₁ ety₂ : EntityType) : Bool := 
  as.kvs.any λ (act, entry) => act.ty = ety₁ && entry.ancestors.any (EntityUID.ty · == ety₂)

def Environment.descendentOf (env : Environment) (ety₁ ety₂ : EntityType) : Bool :=
  if ety₁ = ety₂
  then true
  else match env.ets.find? ety₁ with
    | .some entry => entry.ancestors.contains ety₂
    | .none       => env.acts.maybeDescendentOf ety₁ ety₂ 
```

cedar-lean/Cedar/Validation/RequestEntityValidator.lean
```diff
@@ -143,7 +150,7 @@ def validateRequest (schema : Schema) (request : Request) : RequestValidationRes
   else .error (.typeError "request could not be validated in any environment")
 
 def entitiesMatchEnvironment (env : Environment) (entities : Entities) : EntityValidationResult :=
-  instanceOfEntitySchema entities env.ets >>= λ _ => instanceOfActionSchema entities env.acts
+  instanceOfEntitySchema entities env.ets env.acts >>= λ _ => instanceOfActionSchema entities env.acts
```
emina: Let's remove `updateSchema` from the codebase and make sure diff testing continues to work in its absence.

cedar-lean/Cedar/Validation/RequestEntityValidator.lean
cdisselkoen: If we do that, let's rename `instanceOfEntitySchema` since it no longer directly works on `EntitySchema`

zhengyao-lin: sounds good. just to confirm, do we want to rename the spec version `InstanceOfEntitySchema` too, or just this executable version?

cedar-lean/Cedar/Thm/Validation/RequestEntityValidation.lean
```diff
@@ -159,76 +159,88 @@ theorem instance_of_request_type_refl {request : Request} {reqty : RequestType}
   · exact instance_of_entity_type_refl h₃
   · exact instance_of_type_refl h₄
 
-theorem instance_of_entity_schema_refl {entities : Entities} {ets : EntitySchema} :
-  instanceOfEntitySchema entities ets = .ok () → InstanceOfEntitySchema entities ets
+theorem instance_of_schema_refl {entities : Entities} {ets : EntitySchema} {acts : ActionSchema} :
```
emina: Let's pass the entity and action schema together as one schema argument:

```
theorem instance_of_schema_refl {entities : Entities} {sch : Schema} : 
```

cedar-lean/Cedar/Thm/Validation/Typechecker/Types.lean
```diff
@@ -92,20 +92,42 @@ def IsValidEntityEID (entry: EntitySchemaEntry) (eid: String) : Prop :=
...
+    IsValidEntityEID entry uid.eid ∧
+    InstanceOfType data.attrs (.record entry.attrs) ∧
+    (∀ ancestor, ancestor ∈ data.ancestors → ancestor.ty ∈ entry.ancestors) ∧
+    InstanceOfEntityTags data entry
+
+/--
+Similar to `WellFormedEntityData`, but a special case for action entities
+since they are stored disjoint from `ets`
+-/
+def WellFormedActionData (uid : EntityUID) (data : EntityData) (ets : EntitySchema) (as : ActionSchema) : Prop :=
```
emina: I suggest revising this definition to pass around the schema and to limit the scope of quantifiers (i.e., pull the constraints that don't depend on the quantified variable `entry` outside of the quantifier):

```
def WellFormedActionData (uid : EntityUID) (data : EntityData) (sch : Schema)  : Prop := 
  sch.ets.find? uid = .none ∧
  data.attrs = .empty ∧
  data.tags = .empty ∧ 
  ∃ entry,  sch.acts.find? uid = some entry 
```

cedar-lean/Cedar/Validation/RequestEntityValidator.lean
```diff
@@ -93,14 +93,19 @@ def instanceOfRequestType (request : Request) (reqty : RequestType) (schema: Ent
...
 3. The entity's ancestors' types are consistent with the ancestor information
    in the type store.
 4. The entity's tags' types are consistent with the tags information in the type store.
+
+For every action in the entity store, the action's ancestors are consistent
+with the ancestor information in the action store.
 -/
-def instanceOfEntitySchema (entities : Entities) (ets : EntitySchema) : EntityValidationResult :=
-  entities.toList.forM λ (uid, data) => instanceOfEntityData uid data
+def instanceOfSchema (entities : Entities) (ets : EntitySchema) (as : ActionSchema) : EntityValidationResult :=
```
emina: Let's pass the entity and action schema together as one schema argument:

```
def instanceOfSchema (entities : Entities) (sch : Schema) : EntityValidationResult :=
```

cedar-lean/Cedar/Thm/Validation/Typechecker/Types.lean
```diff
@@ -92,20 +92,42 @@ def IsValidEntityEID (entry: EntitySchemaEntry) (eid: String) : Prop :=
...
+    -- Action entities cannot have attributes or tags
+    data.attrs = .empty ∧
+    -- TODO: or use this? InstanceOfType data.attrs (.record .empty) ∧
+    data.tags = .empty
+    -- `ancestors` consistency is guaranteed by `InstanceOfActionSchema`
+
+/--
+Each entry in the store is valid
+-/
+def InstanceOfEntitySchema (entities : Entities) (ets : EntitySchema) (as : ActionSchema) : Prop :=
```
emina: Let's change this definition so it takes one schema argument:

```
def InstanceOfEntitySchema (entities : Entities) (sch : Schema) : Prop :=
```

cedar-lean/Cedar/Thm/Validation/RequestEntityValidation.lean
zhengyao-lin: are you ok with an Environment argument instead?
```
instance_of_schema_refl {entities : Entities} {env : Environment}
```
the caller of `instance_of_schema_refl` only has an `Environment` and we need to convert that to `Schema` otherwise.

cedar-lean/Cedar/Thm/Validation/Typechecker/Types.lean
```diff
@@ -110,24 +110,24 @@ def WellFormedEntityData (uid : EntityUID) (data : EntityData) (ets : EntitySche
...
 /--
 Each entry in the store is valid
 -/
-def InstanceOfEntitySchema (entities : Entities) (ets : EntitySchema) (as : ActionSchema) : Prop :=
+def InstanceOfEntitySchema (entities : Entities) (env : Environment) : Prop :=
   ∀ (uid : EntityUID) (data : EntityData),
     entities.find? uid = some data →
-    WellFormedEntityData uid data ets ∨
-    WellFormedActionData uid data ets as
+    WellFormedEntityData uid data env.ets ∨
```
emina, line 129: This definition currently allows two conflicting facts to hold: `uid` is in `env.acts` and `uid.ty` is in `env.ets`. 

That's because `WellFormedEntityData uid data env.ets` doesn't check that `uid` is _not_ in `env.acts`.  

We need to strengthen the definition of `WellFormedEntityData` to disallow this.

I'll approve this PR for now, and we can fix it in the next PR that defines well-formedness.

shaobo-he-aws (approved):
LGTM

Issue #653, opened 2025-07-01
Docker Build Failure for Cedar Lean CLI


clairew:
### Before opening, please confirm:

- [x] I have [searched for duplicate or closed issues](https://github.com/cedar-policy/cedar/issues?q=is%3Aissue+).
- [x] I have read the guide for [submitting bug reports](https://github.com/cedar-policy/cedar/blob/main/CONTRIBUTING.md#bug-reports).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### Bug Category

Other

### Describe the bug

I used the provided Dockerfile in `cedar-spec/cedar-lean-cli` to build the CLI, and the docker build fails. 

### Expected behavior

Docker build doesn't fail. 

### Reproduction steps

```
# Create a docker image that is identified with the tag "cedar-lean-cli"
cd cedar-lean-cli                                   # Enter the cedar-lean-cli directory
docker build -t cedar-lean-cli -f ./Dockerfile ..
```

### Code Snippet

In[ `cedar-spec/cedar-lean-ffi/src/messages.rs` ](https://github.com/cedar-policy/cedar-spec/blob/main/cedar-lean-ffi/src/messages.rs#L177)

```rust
impl proto::ValidationRequest {
    pub(crate) fn new(policyset: &PolicySet, schema: &Schema, mode: &ValidationMode) -> Self {
        Self {
            schema: Some(cedar_policy::proto::models::Schema::from(schema)),
            policies: Some(cedar_policy::proto::models::PolicySet::from(policyset)),
            mode: cedar_policy::proto::models::ValidationMode::from(mode).into(),
        }
    }
}
```

### Log output

```
15.83    Compiling cedar-policy-formatter v4.5.0
26.88 error[E0277]: the trait bound `cedar_policy::proto::models::ValidationMode: From<&cedar_policy::ValidationMode>` is not satisfied
26.88    --> /opt/src/cedar-spec/cedar-lean-ffi/src/messages.rs:177:19
26.88     |
26.88 177 |             mode: cedar_policy::proto::models::ValidationMode::from(mode).into(),
26.88     |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `From<&cedar_policy::ValidationMode>` is not implemented for `cedar_policy::proto::models::ValidationMode`
26.88     |
[...]

PR #656, opened 2025-07-01
Update proto conversions to be compatiable with all versions of cedar-policy >= 4.4.0

chaluli:
Addresses issue #653 

Updates protobuf conversion for Validation mode (makes compatible with both main and v4.5.0).


chaluli:
> How did we miss this? Are we not running the Cedar Lean CLI Dockerfile in CI?

We weren't running the Cedar Lean CLI Dockerfile in CI. I've added it to CI on this PR so that it won't go unnoticed.

PR #657, opened 2025-07-02
Define `Environment.WellFormed` and strengthen `InstanceOfType`

zhengyao-lin:
Goals of this PR:
1. It has been [discussed before](https://github.com/cedar-policy/cedar-spec/pull/530#issuecomment-2644517554) that the current definition of `InstanceOfEntityType` needs to be strengthened, because it does not enforce that the entity type is even a valid type in the schema, or that the entity id is valid in the case of an enum entity.
2. `Environment` currently does not have any well-formedness condition (e.g., request types and ancestors are valid entity types, enum entity is non-empty, etc.), which will be important later on if we want to reduce assumptions on, e.g., `SymEnv.WellFormed` to `Environment.WellFormed`, in some theorems about SymCC.

These two goals are dependent, and involve a wide range of changes to definitions and proofs in the validator (and related modules such as TPE).

To summarize the primary changes:
1. A new definition of [`Environment.WellFormed`](https://github.com/cedar-policy/cedar-spec/blob/08e651bd066d1e1a320735271afb34c4fec7cfa6/cedar-lean/Cedar/Thm/Validation/Typechecker/WF.lean#L122), which includes a definition of `CedarType.WellFormed`. **Please help me vet this because this will be a precondition of many important soundness theorem**.
2. `InstanceOfType` and `InstanceOfEntityType` are strengthened
3. `RequestAndEntitiesMatchEnvironment` now include a new condition of `env.WellFormed` (this is to avoid changing the statement of many typechecker theorems; but we could also put this separately)
4. With changes above, it turns out that `type_is_habited`, which is extensively used in type checker soundness proofs, now requires both `env.WellFormed` and `CedarType.WellFormed`. As a result, one proof structure change is that theorems in `Cedar/Thm/Validation/Typechecker.lean` now depends on some results in `Cedar/Thm/Validation/WellTyped.lean` (in particular `typechecked_has_well_formed_type` and `type_of_is_inhabited`).
5. The executable versions in `Cedar/Validation/RequestEntityValidator.lean` are extended with environment/schema well-formedness checks (currently there are some placeholders before settling the definition of `Environment.WellFormed`).
6. The remaining proofs are fixed. Most high-level statements are the same, except that the precondition may have been strengthened to include `Environment.WellFormed`

Thanks in advance for any suggestions!!

This PR might also needs some change after #648 is merged, so making this a draft for now.

### DIFF
--- a/cedar-lean/Cedar/Validation/RequestEntityValidator.lean
+++ b/cedar-lean/Cedar/Validation/RequestEntityValidator.lean
@@ -32,6 +32,11 @@ inductive EntityValidationError where
 
 abbrev EntityValidationResult := Except EntityValidationError Unit
 
+inductive EnvironmentValidationError where
+| typeError (msg : String)
+
+abbrev EnvironmentValidationResult := Except EnvironmentValidationError Unit
+
 def instanceOfBoolType (b : Bool) (bty : BoolType) : Bool :=
   match b, bty with
   | true, .tt => true
@@ -39,11 +44,9 @@ def instanceOfBoolType (b : Bool) (bty : BoolType) : Bool :=
   | _, .anyBool => true
   | _, _ => false
 
-def instanceOfEntityType (e : EntityUID) (ety : EntityType ) (eids: EntityType → Option (Set String)) : Bool :=
+def instanceOfEntityType (e : EntityUID) (ety : EntityType) (env: Environment) : Bool :=
   ety == e.ty &&
-  match eids ety with
-  | .some eids => eids.contains e.eid
-  | _ => true
+  (env.ets.isValidEntityUID e || env.acts.contains e)
 
 def instanceOfExtType (ext : Ext) (extty: ExtType) : Bool :=
 match ext, extty with
@@ -58,17 +61,17 @@ match rty.find? k with
     | .some qty => if qty.isRequired then r.contains k else true
     | _ => true
 
-def instanceOfType (v : Value) (ty : CedarType) (schema: EntitySchema) : Bool :=
+def instanceOfType (v : Value) (ty : CedarType) (env : Environment) : Bool :=
   match v, ty with
   | .prim (.bool b), .bool bty => instanceOfBoolType b bty
   | .prim (.int _), .int => true
   | .prim (.string _), .string => true
-  | .prim (.entityUID e), .entity ety => instanceOfEntityType e ety schema.entityTypeMembers?
-  | .set s, .set ty => s.elts.attach.all (λ ⟨v, _⟩ => instanceOfType v ty schema)
+  | .prim (.entityUID e), .entity ety => instanceOfEntityType e ety env
+  | .set s, .set ty => s.elts.attach.all (λ ⟨v, _⟩ => instanceOfType v ty env)
   | .record r, .record rty =>
     r.kvs.all (λ (k, _) => rty.contains k) &&
     (r.kvs.attach₂.all (λ ⟨(k, v), _⟩ => (match rty.find? k with
-        | .some qty => instanceOfType v qty.getType schema
+        | .some qty => instanceOfType v qty.getType env
         | _ => true))) &&
     rty.kvs.all (λ (k, _) => requiredAttributePresent r rty k)
   | .ext x, .ext xty => instanceOfExtType x xty
@@ -85,11 +88,11 @@ def instanceOfType (v : Value) (ty : CedarType) (schema: EntitySchema) : Bool :=
         simp only [Map.mk.sizeOf_spec]
         omega
 
-def instanceOfRequestType (request : Request) (reqty : RequestType) (schema: EntitySchema) : Bool :=
-  instanceOfEntityType request.principal reqty.principal schema.entityTypeMembers? &&
+def instanceOfRequestType (request : Request) (reqty : RequestType) (env: Environment) : Bool :=
+  instanceOfEntityType request.principal reqty.principal env &&
   request.action == reqty.action &&
-  instanceOfEntityType request.resource reqty.resource schema.entityTypeMembers? &&
-  instanceOfType request.context (.record reqty.context) schema
+  instanceOfEntityType request.resource reqty.resource env &&
+  instanceOfType request.context (.record reqty.context) env
 
 /--
 For every entity in the store,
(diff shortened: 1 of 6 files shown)


cedar-lean/Cedar/Validation/RequestEntityValidator.lean
```diff
@@ -99,21 +102,21 @@ For every entity in the store,
    in the type store.
 4. The entity's tags' types are consistent with the tags information in the type store.
 -/
-def instanceOfEntitySchema (entities : Entities) (ets : EntitySchema) : EntityValidationResult :=
+def instanceOfEntitySchema (env : Environment) (entities : Entities) : EntityValidationResult :=
```
emina, line 105: Let's swap the two arguments to this definition to match their order in other `instanceOf` functions:

```
def instanceOfEntitySchema (entities : Entities) (env : Environment)  : EntityValidationResult :=
```

cedar-lean/Cedar/Validation/RequestEntityValidator.lean
```diff
@@ -135,22 +138,34 @@ where
                        else .error (.typeError "action ancestors inconsistent with type store information")
       | _ => .error (.typeError s!"action type {uid.eid} not defined in type store")
 
-def requestMatchesEnvironment (env : Environment) (request : Request): Bool := instanceOfRequestType request env.reqty env.ets
+def requestMatchesEnvironment (env : Environment) (request : Request): Bool := instanceOfRequestType request env.reqty env
```
emina, line 141: Let's swap the two arguments to this definition to match their order in other `instanceOf` functions.

cedar-lean/Cedar/Validation/RequestEntityValidator.lean
```diff
@@ -135,22 +138,34 @@ where
...
 
 def validateRequest (schema : Schema) (request : Request) : RequestValidationResult :=
   if ((schema.environments.any (requestMatchesEnvironment · request)))
   then .ok ()
   else .error (.typeError "request could not be validated in any environment")
 
 def entitiesMatchEnvironment (env : Environment) (entities : Entities) : EntityValidationResult :=
```
emina, line 148: Let's swap the two arguments to this definition to match their order in other `instanceOf` functions.

cedar-lean/Cedar/Thm/Validation/Typechecker/WF.lean
```diff
@@ -0,0 +1,237 @@
...
+open Cedar.Data
+open Cedar.Validation
+open Cedar.Spec
+
+----- Well-formedness of environment -----
+
+def ActionSchema.isActionEntityType (acts : ActionSchema) (ety : EntityType) : Prop :=
+  ∃ uid, acts.contains uid ∧ uid.ty = ety
+
+/--
```
emina, line 36: This comment is phrased to suggest that we want an exclusive `or` in the definition, but the definition itself doesn't preclude an entity type from being in both `env.ets` and `env.acts`.  

We might want to adjust the definition to force exclusiveness:

```
def EntityType.WellFormed (env : Environment) (ety : EntityType) : Prop :=
  if env.ets.contains ety 
  then ¬(env.acts.isActionEntityType ety) 
  else env.acts.isActionEntityType ety
```

cedar-lean/Cedar/Thm/Validation/Typechecker/WF.lean
```diff
@@ -0,0 +1,237 @@
...
+  | record_wf {rty : RecordType}
+    -- Well-formed as a `Map`
+    (h₁ : rty.WellFormed)
+    -- Each attribute type is well-formed
+    (h₂ : ∀ attr qty, rty.find? attr = some qty → QualifiedType.WellFormed env qty) :
+    CedarType.WellFormed env (.record rty)
+  | ext_wf {xty : ExtType} : CedarType.WellFormed env (.ext xty)
+end
+
+def StandardSchemaEntry.WellFormed (env : Environment) (entry : StandardSchemaEntry) : Prop :=
```
emina, line 74: This definition needs to be stronger.  In particular, a standard schema entry can't have either an action type or an enumerated type as its ancestor.  See [here](https://github.com/cedar-policy/cedar-spec/blob/282c9094aa49000df8cf073c2466e64736baf791/cedar-lean/Cedar/Thm/SymCC/Data/Hierarchy.lean#L19).

cedar-lean/Cedar/Thm/Validation/Typechecker/WF.lean
```diff
@@ -0,0 +1,237 @@
...
+  | enum es => es.WellFormed ∧ ¬es.isEmpty
+
+def EntitySchema.WellFormed (env : Environment) (ets : EntitySchema) : Prop :=
+  Map.WellFormed ets ∧
+  ∀ ety entry, ets.find? ety = some entry → entry.WellFormed env
+
+def EntityUID.WellFormed (env : Environment) (uid : EntityUID) : Prop :=
+  env.ets.isValidEntityUID uid ∨ env.acts.contains uid
+
+def ActionSchemaEntry.WellFormed (env : Environment) (entry : ActionSchemaEntry) : Prop :=
```
emina, line 96: This definition needs to be stronger.  In particular, the only valid ancestors for action entities are other actions.  And the ancestor set for action entities must be acyclic and transitive.  See [here](https://github.com/cedar-policy/cedar-spec/blob/main/cedar-lean/Cedar/Thm/SymCC/Data/Hierarchy.lean).

cedar-lean/Cedar/Thm/Validation/Typechecker/WF.lean
```diff
@@ -0,0 +1,237 @@
...
+  (∀ uid entry, acts.find? uid = some entry → entry.WellFormed env) ∧
+  (∀ uid, acts.contains uid → ¬env.ets.contains uid.ty)
+
+def RequestType.WellFormed (env : Environment) (reqty : RequestType) : Prop :=
+  EntityType.WellFormed env reqty.principal ∧
+  env.acts.contains reqty.action ∧
+  EntityType.WellFormed env reqty.resource ∧
+  -- Context is a well-formed `RecordType`
+  (CedarType.record reqty.context).WellFormed env
+  -- TODO: Enforce that principal/resource/context are valid for the action?
```
emina, line 120: Yes, we should enforce this.

zhengyao-lin:
Closing as #658 and #659 are merged

PR #658, opened 2025-07-02
Add a definition `Environment.WellFormed`

zhengyao-lin:
As suggested by @emina, this PR is separated from #657 to only include a new definition for `Environment.WellFormed` without theorem/proof changes.

NOTE: this PR does not fix the types and proofs yet; so CI will be failing. I won't merge it until the proof is fixed in another PR.

### DIFF
--- a/cedar-lean/Cedar/Thm/Validation/Typechecker/Types.lean
+++ b/cedar-lean/Cedar/Thm/Validation/Typechecker/Types.lean
@@ -19,6 +19,7 @@ import Cedar.Validation
 import Cedar.Thm.Data.Control
 import Cedar.Thm.Data.Set
 import Cedar.Thm.Validation.Typechecker.LUB
+import Cedar.Thm.Validation.Typechecker.WF
 
 /-!
 This file contains useful definitions and lemmas about Cedar types.
--- a/cedar-lean/Cedar/Thm/Validation/Typechecker/Types.lean
+++ b/cedar-lean/Cedar/Thm/Validation/Typechecker/Types.lean
@@ -141,34 +141,34 @@ def InstanceOfWellFormedEnvironment (request : Request) (entities : Entities) (e
 
 ----- Theorems -----
 
-theorem false_is_instance_of_ff :
-  InstanceOfType (Value.prim (Prim.bool false)) (CedarType.bool BoolType.ff)
+theorem false_is_instance_of_ff {env : Environment} :
+  InstanceOfType env (Value.prim (Prim.bool false)) (CedarType.bool BoolType.ff)
 := by
   apply InstanceOfType.instance_of_bool
   simp [InstanceOfBoolType]
 
-theorem true_is_instance_of_tt :
-  InstanceOfType (Value.prim (Prim.bool true)) (CedarType.bool BoolType.tt)
+theorem true_is_instance_of_tt {env : Environment} :
+  InstanceOfType env (Value.prim (Prim.bool true)) (CedarType.bool BoolType.tt)
 := by
   apply InstanceOfType.instance_of_bool
   simp [InstanceOfBoolType]
 
-theorem bool_is_instance_of_anyBool (b : Bool) :
-  InstanceOfType (Value.prim (Prim.bool b)) (CedarType.bool BoolType.anyBool)
+theorem bool_is_instance_of_anyBool {env : Environment} (b : Bool) :
+  InstanceOfType env (Value.prim (Prim.bool b)) (CedarType.bool BoolType.anyBool)
 := by
   apply InstanceOfType.instance_of_bool
   simp [InstanceOfBoolType]
 
-theorem instance_of_bool_is_bool {v₁ : Value} {bty : BoolType} :
-  InstanceOfType v₁ (CedarType.bool bty) →
+theorem instance_of_bool_is_bool {env : Environment} {v₁ : Value} {bty : BoolType} :
+  InstanceOfType env v₁ (CedarType.bool bty) →
   ∃ b, v₁ = .prim (.bool b)
 := by
   intro h₁
   rcases h₁ with ⟨b, _, _⟩
   exists b
 
-theorem instance_of_ff_is_false {v₁ : Value} :
-  InstanceOfType v₁ (CedarType.bool BoolType.ff) →
+theorem instance_of_ff_is_false {env : Environment} {v₁ : Value} :
+  InstanceOfType env v₁ (CedarType.bool BoolType.ff) →
   v₁ = .prim (.bool false)
 := by
   intro h₁
@@ -178,8 +178,8 @@ theorem instance_of_ff_is_false {v₁ : Value} :
     cases b <;> simp at h₁
     rfl
 
-theorem instance_of_tt_is_true {v₁ : Value} :
-  InstanceOfType v₁ (CedarType.bool BoolType.tt) →
+theorem instance_of_tt_is_true {env : Environment} {v₁ : Value} :
+  InstanceOfType env v₁ (CedarType.bool BoolType.tt) →
   v₁ = .prim (.bool true)
 := by
   intro h₁
@@ -189,13 +189,13 @@ theorem instance_of_tt_is_true {v₁ : Value} :
     cases b <;> simp at h₁
     rfl
 
-theorem instance_of_anyBool_is_bool {v₁ : Value} :
-  InstanceOfType v₁ (CedarType.bool BoolType.anyBool) →
+theorem instance_of_anyBool_is_bool {env : Environment} {v₁ : Value} :
+  InstanceOfType env v₁ (CedarType.bool BoolType.anyBool) →
   ∃ b, v₁ = .prim (.bool b)
 := instance_of_bool_is_bool
 
-theorem instance_of_int_is_int {v₁ : Value} :
-  InstanceOfType v₁ CedarType.int →
+theorem instance_of_int_is_int {env : Environment} {v₁ : Value} :
+  InstanceOfType env v₁ CedarType.int →
   ∃ i, v₁ = .prim (.int i)
 := by
   intro h₁
@@ -203,8 +203,8 @@ theorem instance_of_int_is_int {v₁ : Value} :
   rename_i y
   exists y
 
-theorem instance_of_string_is_string {v₁ : Value} :
-  InstanceOfType v₁ CedarType.string →
+theorem instance_of_string_is_string {env : Environment} {v₁ : Value} :
+  InstanceOfType env v₁ CedarType.string →
   ∃ s, v₁ = .prim (.string s)
 := by
   intro h₁
@@ -212,8 +212,8 @@ theorem instance_of_string_is_string {v₁ : Value} :
   rename_i y
   exists y
 
-theorem instance_of_entity_type_is_entity {ety : EntityType} :
-  InstanceOfType v₁ (.entity ety) →
+theorem instance_of_entity_type_is_entity {env : Environment} {ety : EntityType} :
+  InstanceOfType env v₁ (.entity ety) →
   ∃ euid, euid.ty = ety ∧ v₁ = .prim (.entityUID euid)
 := by
   intro h₁
@@ -223,8 +223,8 @@ theorem instance_of_entity_type_is_entity {ety : EntityType} :
   exists euid
   simp [h₁]
 
-theorem instance_of_decimal_type_is_decimal {v₁ : Value} :
-  InstanceOfType v₁ (CedarType.ext ExtType.decimal) →
+theorem instance_of_decimal_type_is_decimal {env : Environment} {v₁ : Value} :
+  InstanceOfType env v₁ (CedarType.ext ExtType.decimal) →
   ∃ d, v₁ = .ext (.decimal d)
 := by
   intro h₁
@@ -235,8 +235,8 @@ theorem instance_of_decimal_type_is_decimal {v₁ : Value} :
   rename_i d _
   exists d
 
-theorem instance_of_ipAddr_type_is_ipAddr {v₁ : Value} :
-  InstanceOfType v₁ (CedarType.ext ExtType.ipAddr) →
+theorem instance_of_ipAddr_type_is_ipAddr {env : Environment} {v₁ : Value} :
+  InstanceOfType env v₁ (CedarType.ext ExtType.ipAddr) →
   ∃ d, v₁ = .ext (.ipaddr d)
 := by
   intro h₁
@@ -247,8 +247,8 @@ theorem instance_of_ipAddr_type_is_ipAddr {v₁ : Value} :
   rename_i ip _
   exists ip
 
-theorem instance_of_datetime_type_is_datetime {v₁ : Value} :
-  InstanceOfType v₁ (CedarType.ext ExtType.datetime) →
+theorem instance_of_datetime_type_is_datetime {env : Environment} {v₁ : Value} :
+  InstanceOfType env v₁ (CedarType.ext ExtType.datetime) →
   ∃ d, v₁ = .ext (.datetime d)
 := by
   intro h₁
@@ -259,8 +259,8 @@ theorem instance_of_datetime_type_is_datetime {v₁ : Value} :
   rename_i d _
   exists d
 
-theorem instance_of_duration_type_is_duration {v₁ : Value} :
-  InstanceOfType v₁ (CedarType.ext ExtType.duration) →
+theorem instance_of_duration_type_is_duration {env : Environment} {v₁ : Value} :
+  InstanceOfType env v₁ (CedarType.ext ExtType.duration) →
   ∃ d, v₁ = .ext (.duration d)
 := by
   intro h₁
@@ -271,8 +271,8 @@ theorem instance_of_duration_type_is_duration {v₁ : Value} :
   rename_i d _
   exists d
 
-theorem instance_of_set_type_is_set {v : Value} {ty : CedarType} :
-  InstanceOfType v (.set ty) →
+theorem instance_of_set_type_is_set {env : Environment} {v : Value} {ty : CedarType} :
+  InstanceOfType env v (.set ty) →
   ∃ s, v = .set s
 := by
   intro h₁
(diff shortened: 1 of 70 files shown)


cedar-lean/Cedar/Thm/Validation/Typechecker/WF.lean
```diff
@@ -0,0 +1,149 @@
...
+  | record_wf {rty : RecordType}
+    -- Well-formed as a `Map`
+    (h₁ : rty.WellFormed)
+    -- Each attribute type is well-formed
+    (h₂ : ∀ attr qty, rty.find? attr = some qty → QualifiedType.WellFormed env qty) :
+    CedarType.WellFormed env (.record rty)
+  | ext_wf {xty : ExtType} : CedarType.WellFormed env (.ext xty)
+end
+
+def StandardSchemaEntry.WellFormed (env : Environment) (entry : StandardSchemaEntry) : Prop :=
```
emina, line 74: Let simplify this definition by factoring out the `match` that checks if an entry is standard or not into a separate function inside of Cedar.Validation.Types:

```
def EntitySchemaEntry.isStandard : EntitySchemaEntry → Bool  
  | .standard _ => true
  | .enum _     => false
```

So, this definition replaces the `match` with a call to the above function:

```
...
(∀ anc ∈ entry.ancestors,
    ∃ entry, env.ets.find? anc = some entry ∧ entry.isStandard)
...
```

cedar-lean/Cedar/Thm/Validation/Typechecker/WF.lean
```diff
@@ -0,0 +1,149 @@
...
+    entry₂.ancestors ⊆ entry₁.ancestors
+
+def ActionSchema.WellFormed (env : Environment) (acts : ActionSchema) : Prop :=
+  Map.WellFormed acts ∧
+  (∀ uid entry, acts.find? uid = some entry → entry.WellFormed env) ∧
+  (∀ uid, acts.contains uid → ¬env.ets.contains uid.ty) ∧
+  acts.AcyclicActionHierarchy ∧
+  acts.TransitiveActionHierarchy
+
+def RequestType.WellFormed (env : Environment) (reqty : RequestType) : Prop :=
```
emina, line 128: I don't think we need the well-formedness constraints on principal, resource, and context. 

They will follow from the well-formedness of the action schema and entity schema for `env`.

So, this definition can be simplified to:
```
def RequestType.WellFormed (env : Environment) (reqty : RequestType) : Prop :=
 ∃ entry, env.acts.find? reqty.action = some entry ∧
    reqty.principal ∈ entry.appliesToPrincipal ∧
    reqty.resource ∈ entry.appliesToResource ∧
    reqty.context = entry.context
```

cedar-lean/Cedar/Thm/Validation/Typechecker/WF.lean
```diff
@@ -0,0 +1,149 @@
...
+  ∀ uid₁ entry₁ uid₂ entry₂,
+    acts.find? uid₁ = .some entry₁ →
+    acts.find? uid₂ = .some entry₂ →
+    uid₂ ∈ entry₁.ancestors →
+    entry₂.ancestors ⊆ entry₁.ancestors
+
+def ActionSchema.WellFormed (env : Environment) (acts : ActionSchema) : Prop :=
+  Map.WellFormed acts ∧
+  (∀ uid entry, acts.find? uid = some entry → entry.WellFormed env) ∧
+  (∀ uid, acts.contains uid → ¬env.ets.contains uid.ty) ∧
```
emina, line 124: Since we have this disjointness check here, which prevents any action type from appearing in the entity schema, we can simplify the definition of `EntityType.WellFormed` back to using just "or".

```
def EntityType.WellFormed (env : Environment) (ety : EntityType) : Prop :=
  env.ets.contains ety ∨  env.acts.isActionEntityType ety
```

Then, we can prove from `Environment.WellFormed` (and therefore `ActionSchema.WellFormed`) that every `ety` in the Environment is either in `ets` or in `acts` but not in both.

cedar-lean/Cedar/Thm/Validation/Typechecker/WF.lean
```diff
@@ -0,0 +1,149 @@
...
+open Cedar.Data
+open Cedar.Validation
+open Cedar.Spec
+
+----- Well-formedness of environment -----
+
+def ActionSchema.isActionEntityType (acts : ActionSchema) (ety : EntityType) : Prop :=
+  ∃ uid, acts.contains uid ∧ uid.ty = ety
+
+/--
```
emina, line 36: Line 129 below lets us make this an inclusive "or" check (see comment).

cedar-lean/Cedar/Thm/Validation/Typechecker/WF.lean
```diff
@@ -0,0 +1,149 @@
...
+
+namespace Cedar.Validation
+
+open Cedar.Data
+open Cedar.Validation
+open Cedar.Spec
+
+----- Well-formedness of environment -----
+
+def ActionSchema.isActionEntityType (acts : ActionSchema) (ety : EntityType) : Prop :=
```
emina: Naming convention:  definitions that return a Prop start with an upper case letter, so `ActionSchema.IsActionEntityType`.

See  [here](https://github.com/cedar-policy/cedar-spec/blob/main/cedar-lean/GUIDE.md).

emina (changes requested):

emina (approved):

cedar-lean/Cedar/Thm/Validation/Typechecker/Types.lean
```diff
@@ -48,42 +49,42 @@ def InstanceOfExtType : Ext → ExtType → Prop
...
+      InstanceOfType env (.prim (.string _)) .string
   | instance_of_entity (e : EntityUID) (ety: EntityType)
-      (h₁ : InstanceOfEntityType e ety) :
-      InstanceOfType (.prim (.entityUID e)) (.entity ety)
+      (h₁ : InstanceOfEntityType env e ety) :
+      InstanceOfType env (.prim (.entityUID e)) (.entity ety)
   | instance_of_set (s : Set Value) (ty : CedarType)
-      (h₁ : forall v, v ∈ s → InstanceOfType v ty) :
-      InstanceOfType (.set s) (.set ty)
+      (h₁ : forall v, v ∈ s → InstanceOfType env v ty) :
```
emina: Let's use `∀` instead of `forall`.

cedar-lean/Cedar/Thm/Validation/Typechecker/Types.lean
```diff
@@ -99,18 +100,18 @@ For every entity `(uid, data)` in the store,
...
+    InstanceOfEntityTags env data entry
 
 /--
 Similar to `WellFormedEntityData`, but a special case for action entities
 since they are stored disjoint from `ets`
 -/
-def WellFormedActionData (uid : EntityUID) (data : EntityData) (env : Environment) : Prop :=
+def WellFormedActionData (env : Environment) (uid : EntityUID) (data : EntityData) : Prop :=
   -- Action entiies types should be disjoint from `ets`
   env.ets.find? uid.ty = none ∧
```
emina: I think that we might not need this condition anymore, since we're requiring the disjointness of entity/action types for WellFormed environments.

If disjointness is needed for type-soundness proofs, then I suggest changing the type soundness theorem to require the input environment to be well-formed.

cedar-lean/Cedar/Thm/Validation/Typechecker/Types.lean
```diff
@@ -48,42 +49,42 @@ def InstanceOfExtType : Ext → ExtType → Prop
...
-      InstanceOfType (.record r) (.record rty)
+      InstanceOfType env (.record r) (.record rty)
   | instance_of_ext (x : Ext) (xty : ExtType)
       (h₁ : InstanceOfExtType x xty) :
-      InstanceOfType (.ext x) (.ext xty)
+      InstanceOfType env (.ext x) (.ext xty)
 
-def InstanceOfRequestType (request : Request) (reqty : RequestType) : Prop :=
-  InstanceOfEntityType request.principal reqty.principal ∧
+def InstanceOfRequestType (env : Environment) (request : Request) (reqty : RequestType) : Prop :=
```
emina: It seems odd to take `reqty` as input here, when `env` itself fixes the request type.  I suggest changing this definition to :

```
def InstanceOfRequestType (env : Environment) (request : Request) : Prop :=
  InstanceOfEntityType env request.principal env.reqty.principal ∧
  request.action = env.reqty.action ∧
  InstanceOfEntityType env request.resource env.reqty.resource ∧
  InstanceOfType env request.context (.record env.reqty.context)
```

cedar-lean/Cedar/Thm/Validation/Typechecker/Types.lean
```diff
@@ -126,8 +127,8 @@ Each entry in the store is valid
 def InstanceOfEntitySchema (entities : Entities) (env : Environment) : Prop :=
   ∀ (uid : EntityUID) (data : EntityData),
     entities.find? uid = some data →
-    WellFormedEntityData uid data env.ets ∨
-    WellFormedActionData uid data env
+    WellFormedEntityData env uid data ∨
+    WellFormedActionData env uid data
```
emina: I think we should rename `WellFormedEntityData` and `WellFormedActionData` to `InstanceOfEntitySchemaEntry` and `InstanceOfActionSchemaEntry`, for consistency with other names in this module.

I  would also suggest reorganizing the constraints across these definitions so that they can be better understood in isolation, along these lines:

```
def InstanceOfEntitySchemaEntry (env : Environment) (uid : EntityUID) (data : EntityData) : Prop :=
  ∃ entry, env.ets.find? uid.ty = some entry ∧
    InstanceOfType env data.attrs (.record entry.attrs) ∧
    InstanceOfEntityTags env data entry

def InstanceOfActionSchemaEntry (env : Environment) (uid : EntityUID) (data : EntityData) : Prop :=
  data.attrs = .empty ∧
  data.tags = .empty ∧
  ∃ entry, 
    env.acts.find? uid = some entry ∧
    data.ancestors = entry.ancestors

def InstanceOfSchemaEntry (env : Environment) (uid : EntityUID) (data : EntityData) : Prop := 
  InstanceOfEntitySchemaEntry env uid data ∨
  InstanceOfActionSchemaEntry env uid data
  
def InstanceOfEntitySchema  (env : Environment) (entities : Entities) : Prop :=
  (∀ (uid : EntityUID) (data : EntityData), 
    entities.find? uid = some data → InstanceOfSchemaEntry env uid data) ∧
  (∀ (uid : EntityUID) (entry : ActionSchemaEntry),
    env.acts.find? uid = some entry → ∃ data, entities.find? uid = some data)
    
def RequestAndEntitiesMatchEnvironment (env : Environment) (request : Request) (entities : Entities) : Prop :=
  InstanceOfRequestType env request  ∧
  InstanceOfSchema env entities
```

emina (changes requested):

cedar-lean/Cedar/Thm/Validation/Typechecker/Types.lean
zhengyao-lin: sg. they are going to require the well-formedness condition anyway, since `type_is_habited` now depends on it (for non-empty enum entities)

cedar-lean/Cedar/Thm/Validation/Typechecker/Types.lean
```diff
@@ -100,51 +100,44 @@ For every entity `(uid, data)` in the store,
...
-    data.ancestors = entry.ancestors
+def InstanceOfSchema (entities : Entities) (env : Environment) : Prop :=
+  -- Each entity data is valid
+  (∀ (uid : EntityUID) (data : EntityData),
+    entities.find? uid = some data → InstanceOfSchemaEntry uid data env) ∧
+  -- Each action in the schema exists
+  (∀ (uid : EntityUID) (entry : ActionSchemaEntry),
+    env.acts.find? uid = some entry → ∃ data, entities.find? uid = some data)
 
 def RequestAndEntitiesMatchEnvironment (env : Environment) (request : Request) (entities : Entities) : Prop :=
```
emina: I suggest renaming this to the following (to match other functions in this file):

```
def InstanceOfWellFormedEnvironment (request : Request) (entities : Entities) (env : Environment) : Prop :=
  env.WellFormed ∧
  InstanceOfRequestType request env  ∧
  InstanceOfSchema entities env
```

emina (approved):

cedar-lean/Cedar/Thm/Validation/Typechecker/Types.lean
```diff
@@ -48,42 +49,42 @@ def InstanceOfExtType : Ext → ExtType → Prop
...
       (h₁ : InstanceOfExtType x xty) :
-      InstanceOfType (.ext x) (.ext xty)
+      InstanceOfType env (.ext x) (.ext xty)
 
-def InstanceOfRequestType (request : Request) (reqty : RequestType) : Prop :=
-  InstanceOfEntityType request.principal reqty.principal ∧
-  request.action = reqty.action ∧
-  InstanceOfEntityType request.resource reqty.resource ∧
-  InstanceOfType request.context (.record reqty.context)
+def InstanceOfRequestType (request : Request) (env : Environment) : Prop :=
```
cdisselkoen, line 79: maybe this should be renamed now, since it doesn't actually take a `RequestType` parameter anymore?  Something like `ConformsToEnvironment`?

cedar-lean/Cedar/Thm/Validation/Typechecker/Types.lean
```diff
@@ -99,51 +100,44 @@ For every entity `(uid, data)` in the store,
    in the type store.
 4. The entity's tags' types are consistent with the tags information in the type store.
 -/
-def WellFormedEntityData (uid : EntityUID) (data : EntityData) (ets : EntitySchema) : Prop :=
-  ∃ entry, ets.find? uid.ty = some entry ∧
+def InstanceOfEntitySchemaEntry (uid : EntityUID) (data : EntityData) (env : Environment) : Prop :=
```
cdisselkoen, line 103: The name `InstanceOfEntitySchemaEntry` is confusing here, because this does not take an `EntitySchemaEntry` parameter; but `InstanceOfEntityTags` actually _does_ take an `EntitySchemaEntry` parameter

cedar-lean/Cedar/Thm/Validation/Typechecker/Types.lean
```diff
@@ -99,51 +100,44 @@ For every entity `(uid, data)` in the store,
...
+    InstanceOfEntityTags data entry env
 
 /--
 Similar to `WellFormedEntityData`, but a special case for action entities
 since they are stored disjoint from `ets`
 -/
-def WellFormedActionData (uid : EntityUID) (data : EntityData) (env : Environment) : Prop :=
-  -- Action entiies types should be disjoint from `ets`
-  env.ets.find? uid.ty = none ∧
+def InstanceOfActionSchemaEntry (uid : EntityUID) (data : EntityData) (env : Environment) : Prop :=
```
cdisselkoen, line 114: Might make sense for `env` to be the first parameter in all of these definitions?  Curious about others' thoughts

zhengyao-lin: I have no opinion on the order. I believe @emina prefers env at the end, maybe because in some definitions, e.g., `InstanceOfWellFormedEnvironment (request : Request) (entities : Entities) (env : Environment)`, this order reads more naturally?

cdisselkoen: For `InstanceOfWellFormedEnvironment` I can see that.  For others, like `InstanceOfActionSchemaEntry`, I could imagine perhaps wanting to partially-apply to `env` ie that `InstanceOfActionSchemaEntry env` might make sense as an expression, more than `InstanceOfActionSchemaEntry uid`, but this is a very minor point.

emina: Yes, given the `InstanceOf*` naming convention, it seems more natural to me for the instance arguments to come first (so, concrete request and entities) and then the environment (of which they are instance) last. 

I copied the `InstanceOf*` convention from the old Dafny code when I was originally porting these proofs, and I have no objection to adopting a different naming scheme and then making the Environment the first argument.  Maybe `Environment.HasActionSchemaEntryFor (uid : EntityUID) (data : EntityData)` or something like that?

Also, I would really like us to rename `Environment` to `TypeEnv` (maybe after we merge all these changes :))

zhengyao-lin: How about these new names?
- `InstanceOfRequestType => Request.ConformsToEnvironment `
- `InstanceOfActionSchemaEntry => Environment.HasActionSchemaEntryFor`
- `InstanceOfEntitySchemaEntry => Environment.HasEntitySchemaEntryFor`

I'll change the order accordingly.

Can I actually make these changes after #658 #659 and #661 are merged?
Since these new names requires some proof changes, I was hoping to do them in one go instead of separately in these PRs.

zhengyao-lin:
Merged #659 to here.

The only unreviewed changes are:
- A new `ToString` instance in [Cedar/Validation/EnvironmentValidator.lean](https://github.com/cedar-policy/cedar-spec/pull/658/files#diff-e85822ecc48f7b833d52e4e2a6081c90c7a524f6da6ebca1ee6006b7871c0e5d)
    ```
    instance : ToString EnvironmentValidationError where
      toString err :=
        match err with
        | .typeError msg => s!"type error: {msg}"
    ```
- [cedar-lean/CedarFFI/Main.lean](https://github.com/cedar-policy/cedar-spec/pull/658/files#diff-a345ced4ef8739a51a456e4fac59924227e928072b0b524b816887de37d2ec17)

cedar-lean/Cedar/Thm/Validation/Typechecker/Types.lean
emina: Yes, let's do the renamings after we merge.  

To be consistent though, I think all of these functions should be defined on TypeEnv (if we make env the first argument), and in that case, `InstanceOfRequestType ` would be something like `TypeEnv.HasRequest (req : Request).` 

Basically, as long as we're consistent, I'm happy :)

If we make some of these `TypeEnv.Has*`, we should make all of them follow that convention.

(I don't feel strongly about using the word `Has` as the prefix here; `Contains`, `Admits`, etc. are other possiblities.)

shaobo-he-aws: I'm in favor of existing naming scheme (i.e., `InstanceOf*` with instance before environment).

PR #659, opened 2025-07-04
Fix broken Validation/WellTyped/TPE proofs from #658

zhengyao-lin:
This PR is a continuation of #658 (second part of the split #657). It fixes all proofs affected by the spec changes.

There are quite a lot of changes, but they are mostly proof fixes.
The only high-level changes that might require more careful vetting are:
- In type checking and  soundness theorems (e.g., `type_of_is_sound`), the old precondition `RequestAndEntitiesMatchEnvironment` is renamed and stenghtened to `InstanceOfWellFormedEnvironment` (see #658), which in particular enforces a new requirement `env.WellFormed`.
- An executable version of `env.WellFormed` is added to a new file `Cedar.Validation.EnvironmentValidator`
- Validation soundness theorems (e.g., `validation_is_sound`) has a new precondition of `schema.validateWellFormed = .ok ()` now, which essentially checks `env.WellFormed`

Thank you!
(also note that this PR is merging to the branch of #658 instead of main to make the diff nicer)

### DIFF
--- a/cedar-lean/Cedar/Validation.lean
+++ b/cedar-lean/Cedar/Validation.lean
@@ -20,4 +20,5 @@ import Cedar.Validation.Subtyping
 import Cedar.Validation.Typechecker
 import Cedar.Validation.Validator
 import Cedar.Validation.RequestEntityValidator
+import Cedar.Validation.EnvironmentValidator
 import Cedar.Validation.Levels
--- a/cedar-lean/Cedar/Validation/EnvironmentValidator.lean
+++ b/cedar-lean/Cedar/Validation/EnvironmentValidator.lean
@@ -31,6 +31,11 @@ open Cedar.Spec
 inductive EnvironmentValidationError where
 | typeError (msg : String)
 
+instance : ToString EnvironmentValidationError where
+  toString err :=
+    match err with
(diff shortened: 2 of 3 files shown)


shaobo-he-aws (approved):
LGTM

emina (commented):
Very nice PR!

It seems like we're missing the code that calls the new EnvironmentValidator during DRT?  Or maybe I'm not seeing it?  

It also looks like some CI actions are failing. 

I think we can merge once these two issues are fixed.

zhengyao-lin:
@emina Thanks! The [DRT](https://github.com/cedar-policy/cedar-spec/blob/main/cedar-lean/DiffTest/Main.lean) seems to be testing difference pieces of validation separately (policy, request, and entities validations are done separately).  So maybe we shouldn't modify that for now? @chaluli should probably confirm.
In the future, maybe it would be nice to add a fuzzing test that checks if the schema from the Rust side satisfies this new requirement.

I did just add the new schema/environment validation to `parseCheckPolicyReq` in the `cedar-lean-cli`.

As for CI failures, I think it might be due to the fact that I'm merging a non-main branch into a non-main branch. It failed during git checkout.

I'm going to merge this branch into the branch of #658 and see if the CI passes.

PR #662, opened 2025-07-07
[2.5.0] Ignore unexpected Rust validation errors

john-h-kastner-aws:
Our nightly testing intermittently reports a test target failure for the `validation-drt-type-directed` and `strict-validation-drt-type-directed` targets on version 2.5.0. In the failure I've looked at, this is due to the Rust implementation being less precise than the Dafny and reporting an expected error.

We are not concerned with this sort of discrepancy on version 2.5.0, so this PR ignores unexpected Rust validation errors to reduce reduce the noise from our nightly test. Nightly testing will still detect any inputs where Rust _does not_ error when Dafny does, as these could be soundness issues in 2.5.0.

### DIFF
--- a/cedar-drt/fuzz/src/lib.rs
+++ b/cedar-drt/fuzz/src/lib.rs
@@ -116,7 +116,9 @@ impl<'e> DifferentialTester<'e> {
     }
 
     /// Differentially test validation on the given policy and schema.
-    /// Panics if the two engines do not agree.
+    /// Panics if Dafny report an error but Rust does not. The Rust
+    /// implementation is known ot be less precise, so this function does not
+    /// panic if Rust reports an unexpected error.
     pub fn run_validation(
         &self,
         schema: ValidatorSchema,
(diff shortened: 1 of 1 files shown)


cedar-drt/fuzz/src/lib.rs
```diff
@@ -116,7 +116,9 @@ impl<'e> DifferentialTester<'e> {
     }
 
     /// Differentially test validation on the given policy and schema.
-    /// Panics if the two engines do not agree.
+    /// Panics if Dafny report an error but Rust does not. The Rust
+    /// implementation is known ot be less precise, so this function does not
```
cdisselkoen:
```suggestion
    /// Panics if Dafny reports an error but Rust does not. The Rust
    /// implementation is known to be less precise, so this function does not
```

chaluli (approved):

chaluli (approved):

PR #663, opened 2025-07-07
Adds a fix to fuzz target generation for mismatched authorization errors during corpus generation.

chaluli:
Changes fuzz target generation for targets that may dump a cedar-test-json format (e.g., may be used for corpus generation) to ensure that the generated `Entities` always includes actions generated during `Schema` generation.

This fixes an error in which action entities were not considered when calculating the expected authorization decision but were considered when actually performing an integration test (causing an assertion violation during corpus generation).

### DIFF
--- a/cedar-drt/fuzz/fuzz_targets/abac-type-directed.rs
+++ b/cedar-drt/fuzz/fuzz_targets/abac-type-directed.rs
@@ -83,7 +83,7 @@ impl<'a> Arbitrary<'a> for FuzzTargetInput {
                 .map_err(|e| Error::SchemaError(e))?;
         let actions = validator_schema
             .action_entities()
-            .map_err(|_| Error::EntitiesError("Error fetching action entities".into()))?;
+            .map_err(|e| Error::EntitiesError(format!("Error fetching action entities: {e}")))?;
         let entities = entities
             .add_entities(
                 actions.into_iter().map(|e| std::sync::Arc::new(e)),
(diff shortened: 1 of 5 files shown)


cedar-drt/fuzz/fuzz_targets/abac-type-directed.rs
```diff
@@ -78,6 +78,20 @@ impl<'a> Arbitrary<'a> for FuzzTargetInput {
...
+            .action_entities()
+            .map_err(|_| Error::EntitiesError("Error fetching action entities".into()))?;
+        let entities = entities
+            .add_entities(
+                actions.into_iter().map(|e| std::sync::Arc::new(e)),
+                None::<&cedar_policy_core::validator::CoreSchema>,
+                cedar_policy_core::entities::TCComputation::AssumeAlreadyComputed,
+                cedar_policy_core::extensions::Extensions::all_available(),
+            )
+            .map_err(|_| Error::EntitiesError("Error adding action entities to entities".into()))?;
```
cdisselkoen: would probably be helpful to include the text of the underlying error, something like
```suggestion
            .map_err(|e| Error::EntitiesError(format!("Error adding action entities to entities: {e}")))?;
```

cedar-drt/fuzz/fuzz_targets/entity-slicing-drt-type-directed.rs
```diff
@@ -76,9 +77,26 @@ impl<'a> Arbitrary<'a> for FuzzTargetInput {
...
+            .action_entities()
+            .map_err(|_| Error::EntitiesError("Error fetching action entities".into()))?;
+        let entities = entities
+            .add_entities(
+                actions.into_iter().map(|e| std::sync::Arc::new(e)),
+                None::<&cedar_policy_core::validator::CoreSchema>,
+                cedar_policy_core::entities::TCComputation::AssumeAlreadyComputed,
+                cedar_policy_core::extensions::Extensions::all_available(),
+            )
+            .map_err(|_| Error::EntitiesError("Error adding action entities to entities".into()))?;
```
cdisselkoen: same here, and above (line 87)

cdisselkoen (approved):
Looks good.  Would be better if we could share this code a little more

shaobo-he-aws (approved):
LGTM.

QQ: Why don't we make this functionality an API in core?

cdisselkoen (approved):

Issue #654, opened 2025-07-01
abac-type-directed fails at dump


LLLLLQ-4:
### Before opening, please confirm:

- [x] I have [searched for duplicate or closed issues](https://github.com/cedar-policy/cedar/issues?q=is%3Aissue+).
- [x] I have read the guide for [submitting bug reports](https://github.com/cedar-policy/cedar/blob/main/CONTRIBUTING.md#bug-reports).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### Bug Category

Other

### Describe the bug

When I run fuzzing targeting at abac-type-directed, the process fails at dump.
The command is as follows:
```
FUZZ_TARGET="abac-type-directed" ./create_corpus.sh
```

And the error messages as follows:
```
./fuzz/target/x86_64-unknown-linux-gnu/release/abac-type-directed: Running 1 inputs 1 time(s) each.
Running: fuzz/corpus/abac-type-directed/085b30698714151cecfa15e9d75fc2464c269c19

thread '<unnamed>' panicked at src/dump.rs:235:53:
called `Result::unwrap()` on an `Err` value: ExtnCall2OrMoreArguments(ExtnCall2OrMoreArguments { func: Name(InternalName { id: Id("offset"), path: [], loc: None }) })
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
==48367== ERROR: libFuzzer: deadly signal
NOTE: libFuzzer has rudimentary signal handlers.
      Combine libFuzzer with AddressSanitizer or similar for better crash reports.
SUMMARY: libFuzzer: deadly signal
```

### Expected behavior

The dump process should run without any error.

### Reproduction steps

run 
```
FUZZ_TARGET="abac-type-directed" ./create_corpus.sh
```

### Code Snippet

```rust
// Put your code below this line.

```

### Log output

```
// Put your output below this line

```

PR #665, opened 2025-07-07
Ignore dumping of corpus tests if it contains an entity or context including the offset external function.

chaluli:
This PR addresses issue #654 .

The root cause of the issue in #654 is that we do not allow representing datetimes before year 0 using the `datetime` constructor. They must be constructed using the `offset` function to combine a `datetime` and a `duration`. However, we do not allow serialization of such expressions.

This PR avoids the need to serialize such expressions by ignoring test cases the include offsets.

### DIFF
--- a/cedar-drt/fuzz/src/dump.rs
+++ b/cedar-drt/fuzz/src/dump.rs
@@ -17,7 +17,7 @@
 use cedar_policy::{AuthorizationError, Policy};
 use cedar_policy_core::ast::{Context, EntityUID, EntityUIDEntry, PolicySet, Request};
 use cedar_policy_core::authorizer::Response;
-use cedar_policy_core::entities;
+use cedar_policy_core::entities::{self, err::EntitiesError, json::err::JsonSerializationError};
 use cedar_policy_core::entities::{Entities, TypeAndId};
 use cedar_policy_core::extensions::Extensions;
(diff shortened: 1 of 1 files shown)


cedar-drt/fuzz/src/dump.rs
```diff
@@ -62,13 +62,31 @@ pub fn dump(
...
+        .open(&entities_filename)?;
+    match entities.write_to_json(entities_file) {
+        Ok(()) => (),
+        Err(EntitiesError::Serialization(JsonSerializationError::ExtnCall2OrMoreArguments(e)))
+            if e.to_string().contains("offset") =>
+        {
+            return Ok(());
+        }
+        Err(e) => panic!("Unexpected error while serializing entities: {e}"),
+    }
```
cdisselkoen, line 79: so we simply don't dump any testcases from the corpus that happen to include `offset()` in an entity attribute?

chaluli: Correct. It's a solution to stop the corpus generation from failing until we find a longer term solution to serializing external functions that take more than 1 argument.

katherine-hough: Could you add a comment explaining this and open an issue requesting a more permanent solution?

chaluli: We are discussing offline if instead of this PR updating cedar to do the work for the longer term solution, now.

PR #666, opened 2025-07-07
Add an action to test corpus generatrion for both abac and abac-type-directed fuzz-targets.

chaluli:
Adds actions to CI to ensure that changes to cedar-spec do not cause corpus generation to fail.

The action performs corpus generation (using the default of 15 minutes for fuzzing) for both the `abac` and `abac-type-directed` fuzzing targets. Currently these are the only targets that can be used to generate a testing corpus (i.e., these are the only fuzz-targets that include a call to `dump`).

The new CI actions are unlikely to pass until PRs #663 and #665 are merged.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,6 +78,13 @@ jobs:
       cedar_spec_ref: ${{ github.ref }}
       cedar_policy_ref: ${{ needs.get-branch-name.outputs.branch_name }}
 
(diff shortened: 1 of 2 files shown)


cdisselkoen (approved):
For CI purposes, perhaps we should do 5 min instead of the default 15, to reduce CI latency and also conserve CI resources?

adpaco-aws (approved):

PR #667, opened 2025-07-08
[2.5.x] Avoid repeatedly parsing trivial policies in rbac-authorizer target

john-h-kastner-aws:
See #664

### DIFF
--- a/cedar-drt/fuzz/fuzz_targets/rbac-authorizer.rs
+++ b/cedar-drt/fuzz/fuzz_targets/rbac-authorizer.rs
@@ -15,12 +15,12 @@
  */
 
 #![no_main]
+use cedar_drt::ast::PolicyID;
 use cedar_drt::*;
 use cedar_drt_inner::fuzz_target;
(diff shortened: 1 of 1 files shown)


cedar-drt/fuzz/fuzz_targets/rbac-authorizer.rs
```diff
@@ -45,38 +44,57 @@ enum AbstractPolicy {
...
-                Some(id),
-                "forbid(principal, action, resource) when { 1 == 0 };",
-            )
-            .expect("should be a valid policy"),
-            AbstractPolicy::ForbidError => parser::parse_policy(
-                Some(id),
-                "forbid(principal, action, resource) when { 1 < \"hello\" };",
-            )
-            .expect("should be a valid policy"),
+            AbstractPolicy::PermitTrue => concrete_policies::PERMIT_TRUE.new_id(id),
```
chaluli: It looks like you need to cast the `id` to `PolicyID` before giving as an argument to `new_id`

PR #668, opened 2025-07-08
 [3.4.x] Avoid repeatedly parsing trivial policies in rbac-authorizer target

john-h-kastner-aws:
See #664

### DIFF
--- a/cedar-drt/fuzz/fuzz_targets/rbac-authorizer.rs
+++ b/cedar-drt/fuzz/fuzz_targets/rbac-authorizer.rs
@@ -15,12 +15,12 @@
  */
 
 #![no_main]
+use cedar_drt::ast::PolicyID;
 use cedar_drt::*;
 use cedar_drt_inner::*;
 use cedar_policy_core::ast;
 use cedar_policy_core::entities::Entities;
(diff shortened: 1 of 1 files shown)


cdisselkoen (approved):

cedar-drt/fuzz/fuzz_targets/rbac-authorizer.rs
```diff
@@ -46,38 +45,57 @@ enum AbstractPolicy {
...
-                Some(id),
-                "forbid(principal, action, resource) when { 1 == 0 };",
-            )
-            .expect("should be a valid policy"),
-            AbstractPolicy::ForbidError => parser::parse_policy(
-                Some(id),
-                "forbid(principal, action, resource) when { 1 < \"hello\" };",
-            )
-            .expect("should be a valid policy"),
+            AbstractPolicy::PermitTrue => concrete_policies::PERMIT_TRUE.new_id(id),
```
chaluli, line 94: It seems that this version of cedar doesn't allow for `String`s or other elements to be used as IDs in `new_id`. We need to construct a PolicyID from the string representation.

PR #671, opened 2025-07-10
Update cedar-drt to use Cedar API types when possible + Refactor to use `cedar-lean-ffi` for Rust-to-Lean Bindings.

chaluli:
1.) Updates `cedar-drt` to use Cedar API (public) types instead of Cedar AST (intertnal) types when possible.
2.) Updates `cedar-drt` to use `cedar-lean-ffi`for Rust-to-Lean communication.

This PR depends on PR [#1672](https://github.com/cedar-policy/cedar/pull/1672) of cedar which refactors `cedar-testing` to use Cedar API (public) types.

These PRs should be merged at the same time to ensure neither causes breaking changes.


chaluli:
I expect the `run-integration-tests` CI to fail as there is not corresponding `cdisselkoen/cedar-test-public` branch of the integration-tests repo.

I also expect the DRT Docker Build CI to fail as the DRT changes will not build against the `main` branch of cedar (but is based on the corresponding `cdisselkoen/cedar-test-public` branch of cedar-policy).

cedar-drt/Cargo.toml
```diff
@@ -1,37 +1,32 @@
...
 edition = "2021"
-version = "4.0.0"
+version = "4.4.0"
 publish = false
 
 [dependencies]
+libfuzzer-sys = "0.4"
+cedar-lean-ffi = { path = "../cedar-lean-ffi", version = "4.4.0" }
+cedar-policy = { version = "4.4.0" }
+cedar-policy-core = { path = "../cedar/cedar-policy-core", version = "4.4.0", features = ["arbitrary", "entity-manifest"] }
```
cdisselkoen, line 11: why take cedar-policy from crates but cedar-policy-core from the path dependency?  could we take both from crates if we locked the versions to match (with an `=` constraint)?

shaobo-he-aws (approved):

cedar-drt/Cargo.toml
cdisselkoen: oh, I see that we are using `[patch.crates-io]` below.  So then my comment becomes, why use `patch` for cedar-policy but not for cedar-policy-core?

cedar-drt/fuzz/Cargo.toml
```diff
@@ -1,43 +1,37 @@
...
-cedar-drt = { version = "4.0.0", path = ".." }
-cedar-policy = { path = "../../cedar/cedar-policy", version = "4.*", features = ["protobufs"] }
-cedar-policy-core = { path = "../../cedar/cedar-policy-core", version = "4.*", features = ["arbitrary", "entity-manifest"] }
-cedar-policy-formatter = { path = "../../cedar/cedar-policy-formatter", version = "4.*" }
-cedar-testing = { path = "../../cedar/cedar-testing", version = "4.*" }
-cedar-policy-generators = { path = "../../cedar-policy-generators", version = "4.*" }
+cedar-drt = { version = "4.4.0", path = ".." }
+cedar-lean-ffi = { path = "../../cedar-lean-ffi", version = "4.4.0" }
+cedar-policy = { version = "4.4.0", features = ["permissive-validate", "entity-manifest"] }
+cedar-policy-core = { path = "../../cedar/cedar-policy-core", version = "4.4.0", features = ["arbitrary"] }
```
cdisselkoen, line 15: same question in this crate

cedar-drt/build_lean_lib.sh
cdisselkoen, line 1: maybe add `#!/bin/sh` to this file?

cedar-drt/fuzz/Cargo.toml
```diff
@@ -1,43 +1,37 @@
...
 miette = "7.1.0"
-smol_str = { version = "0.3", features = ["serde"] }
-regex = "1"
-rayon = { version = "1.5", optional = true }
+prost = "0.13"
 rand = { version = "0.9", features = ["small_rng"] }
-clap = { version = "4.0", features = ["derive"], optional = true }
-rand_chacha = { version = "0.9", optional = true }
+regex = "1"
+serde = { version = "1.0" }
```
cdisselkoen, line 26: nit: this can be written `serde = "1.0"`

cedar-drt/fuzz/Cargo.toml
```diff
@@ -1,43 +1,37 @@
...
-prost = "0.13"
-itertools = "0.14.0"
-
-[dependencies.uuid]
-version = "1.3.1"
-features = ["v4", "fast-rng"]
+uuid = { version = "1.3.1", features = ["v4", "fast-rng"] }
 
-[features]
-prt = ["dep:rayon", "dep:clap", "dep:rand_chacha"]
```
cdisselkoen, line 40: are we removing the `prt` feature? what was it used for?

cedar-drt/fuzz/fuzz_targets/entity-slicing-drt-type-directed.rs
```diff
@@ -77,28 +76,9 @@ impl<'a> Arbitrary<'a> for FuzzTargetInput {
...
-        let entities = entities
-            .add_entities(
-                actions.into_iter().map(|e| std::sync::Arc::new(e)),
-                None::<&cedar_policy_core::validator::CoreSchema>,
-                cedar_policy_core::entities::TCComputation::AssumeAlreadyComputed,
-                cedar_policy_core::extensions::Extensions::all_available(),
-            )
-            .map_err(|e| {
-                Error::EntitiesError(format!("Error adding action entities to entities: {e}"))
-            })?;
```
cdisselkoen, line 97: do we need an `add_actions_to_entities` call like in the other targets?

cedar-drt/fuzz/fuzz_targets/eval-type-directed.rs
```diff
@@ -73,23 +79,6 @@ impl<'a> Arbitrary<'a> for FuzzTargetInput {
...
-            .add_entities(
-                actions.into_iter().map(|e| std::sync::Arc::new(e)),
-                None::<&cedar_policy_core::validator::CoreSchema>,
-                cedar_policy_core::entities::TCComputation::AssumeAlreadyComputed,
-                cedar_policy_core::extensions::Extensions::all_available(),
-            )
-            .map_err(|e| {
-                Error::EntitiesError(format!("Error adding action entities to entities: {e}"))
-            })?;
-
```
cdisselkoen, line 92: same question here

cedar-drt/fuzz/fuzz_targets/eval-type-directed.rs
```diff
@@ -120,15 +109,14 @@ impl<'a> Arbitrary<'a> for FuzzTargetInput {
 // Type-directed fuzzing of expression evaluation.
 fuzz_target!(|input: FuzzTargetInput| {
     initialize_log();
-    let def_impl = LeanDefinitionalEngine::new();
+    let def_impl = CedarLeanEngine::new();
     debug!("Schema: {}\n", input.schema.schemafile_string());
     debug!("expr: {}\n", input.expression);
-    debug!("Entities: {}\n", input.entities);
+    debug!("Entities: {}\n", input.entities.as_ref());
```
cdisselkoen, line 115: nit: somewhere else we switched to `{:?}` instead.  This is probably better?  we should probably be consistent though

cedar-drt/fuzz/fuzz_targets/request-validation.rs
```diff
@@ -76,43 +83,37 @@ impl<'a> Arbitrary<'a> for FuzzTargetInput {
...
+            schema::Schema::arbitrary_request_size_hint(depth),
         ]))
     }
 }
 
 // Non-type-directed fuzzing of (strict) validation.
 fuzz_target!(|input: FuzzTargetInput| {
     initialize_log();
-    let def_impl = LeanDefinitionalEngine::new();
+    let def_impl = CedarLeanEngine::new();
```
cdisselkoen, line 103: like you did in rbac.rs, we could shift this inside the `if` for some performance savings

cedar-drt/fuzz/fuzz_targets/validation-pbt.rs
```diff
@@ -165,46 +166,52 @@ fn log_err<T>(res: Result<T>, doing_what: &str) -> Result<T> {
...
+fn maybe_log_schemastats(schema: Option<&Schema>, suffix: &str) {
     if std::env::var("FUZZ_LOG_STATS").is_ok() {
         let schema = schema.expect("should be SOME if FUZZ_LOG_STATS is ok");
         checkpoint(
             LOG_FILENAME_TOTAL_ENTITY_TYPES.to_string()
                 + "_"
                 + suffix
                 + "_"
-                + &format!("{:03}", schema.entity_types.len()),
+                + &format!("{:03}", schema.entity_types().collect_vec().len()),
```
cdisselkoen, line 177: we can avoid collecting into a vec by replacing `.collect_vec().len()` with `.count()`

cedar-drt/fuzz/fuzz_targets/validation-pbt.rs
```diff
@@ -165,46 +166,52 @@ fn log_err<T>(res: Result<T>, doing_what: &str) -> Result<T> {
...
         );
         checkpoint(
             LOG_FILENAME_TOTAL_ACTIONS.to_string()
                 + "_"
                 + suffix
                 + "_"
-                + &format!("{:03}", schema.actions.len()),
+                + &format!("{:03}", schema.actions().collect_vec().len()),
```
cdisselkoen, line 184: same here

cedar-drt/fuzz/fuzz_targets/validation-pbt.rs
```diff
@@ -165,46 +166,52 @@ fn log_err<T>(res: Result<T>, doing_what: &str) -> Result<T> {
...
-                            + "_"
-                            + &format!("{:03}", resource_types.len()),
-                    );
-                }
+        for action in schema.actions() {
+            let n_principals = schema
+                .principals_for_action(action)
+                .unwrap()
+                .collect_vec()
+                .len();
```
cdisselkoen, line 191: and here, and below

cedar-drt/fuzz/src/lib.rs
cdisselkoen, line 1: nice -- basically we don't need any of this code anymore because we replaced it with a dependency on cedar-lean-ffi?

cedar-drt/set_env_vars.sh
cdisselkoen, line 1: same comment about `#!/bin/sh`

cedar-drt/src/definitional_request_types.rs
cdisselkoen, line 1: nice, presumably versions of all of these are in cedar-lean-ffi now too?

cedar-drt/src/lib.rs
```diff
@@ -1,23 +1,13 @@
...
-
-pub use definitional_request_types::*;
-pub use lean_impl::*;
-pub use logger::*;
+pub use lean_engine::CedarLeanEngine;
+pub use parsing_utils::{
+    check_for_internal_errors, check_policy_equivalence, check_policy_set_equivalence,
+    policy_set_to_text,
+};
+pub use prt::fuzz_target;
```
cdisselkoen, line 13: does this mean all of our targets are now using this `fuzz_target` macro instead of the `lean_sys` one?  what's the difference?

cdisselkoen (approved):
Looks great, minor comments

cedar-drt/Cargo.toml
chaluli: Good question. we could do for both. But it's only necessary for `cedar-policy`.

The problem the patch is doing is replacing all dependencies of `cedar-drt` that also use `cedar-policy` from crates.io (namely `cedar-lean-ffi`) to use the same local path for `cedar-policy`.

cedar-drt/build_lean_lib.sh
chaluli: This is a symlink to `cedar-lean-ffi/build_lean_lib.sh` which is for convenience of not having to update other packages that have scripts to auto-build `cedar-drt`.

cedar-drt/fuzz/fuzz_targets/entity-slicing-drt-type-directed.rs
chaluli: We could. But not needed. it is required in `abac` and `abac-type-directed` as both use the `dump` which requires the pre and post-parsed responses agree (and the post-parse will always add the schema's actions) but that doesn't apply for this target.

cedar-drt/fuzz/fuzz_targets/eval-type-directed.rs
chaluli: We could. But not needed. it is required in `abac` and `abac-type-directed` as both use the `dump` which requires the pre and post-parsed responses agree (and the post-parse will always add the schema's actions) but that doesn't apply for this target.

cedar-drt/fuzz/fuzz_targets/eval-type-directed.rs
chaluli: Agreed. I think this uses a friendlier view so I'll replace the other with display formatter of as_ref().

cedar-drt/fuzz/src/lib.rs
chaluli: Some of it was moved to the outer crate in a "tests.rs" for running tests. But a lot of it is no longer necessary.

cedar-drt/set_env_vars.sh
chaluli: This is a symlink to cedar-lean-ffi/build_lean_lib.sh which is for convenience of not having to update other packages that have scripts to auto-build cedar-drt.

cedar-drt/src/definitional_request_types.rs
chaluli: Yes, this entire file is no longer needed due to dependency on `cedar-lean-ffi`.

cedar-drt/fuzz/Cargo.toml
chaluli: The feature should be added back and depend on the parent crate's `prt` feature.

The prt feature controls if the fuzz_target is simply `lib_fuzz`'s fuzz_target or our own implementation that replaces it with an actual main function. Not sure exactly what it's doing in the `prt` feature case.

PR #677, opened 2025-07-11
Simple Entity Slicing using Straight Line Expr Sets

oflatt:
*Issue #, if available:*

*Description of changes:*

The end-goal is to prove that entity manifests and slicing using the SimpleEntityLoader is sound (produces the same result as without slicing).
This PR is a step towards that goal, adding a much simpler version of slicing on cedar expressions.

This PR adds Straight Line Expr (SLExpr) and a entity slicing based on sets of SLExpr.
The idea is that entity slicing is simpler and easier to prove correct as an analysis that gathers expressions without if statements.  In the future, we could prove this implementation equivalent to the more efficient one implemented in rust.

The `all_sl_exprs` function analyzes an expression and gets all the SLExprs.
The `simple_slice_sl` function performs entity slicing by loading entire entities. In the future, we could make this more precise by loading only parts of entities that are needed.


cedar-lean/Cedar/Data/List.lean
```diff
@@ -116,4 +116,18 @@ def isSorted {α} [LT α] [DecidableLT α] (l : List α) : Bool :=
...
+/--
+Performs a cartesian product l1 X l2 X l3 ...
+for every li in L
+-/
+def cartesianProduct {α} (L : List (List α)) : List (List α) :=
+  match L with
+  | [] => [[]]
+  | [l] => l.map (fun e => [e])
+  | l :: Lrest =>
+    let rest_product := cartesianProduct Lrest
```
emina, line 128: Let's rename `rest_product` to `rest` or something like that.

The Lean naming convention is to use snake_case only for theorem names (or, more generally to name things of type `Prop`).  For everything else, we use camelCase.

See here for details: https://github.com/cedar-policy/cedar-spec/blob/main/cedar-lean/GUIDE.md

Let's change all variable / function names in the PR to match the naming convention.

cedar-lean/Cedar/Spec/EntityManifest.lean
```diff
@@ -0,0 +1,367 @@
...
+import Std.Data
+
+namespace Cedar.Spec
+
+
+open Cedar.Data
+open Error
+
+/--
+A SLExpr or Straight Line Expr is like a Cedar Expr, but without if statements.
```
emina, line 37: `if statements` --> `if expressions`

Why can't we keep `if` expressions and use `Expr` as part of the manifest encoding?

The SLExpr translation will result in exponential blow-up if there are diamonds in the CFG.

That's probably ok in the case of Cedar since `if` expressions aren't used that often, but it would be good to see if we can avoid this. 

Another advantage of keeping `Expr` is that we don't have to define an evaluator for `SLExpr`.

cedar-lean/Cedar/Spec/EntityManifest.lean
```diff
@@ -0,0 +1,367 @@
...
+
+instance : Coe Value (SLResult (Data.Set Value)) where
+  coe v := v.asSet.toSLResult
+
+
+/--
+Converts an expr to a set of straight line exprs,
+exploring all possibilities when an if statement is encountered.
+-/
+def all_sl_exprs (expr: Expr) : SLExprs :=
```
emina, line 177: See naming convention comment above :)

Let's rename this to `Expr.toSLExprs`.  Note that in Lean, this name allows us to use the receiver syntax for calling this function:  `x.toSLExprs`.

cedar-lean/Cedar/Spec/EntityManifest.lean
```diff
@@ -0,0 +1,367 @@
...
+  match expr with
+  | .ite cond then_expr else_expr =>
+    let cond_exprs := all_sl_exprs cond
+    let then_exprs := all_sl_exprs then_expr
+    let else_exprs := all_sl_exprs else_expr
+
+    let then_with_cond := List.productTR cond_exprs.toList then_exprs.toList
+    let else_with_cond := List.productTR cond_exprs.toList else_exprs.toList
+
+    let then_results := then_with_cond.map (fun pair => .assertTrue pair.1 pair.2)
```
emina, line 187: We tend to use `λ` for functions because it helps us avoid extra parentheses:  

```
let trs := tc.map λ (fst, snd) => .assertTrue fst snd
```

cedar-lean/Cedar/Spec/EntityManifest.lean
```diff
@@ -0,0 +1,367 @@
...
+    rename_i h
+    have h2 := List.sizeOf_lt_of_mem h
+    omega
+
+
+
+
+
+-- Like evaluate but returns None if any asserts failed
+def evaluate_sl(x : SLExpr) (req : Request) (es : Entities) : SLResult Value :=
```
emina, line 260: Rename to `evaluateSLExpr`.

cedar-lean/Cedar/Spec/EntityManifest.lean
```diff
@@ -0,0 +1,367 @@
...
+    let expr_lists := map.map (fun pair => Set.toList (all_sl_exprs pair.2))
+    let attrs := map.map (fun pair => pair.1)
+    let all_combinations := List.cartesianProduct expr_lists
+    .mk (all_combinations.map (fun combo =>
+      .record (List.zipWith (fun attr expr => (attr, expr)) attrs combo)))
+  | .call xfn args =>
+    let args_exprs := args.map (fun e => (all_sl_exprs e).toList)
+    let all_combinations := List.cartesianProduct args_exprs
+    .mk (all_combinations.map (fun combo => .call xfn combo))
+termination_by sizeOf expr
```
emina, line 230: We can often avoid these termination proofs by using variants of `map` that attach a proof to each element that it's in the List.  See how we use `.mapM₁` for sets in the evaluator:  https://github.com/cedar-policy/cedar-spec/blob/main/cedar-lean/Cedar/Spec/Evaluator.lean

For the non-monadic maps, we can just write `xs.attach.map ...`.  Here, `attach` attaches a proof to each element `x` in the list that says `x \in xs`.

We have several versions of attach that attach different kinds of proof terms that often come up in these recursive functions.  

See https://github.com/cedar-policy/cedar-spec/blob/main/cedar-lean/Cedar/Data/List.lean

cedar-lean/Cedar/Spec/EntityManifest.lean
```diff
@@ -0,0 +1,367 @@
...
+    omega
+  -- Call
+  · simp [*]
+    rename_i h
+    have h2 := List.sizeOf_lt_of_mem h
+    omega
+
+
+
+-- evaluate the SLExpr, gathing all entity ids required
```
emina, line 345: Ah, I see why this PR introduces `SLExpr`s.  It's trying to compute the most precise slice possible in the presence of `if` expressions.

How about we just give on precision in the case of `if`s (because they come up rarely) and just conservatively include all subexpressions from both branches of the `if` (as well as the condition)?

Basically, collect every subexpression that has an entity type.  Note that this would require us to work on `TypedExpr` (https://github.com/cedar-policy/cedar-spec/blob/main/cedar-lean/Cedar/Validation/TypedExpr.lean) rather than `Expr`.  That's okay though:  we can say that using entity manifests requires a schema.

That way, we won't have to introduce `SLExpr` and the precision is going to be the same for most policies.

oflatt: I introduced `SLExpr` because the end goal is to compute an `EntityManifest` DAG, which doesn't support things like if expressions. I'm hoping the DAG is straightforward to compute from the SLExprs. We'll also drop things like unary and binary operators, and handle corners of the cedar semantics like equivalence between structs correctly.

My plan is to prove that this implementation is equivalent to the more efficient one we have implemented in rust. That way, users can load entities in batches based on the dag. 

We could totally collect every entity-typed subexpression, but then we'd be handing the user raw cedar expressions and hoping they evaluate them correctly over their own entity store. 

Maybe there's an easier path to entity manifests than introducing `SLExpr`s that I don't see, though.

oflatt: Here's an example:

```
(if(principal.isAdmin) { principal.fielda } else { principal.fieldb }).name
```

Results in three SLExprs:
```
assertTrue(principal.isAdmin, principal.fielda).name
assertFalse(principal.isAdmin, principal.fielda).name
```

Now the user of entity slicing doesn't need to worry about carefully collecting both of these paths, since the if statement could result in multiple entities.

oflatt: (The rust implementation also results in exponential blowup in the number of if statements)

emina: Maybe I don't have a sufficiently clear view of how this gets us to the final proof that we want :) 

Let me try to explain where my thinking is and we can see if how far / close that is from the final goal.

Abstractly, the manifest DAG connects a set of access path expressions, so that if there is an edge from p1 to p2, then the entity to which p1 evaluates must be loaded before the entity to which p2 evaluates.

We can view all entity slicing algorithms as procedures that make this DAG more or less precise in two dimensions:  the set of nodes it includes, and the set of edges it includes.  

For example, we can view level-based checking as implicitly constructing a DAG where the roots are the principal, resource, and all context paths that end in an entity.  Then, the nodes are the full set of access paths of length <= k that can be reached from these roots by traversing all declared entity-valued attributes.  The edges indicate the loading path, so `principal -> principal.friend`.  

Now, the existing Rust code makes this DAG much more precise by analyzing the policies to (1) include fewer nodes and (2) fewer edges between nodes, in a way that still remains sound (we don't include too few).  This algorithm is pretty complicated so we want to define and prove correct a simpler one to start with.

And this is where I'm struggling to make the connection between this SLExpr algorithm and the final goal.  And also, I'm struggling to see why an algorithm for collecting access paths would need to be exponential in the presence of `if` expressions.

A overapproximate / naive version of this collection algorithm would look something like this:

```
inductive AccessPath where
  | lit  : EntityUID → AccessPath
  | var  : Var → AccessPath
  | path : AccessPath → Attr → AccessPath
deriving Repr, DecidableEq

instance : LT AccessPath where
  lt := sorry

instance AccessPath.decLt (x y : AccessPath) : Decidable (x < y) :=
  sorry

def entityAccessPaths (tx : TypedExpr) : Set AccessPath :=
  match tx with
  | .lit (.entityUID uid) _ => Set.singleton (.lit uid)
  | .lit _ _                => Set.empty
  | .var v _                => Set.singleton (.var v)
  | .ite tx₁ tx₂ tx₃ _ =>
    (entityAccessPaths tx₁) ∪ (entityAccessPaths tx₂) ∪ (entityAccessPaths tx₃)
  | .and tx₁ tx₂ _
  | .or tx₁ tx₂ _
  | .binaryApp _ tx₁ tx₂ _ => (entityAccessPaths tx₁) ∪ (entityAccessPaths tx₂)
  | .hasAttr tx₁ _ _
  | .unaryApp _ tx₁ _      => entityAccessPaths tx₁
  | .getAttr tx₁ a ty      =>
    let ps := entityAccessPaths tx₁
    match ty with
    | .entity _
    | .record _ => Set.make (ps.elts.map (AccessPath.path · a))
    | _         => ps
  | .call _ xs _
  | .set xs _      => xs.attach.mapUnion (λ ⟨xᵢ, _⟩ => entityAccessPaths xᵢ)
  | .record axs _  => axs.attach₂.mapUnion (λ ⟨(_, xᵢ), _⟩ => entityAccessPaths xᵢ)
decreasing_by all_goals sorry
```

Here, an AccessPath is either an entity literal, variable, or a `.path p a` expression, and the type of an AccessPath is either an entity or record type.  

The size of the resulting set is linear in the size of the input expression.   

We should be able to prove something like this about the above algorithm:

Consider the set `P` of all paths `accessPaths tx` and their subpaths, where each path in `P` has an entity type.  Suppose that we have a request `r` and entity store `es`. Suppose that we have an entity slice `es'` (doesn't matter how it's computed) that maps each entity identified by a path in `P` (with respect to `r`) to all of its attributes, ancestors, and tags in `es`.  Then, evaluating `tx` against `r` and `es'` gives the same result as evaluating `tx` against `r` and `es`.

The next step then would be to show how to load the slice via a BFS of the entity manifest DAG.  (My understanding is that this is the simple entity loader does.)

And then eventually we could consider more efficient `entityAccessPaths` algorithms that gather finer-grained access paths to load individual attributes etc.

What am I missing? :)

emina: When I say "entity manifest DAG" above, I mean the DAG that we could re/construct from a set of access paths.

emina: (I didn't handle record literals correctly above but it can be done :)

oflatt: We'll chat about this in person and I'll try to explain here for future reference. I realized the complexity analysis is not easy!

# Complexity Analysis

First, Cedar doesn't have diamonds in the CFG (there are no let statements, so expressions are trees).

Your code is fairly similar to what is implemented in rust, and has complexity O(n^2) in the size of the program (ignoring `getTag`).
The O(n^2) case happens when there are lots of if statements then lots of accesses:
```
(if true
   { if true { principal.a  } else { principal.b } }
   { if true { principal.c } else { principal.d } })
  .attr.attr.attr.attr
```

But if you consider `getTag` the complexity becomes exponential.
```
principal.getTag (if true {principal.a} {principal.b}).getTag(if true {principal.a } {principal.b}) ...
```

So the complexity is `O(2^n)`. We could probably come up with a bound in the number of `getTag`.
The algorithm I have implemented in this PR has the same complexity.



# Why use SLExpr instead of AccessPath?

I'm hoping that SLExpr makes the proofs easier than using AccessPath.
See this file branch: https://github.com/cedar-policy/cedar-spec/blob/c76d0634e15cebd2e416d4d4ea4216d6073bb83f/cedar-lean/Cedar/Thm/EntityManifest.lean#L240

emina: Eeek, yes, if we have to handle tags precisely, we're in trouble.  And if we don't handle them precisely, we may end up loading too much data.

cdisselkoen: Jumping in here to leave my two cents (without reading the technical discussion above in too much detail):

I think it's OK to sacrifice some precision in the case of tags having entity type, as I imagine that's an uncommon case.

In general, I also think it's OK for the first release of entity slicing/manifests to sacrifice precision on a number of axes in order to create simpler/faster implementation or simplify proofs.

oflatt:
Closing in favor of https://github.com/cedar-policy/cedar-spec/pull/681

PR #678, opened 2025-07-15
PBT on TPE

shaobo-he-aws:
*Description of changes:*

This PR introduces a PBT on the (restricted form of) soundness property of TPE. That is, evaluating an original policy produces the same outcome as evaluating its TPE-derived residual - whether that's an identical value on successful evaluation or errors when evaluation fails. This target only makes the request unknown but not the entity store, which will be done in the future.

### DIFF
--- a/cedar-drt/fuzz/fuzz_targets/tpe-pbt.rs
+++ b/cedar-drt/fuzz/fuzz_targets/tpe-pbt.rs
@@ -16,7 +16,7 @@
 
 #![no_main]
 use cedar_drt::logger::initialize_log;
-use cedar_drt_inner::*;
+use cedar_drt_inner::fuzz_target;
 use cedar_policy_core::ast;
 use cedar_policy_core::ast::RequestSchema;
 use cedar_policy_core::ast::{EntityUID, Expr, Request, Value};
@@ -213,15 +213,18 @@ fn entities_to_partial_entities<'a>(
     ))
 }
 
-fn test_weak_equiv(residual: Residual, e: &Expr, req: Request, entities: &Entities) -> bool {
-    let eval = Evaluator::new(req, entities, Extensions::all_available());
+fn test_weak_equiv(residual: Residual, e: &Expr, req: &Request, entities: &Entities) -> bool {
+    let eval = Evaluator::new(req.clone(), entities, Extensions::all_available());
     let slots = HashMap::new();
 
     let expr = Expr::from(residual);
+    debug!("request: {req}");
     debug!("expr: {e}");
     debug!("residual: {expr}");
     let concrete_res = eval.interpret(e, &slots);
     let reeval_res = eval.interpret(&expr, &slots);
+    debug!("concrete evaluation result: {concrete_res:?}");
+    debug!("re-evaluation result: {reeval_res:?}");
     concrete_res.ok() == reeval_res.ok()
 }
 
@@ -247,9 +250,10 @@ fuzz_target!(|input: FuzzTargetInput| {
                     .compute_tc()
                     .expect("tc computation failed");
                 let expr = policy.condition();
-                for i in 0..8 {
-                    let request: ast::Request = input.requests[i].clone().into();
-                    let partial_request = &input.partial_requests[i];
+                for (request, partial_request) in
+                    input.requests.into_iter().zip(input.partial_requests)
+                {
+                    let request: ast::Request = request.into();
                     if passes_request_validation(&schema, &request) {
                         let residuals =
                             tpe_policies(&policyset, &partial_request, &partial_entities, &schema)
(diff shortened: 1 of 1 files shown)


cedar-drt/fuzz/fuzz_targets/tpe-pbt.rs
```diff
@@ -0,0 +1,272 @@
...
+        debug!("Schema: {schemafile_string}");
+        if let Ok(entities) = Entities::try_from(input.hierarchy.clone()) {
+            let validator = Validator::new(schema.clone());
+            let mut policyset = ast::PolicySet::new();
+            let policy: ast::StaticPolicy = input.policy.into();
+            policyset.add_static(policy.clone()).unwrap();
+            let passes_strict = passes_policy_validation(&validator, &policyset);
+            if passes_strict {
+                for e in input.partial_entities.entities() {
+                    e.validate(&schema).expect("entities should be valid");
```
john-h-kastner-aws, line 246: should be able to call `validator.schema()` to avoid cloning above

cedar-drt/fuzz/fuzz_targets/tpe-pbt.rs
```diff
@@ -0,0 +1,272 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#![no_main]
+use cedar_drt::logger::initialize_log;
+use cedar_drt_inner::*;
```
chaluli: Please, fully specify the items you use. The previous version of `cedar-drt` was a huge pain to read and understand because all the imports were glob imported and for some reason rust-analyzer and VSCode could not figure out where the elements were imported from.

cedar-drt/fuzz/fuzz_targets/tpe-pbt.rs
```diff
@@ -0,0 +1,272 @@
...
+            if passes_strict {
+                for e in input.partial_entities.entities() {
+                    e.validate(&schema).expect("entities should be valid");
+                }
+                let mut partial_entities = input.partial_entities;
+                partial_entities
+                    .compute_tc()
+                    .expect("tc computation failed");
+                let expr = policy.condition();
+                for i in 0..8 {
```
john-h-kastner-aws: `for (request, partial_request) in input.requests.into_iter().zip(input.partial_requests)`

john-h-kastner-aws (approved):

cedar-drt/fuzz/fuzz_targets/tpe-pbt.rs
```diff
@@ -0,0 +1,272 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#![no_main]
+use cedar_drt::logger::initialize_log;
+use cedar_drt_inner::*;
+use cedar_policy_core::ast;
+use cedar_policy_core::ast::RequestSchema;
+use cedar_policy_core::ast::{EntityUID, Expr, Request, Value};
```
chaluli, line 22: We provide a public interface for TPE in `cedar-policy`, please update this fuzzing target to use the public interface rather than `core`. This has already been done for other authorization and validation targets.

chaluli (changes requested):
Overall, looks good! Please fully specify imports so that we don't degrade the code's understandability and use the public interface for TPE.

cedar-drt/fuzz/fuzz_targets/tpe-pbt.rs
shaobo-he-aws: Got it. The current public APIs of TPE is very restrictive (e.g., they do not allow construction of `PartialEntities` from an iterator of `PartialEntity`. How about creating an issue to keep track it?

chaluli: An issue will work.

cedar-drt/fuzz/fuzz_targets/tpe-pbt.rs
shaobo-he-aws: Which clone did you refer to?

cedar-drt/fuzz/fuzz_targets/tpe-pbt.rs
shaobo-he-aws: Issue: #679

cedar-drt/fuzz/fuzz_targets/tpe-pbt.rs
john-h-kastner-aws: Replacing

```rust
let validator = Validator::new(schema.clone());
...
e.validate(&schema).expect("entities should be valid");
```

with

```rust
let validator = Validator::new(schema);
...
e.validate(validator.schema()).expect("entities should be valid");
```

PR #681, opened 2025-07-16
Batched evaluation algorithm (without proofs)

oflatt:
This PR adds a batched evaluation algorithm for cedar expressions.
This will enable efficient loading of entities from a backing database during a request. Compared to entity manifests, the algorithm is simple and it will load the same or fewer entities.

To make batched evaluation sound, it was necessary to extend partial evaluation so that the user can include `MissingEntity` objects in the partial entities store.

### DIFF
--- a/cedar-lean/Cedar/Data/Map.lean
+++ b/cedar-lean/Cedar/Data/Map.lean
@@ -15,7 +15,6 @@
 -/
 
 import Cedar.Data.Set
-import Cedar.Data.List
 /-!
 
 This file defines a simple map data types, backed by a sorted duplicate-free
@@ -53,7 +52,7 @@ def toList {α : Type u} {β : Type v} (m : Map α β) : List (Prod α β) := m.
 def insert {α β} [LT α] [DecidableLT α] (m : Map α β) (k : α) (v : β) : Map α β :=
   -- TODO more efficient implementation
   -- right now just rebuilds the map
-  Map.mk (((Prod.mk k v) :: m.kvs).canonicalize Prod.fst)
+  Map.make ((k, v) :: m.kvs)
 
 /-- Returns the keys of `m` as a set. -/
 def keys {α β} (m : Map α β) : Set α :=
--- a/cedar-lean/Cedar/Data/Map.lean
+++ b/cedar-lean/Cedar/Data/Map.lean
@@ -114,6 +114,9 @@ instance decLt [LT (Prod α β)] [DecidableEq (Prod α β)] [DecidableLT (Prod �
 instance : Membership α (Map α β) where
   mem m a := List.Mem a (m.kvs.map Prod.fst)
 
+instance [LT α] [DecidableLT α] : HAppend (Map α β) (Map α β) (Map α β) where
+  hAppend a b := Map.make (a.kvs ++ b.kvs)
+
 end Map
 
 end Cedar.Data
--- a/cedar-lean/Cedar/Data/Map.lean
+++ b/cedar-lean/Cedar/Data/Map.lean
@@ -15,6 +15,7 @@
 -/
 
 import Cedar.Data.Set
+import Cedar.Data.List
 /-!
 
 This file defines a simple map data types, backed by a sorted duplicate-free
@@ -47,6 +48,12 @@ def empty {α β} : Map α β := .mk []
 /-- Returns an ordered and duplicate free list provided the given map is well-formed. -/
 def toList {α : Type u} {β : Type v} (m : Map α β) : List (Prod α β) := m.kvs
 
+
+/-- inserts an element, overwriting any existing element -/
+def insert {α β} [LT α] [DecidableLT α] (m : Map α β) (k : α) (v : β) : Map α β :=
+  -- TODO more efficient implementation
+  -- right now just rebuilds the map
+  Map.mk (((Prod.mk k v) :: m.kvs).canonicalize Prod.fst)
 
 /-- Returns the keys of `m` as a set. -/
 def keys {α β} (m : Map α β) : Set α :=
--- a/cedar-lean/Cedar/Data/Map.lean
+++ b/cedar-lean/Cedar/Data/Map.lean
@@ -48,12 +48,6 @@ def empty {α β} : Map α β := .mk []
 def toList {α : Type u} {β : Type v} (m : Map α β) : List (Prod α β) := m.kvs
 
 
-/-- inserts an element, overwriting any existing element -/
-def insert {α β} [LT α] [DecidableLT α] (m : Map α β) (k : α) (v : β) : Map α β :=
-  -- TODO more efficient implementation
-  -- right now just rebuilds the map
-  Map.make ((k, v) :: m.kvs)
-
 /-- Returns the keys of `m` as a set. -/
 def keys {α β} (m : Map α β) : Set α :=
   Set.mk (m.kvs.map Prod.fst) -- well-formed by construction
--- a/cedar-lean/Cedar/TPE/BatchedEvaluator.lean
+++ b/cedar-lean/Cedar/TPE/BatchedEvaluator.lean
@@ -80,13 +80,12 @@ partial def batchedEvalLoop
   : Result Value :=
   let toLoad := (Residual.allLiteralUIDs residual).filter (λ uid => (store.find? uid).isNone)
   let newEntities := loader toLoad
-  let newStore := newEntities.kvs.foldl (λ acc ed => acc.insert ed.1 ed.2) store
+  let newStore := Map.make (newEntities.kvs ++ store.kvs)
   let newRes := Cedar.TPE.evaluate residual (Request.asPartialRequest req) newStore
 
   match newRes with
   | .val v _ty => .ok v
-  | _ =>
-    batchedEvalLoop newRes req loader newStore
+  | _ => batchedEvalLoop newRes req loader newStore
 
 
 /--
@@ -100,18 +99,14 @@ def batchedEvaluate
   (loader : EntityLoader)
   : Result Value :=
   let emptyStore : PartialEntities := Map.mk []
-  let residual := TypedExpr.toResidual x
-  -- an initial partial evaluation, removing all variables
-  let residual₂ := Cedar.TPE.evaluate residual (Request.asPartialRequest req) emptyStore
+  let residual := Cedar.TPE.evaluate (TypedExpr.toResidual x) (Request.asPartialRequest req) Map.empty
   -- start the batched evaluation loop
-  batchedEvalLoop residual₂ req loader emptyStore
+  batchedEvalLoop residual req loader emptyStore
 
-def entityLoaderFor : (e: Entities) -> EntityLoader :=
-  fun e =>
-   fun uids =>
-    Map.make (uids.toList.map (fun uid =>
-      match (e.find? uid) with
-      | .some data =>
-        (uid, EntityData.asPartial data)
-      | .none =>
-        (uid, PartialEntityData.MissingEntity)))
+def entityLoaderFor (es: Entities) (uids : Set EntityUID) :=
+  Map.make (uids.toList.map (λ uid =>
+        match (es.find? uid) with
+        | .some data =>
+          (uid, EntityData.asPartial data)
+        | .none =>
+          (uid, PartialEntityData.abset)))
--- a/cedar-lean/Cedar/TPE/BatchedEvaluator.lean
+++ b/cedar-lean/Cedar/TPE/BatchedEvaluator.lean
@@ -37,7 +37,7 @@ def Residual.allLiteralUIDs (x : Residual) : Set EntityUID :=
   | .error _e                          => Set.empty
   | .var _ _                           => Set.empty
   | .ite x₁ x₂ x₃ _      =>
-    x₁.allLiteralUIDs ∪ x₂.allLiteralUIDs ∪ x₃.allLiteralUIDs
+    Residual.allLiteralUIDs x₁ ∪ Residual.allLiteralUIDs x₂ ∪ Residual.allLiteralUIDs x₃
   | .and x₁ x₂ _         =>
     Residual.allLiteralUIDs x₁ ∪ Residual.allLiteralUIDs x₂
   | .or x₁ x₂ _          =>
@@ -82,7 +82,7 @@ partial def batchedEvalLoop
   let toLoad := (Residual.allLiteralUIDs residual).filter (λ uid => (store.find? uid).isNone)
   let newEntities := loader toLoad
   let newStore := Map.make (newEntities.kvs ++ store.kvs)
-  let newRes := Cedar.TPE.evaluate residual req.asPartialRequest newStore
+  let newRes := Cedar.TPE.evaluate residual (Request.asPartialRequest req) newStore
 
   match newRes with
   | .val v _ty => .ok v
--- a/cedar-lean/Cedar/TPE/BatchedEvaluator.lean
+++ b/cedar-lean/Cedar/TPE/BatchedEvaluator.lean
@@ -30,43 +30,6 @@ open Cedar.Validation
 
 abbrev EntityLoader := Set EntityUID → Map EntityUID PartialEntityData
 
-def Residual.allLiteralUIDs (x : Residual) : Set EntityUID :=
-  match x with
-  | .val (.prim (.entityUID uid)) _ty  => Set.singleton uid
-  | .val _ _                           => Set.empty
-  | .error _e                          => Set.empty
-  | .var _ _                           => Set.empty
-  | .ite x₁ x₂ x₃ _      =>
-    Residual.allLiteralUIDs x₁ ∪ Residual.allLiteralUIDs x₂ ∪ Residual.allLiteralUIDs x₃
-  | .and x₁ x₂ _         =>
-    Residual.allLiteralUIDs x₁ ∪ Residual.allLiteralUIDs x₂
-  | .or x₁ x₂ _          =>
-    Residual.allLiteralUIDs x₁ ∪ Residual.allLiteralUIDs x₂
-  | .unaryApp _ x _      =>
-    Residual.allLiteralUIDs x
-  | .binaryApp _ x₁ x₂ _ =>
-    Residual.allLiteralUIDs x₁ ∪ Residual.allLiteralUIDs x₂
-  | .getAttr x _ _       => Residual.allLiteralUIDs x
-  | .hasAttr x _ _       => Residual.allLiteralUIDs x
-  | .set x _             =>
-    x.mapUnion₁ (λ ⟨v, _⟩ => Residual.allLiteralUIDs v)
-  | .record x _          =>
-    x.mapUnion₂ (λ ⟨⟨_attr, v⟩, _⟩ => Residual.allLiteralUIDs v)
-  | .call _ x _          =>
-    x.mapUnion₁ (λ ⟨v, _⟩ => Residual.allLiteralUIDs v)
-termination_by sizeOf x
-decreasing_by
-  any_goals
-    simp
-    try simp at *
-    try omega
-  all_goals
-    rename_i h
-    let so := List.sizeOf_lt_of_mem h
-    simp at *
-    omega
-
-
 /--
 The batched evaluation loop
   1. Asks for any new entities referenced by the residual
@@ -79,7 +42,7 @@ partial def batchedEvalLoop
   (loader : EntityLoader)
   (store : PartialEntities)
   : Result Value :=
-  let toLoad := (Residual.allLiteralUIDs residual).filter (λ uid => (store.find? uid).isNone)
+  let toLoad := residual.allLiteralUIDs.filter (λ uid => (store.find? uid).isNone)
   let newEntities := loader toLoad
   let newStore := Map.make (newEntities.kvs ++ store.kvs)
   let newRes := Cedar.TPE.evaluate residual (Request.asPartialRequest req) newStore
--- a/cedar-lean/Cedar/TPE/BatchedEvaluator.lean
+++ b/cedar-lean/Cedar/TPE/BatchedEvaluator.lean
@@ -56,16 +56,12 @@ def Residual.allLiteralUIDs (x : Residual) : Set EntityUID :=
     x.mapUnion₁ (λ ⟨v, _⟩ => Residual.allLiteralUIDs v)
 termination_by sizeOf x
 decreasing_by
-  repeat case _ =>
-    simp [*]; try omega
-  . rename_i h
-    let so := List.sizeOf_lt_of_mem h
+  any_goals
     simp
-    omega
-  . rename_i h
-    simp at *
-    omega
-  . rename_i h
+    try simp at *
+    try omega
+  all_goals
+    rename_i h
     let so := List.sizeOf_lt_of_mem h
     simp at *
     omega
--- a/cedar-lean/Cedar/TPE/BatchedEvaluator.lean
+++ b/cedar-lean/Cedar/TPE/BatchedEvaluator.lean
@@ -37,7 +37,7 @@ def Residual.allLiteralUIDs (x : Residual) : Set EntityUID :=
   | .error _e                          => Set.empty
   | .var _ _                           => Set.empty
   | .ite x₁ x₂ x₃ _      =>
-    Residual.allLiteralUIDs x₁ ∪ Residual.allLiteralUIDs x₂ ∪ Residual.allLiteralUIDs x₃
+    x₁.allLiteralUIDs ∪ x₂.allLiteralUIDs ∪ x₃.allLiteralUIDs
   | .and x₁ x₂ _         =>
     Residual.allLiteralUIDs x₁ ∪ Residual.allLiteralUIDs x₂
   | .or x₁ x₂ _          =>
--- a/cedar-lean/Cedar/TPE/BatchedEvaluator.lean
+++ b/cedar-lean/Cedar/TPE/BatchedEvaluator.lean
@@ -66,6 +66,14 @@ def batchedEvaluate
   -- start the batched evaluation loop
   batchedEvalLoop residual req loader Map.empty
 
+/--
+Create an entity loader for an entity store Entities.
+This is used to model user-provided entity loaders which
+load entities from a backing database.
+
+Given Entities es, the entity loader provides the requested
+entities specified by a set of entity ids.
+-/
 def entityLoaderFor (es: Entities) (uids : Set EntityUID) :=
   Map.make (uids.toList.map (λ uid =>
         match (es.find? uid) with
--- a/cedar-lean/Cedar/TPE/BatchedEvaluator.lean
+++ b/cedar-lean/Cedar/TPE/BatchedEvaluator.lean
@@ -99,10 +99,9 @@ def batchedEvaluate
   (req : Request)
   (loader : EntityLoader)
   : Result Value :=
-  let emptyStore : PartialEntities := Map.mk []
   let residual := Cedar.TPE.evaluate (TypedExpr.toResidual x) (Request.asPartialRequest req) Map.empty
   -- start the batched evaluation loop
-  batchedEvalLoop residual req loader emptyStore
+  batchedEvalLoop residual req loader Map.empty
 
 def entityLoaderFor (es: Entities) (uids : Set EntityUID) :=
   Map.make (uids.toList.map (λ uid =>
--- a/cedar-lean/Cedar/TPE/BatchedEvaluator.lean
+++ b/cedar-lean/Cedar/TPE/BatchedEvaluator.lean
@@ -109,4 +109,4 @@ def entityLoaderFor (es: Entities) (uids : Set EntityUID) :=
         | .some data =>
           (uid, EntityData.asPartial data)
         | .none =>
-          (uid, PartialEntityData.abset)))
+          (uid, PartialEntityData.absent)))
--- a/cedar-lean/Cedar/TPE/BatchedEvaluator.lean
+++ b/cedar-lean/Cedar/TPE/BatchedEvaluator.lean
@@ -1,0 +1,123 @@
+/-
+ Copyright Cedar Contributors
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+      https://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-/
+
+
+import Cedar.Spec
+import Cedar.Data
+import Cedar.TPE.Residual
+import Cedar.TPE.Evaluator
+import Cedar.TPE.Input
+
+namespace Cedar.TPE
+
+open Cedar.Data
+open Cedar.Spec
+open Cedar.Validation
+
+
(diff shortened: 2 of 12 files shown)


cedar-lean/Cedar/TPE/BatchedEvaluator.lean
```diff
@@ -0,0 +1,156 @@
...
+    then batchedEvalLoop new_expr_res req loader new_store
+    else Cedar.Spec.evaluate expr.toExpr req new_store
+
+termination_by TypedExpr.size expr
+decreasing_by
+  simp [TypedExpr.size]
+  omega
+
+
+/--
```
emina, line 55: Thanks for the PR!  

The current approach is feasible but it will take a lot of effort to continue down this path.  It'll be really hard to define `Value.asTypedExpr` and prove it correct for extension types.  Effectively, we'd have to find a way to convert the internal representation of each extension value back into a string argument to the right constructor, and then prove that this transformation is correct (a roundtrip lemma).  This would be quite challenging for Datetimes and IpAddr.

Here is what I'd do instead.  It is a more direct path (I think) though direct != easy in this case :) 

### Step 1:  Change TPE to work on Residuals rather than TypedExprs

The first step is the change the TPE code as follows:
```
import Cedar.TPE.Input
import Cedar.TPE.Residual

namespace Cedar.TPE

open Cedar.Data
open Cedar.Spec
open Cedar.Validation

--- Same as before ---

def evaluate
  (x : Residual)
  (req : PartialRequest)
  (es : PartialEntities) : Residual :=
  match x with
  | .val v ty => .val v ty
  | .var v ty => varₚ req v ty
  | .error ty => .error ty 
  | .ite x₁ x₂ x₃ ty =>
    ite (evaluate x₁ req es) (evaluate x₂ req es) (evaluate x₃ req es) ty
  | .and x₁ x₂ ty =>
    and (evaluate x₁ req es) (evaluate x₂ req es) ty
  | .or x₁ x₂ ty =>
    or (evaluate x₁ req es) (evaluate x₂ req es) ty
  | .unaryApp op₁ x₁ ty =>
    apply₁ op₁ (evaluate x₁ req es) ty
  | .binaryApp op₂ x₁ x₂ ty =>
    apply₂ op₂ (evaluate x₁ req es) (evaluate x₂ req es) es ty
  | .hasAttr x₁ a ty =>
    hasAttr (evaluate x₁ req es) a es ty
  | .getAttr x₁ a ty =>
    getAttr (evaluate x₁ req es) a es ty
  | .set xs ty =>
    set (xs.map₁ (λ ⟨x₁, _⟩ => evaluate x₁ req es)) ty
  | .record axs ty =>
    record (axs.map₁ (λ ⟨(a, x₁), _⟩ => (a, (evaluate x₁ req es)))) ty
  | .call xfn xs ty =>
    call xfn (xs.map₁ (λ ⟨x₁, _⟩ => evaluate x₁ req es)) ty
termination_by x
decreasing_by
  all_goals
    simp_wf
    try omega
  any_goals
    rename_i h
    replace h := List.sizeOf_lt_of_mem h
    try simp at h
    omega

end Cedar.TPE

namespace Cedar.Validation 

open Cedar.TPE 

def TypedExpr.toResidual : TypedExpr → Residual
  | .lit p ty => .val (.prim p) ty 
  | .var v ty => .var v ty
  | .ite x₁ x₂ x₃ ty => .ite x₁.toResidual x₂.toResidual x₃.toResidual ty 
  | _ => sorry -- the rest is similar

end Cedar.Validation 


namespace Cedar.TPE

open Cedar.Spec Cedar.Validation

def evaluatePolicy (schema : Schema)
  (p : Policy)
  (req : PartialRequest)
  (es : PartialEntities)
  : Except Error Residual :=
  match schema.environment? req.principal.ty req.resource.ty req.action with
    | .some env =>
      if requestAndEntitiesIsValid env req es
      then
        do
          let expr := substituteAction env.reqty.action p.toExpr
          let (te, _) ← (typeOf expr ∅ env).mapError Error.invalidPolicy
          .ok (evaluate te.liftBoolTypes.toResidual req es)
      else .error .invalidRequestOrEntities
    | .none => .error .invalidEnvironment

end Cedar.TPE
```

### Step 2: Fix the soundness proof for TPE

The new theorem will look like this:

```
theorem partial_evaluate_is_sound
  {x : TypedExpr}
  {req : Request}
  {es : Entities}
  {preq : PartialRequest}
  {pes : PartialEntities}
  {env : TypeEnv} :
  TypedExpr.WellTyped env x →
  InstanceOfWellFormedEnvironment req es env →
  RequestAndEntitiesRefine req es preq pes →
  (Spec.evaluate x.toExpr req es).toOption = (Residual.evaluate (Cedar.TPE.evaluate x.toResidual preq pes) req es).toOption
:= by sorry 
```

For this to proof, you will need to define a `WellTypedResidual` predicate, which will be very similar to `WellTypedExpr` predicate, and then you'll need to prove that `TypedExpr.toResidual` results in a well-typed output when given a well-typed input.

### Step 3: Change the implementation of `batchedEvaluate`

The new `batchedEvaluate` will look something like this:

```
def Residual.allLiteralUIDs (x : Residual) : Set EntityUID := sorry

def Residual.batchedEvaluate
  (x : Residual)
  (req : Request)
  (loader : EntityLoader)
  (store : Entities)
  : Result Value :=
  let toLoad := (TypedExpr.allLiteralUIDs expr).filter (λ uid => (store.find? uid).isNone)
  let newEntities := loader toLoad
  let newStore := newEntities.kvs.foldl (λ acc ed => acc.insert ed.1 ed.2) store

  do
    let newRes := Cedar.TPE.evaluate x (Request.asPartialRequest req) (Entities.asPartial newStore)
    if x = newRes  -- nothing has changed, we are done
    then newRes.evaluate ... -- just evaluate the residual as though the new store / request are complete
    else newRes.batchedEvaluate req loader newStore



def batchedEvaluate
  (x : TypedExpr)
  (req : Request)
  (loader : EntityLoader)
  : Result Value := do 
  let empty_store : Entities := Map.mk []
  -- an initial partial evaluation, removing all variables
  let residual := Cedar.TPE.evaluate x.toResidual (Request.asPartialRequest req) (Entities.asPartial empty_store)
  -- start the batched evaluation loop
  residual.batchedEvaluate req loader empty_store


```

### Step 4:  Prove soundness of `batchedEvaluate`

For this proof, you will need to show two lemmas:
1.  Partial evaluation produces well-typed residuals (given well-typed / well-formed inputs)
2. From that, you'll need to show that partial re-evaluation can be done incrementally.  So, if we have a partial input I1 and I2 such that I1 refines I2 (where you define refinement of partial inputs, similar to the existing definition), then partially evaluating against I1 and then against I2 is the same as just partially evaluating against I2.

oflatt: Good plan! Working on it now

shaobo-he-aws: Could you please break it down to multiple PRs so that we can DRT the changes while you work on other parts of the plan?

emina: Actually, we can simplify the termination condition to just `newEntites = Map.empty` (nothing new loaded), in which case we're done.

oflatt: Yep, I'm working in another branch

oflatt: I like the new termination condition, though it makes proving termination hard. I marked it as partial for now

cedar-lean/Cedar/Data/Map.lean
```diff
@@ -15,6 +15,7 @@
 -/
 
 import Cedar.Data.Set
+import Cedar.Data.List
```
emina: We shouldn't need to import the List module here (since it's already imported by Set, which is imported by Map).

cedar-lean/Cedar/Data/Map.lean
```diff
@@ -48,6 +49,12 @@ def empty {α β} : Map α β := .mk []
 def toList {α : Type u} {β : Type v} (m : Map α β) : List (Prod α β) := m.kvs
 
 
+/-- inserts an element, overwriting any existing element -/
+def insert {α β} [LT α] [DecidableLT α] (m : Map α β) (k : α) (v : β) : Map α β :=
+  -- TODO more efficient implementation
+  -- right now just rebuilds the map
+  Map.mk (((Prod.mk k v) :: m.kvs).canonicalize Prod.fst)
```
emina:
```suggestion
  Map.make ((k, v) :: m.kvs)
```

cedar-lean/Cedar/TPE/BatchedEvaluator.lean
```diff
@@ -0,0 +1,123 @@
...
+import Cedar.TPE.Input
+
+namespace Cedar.TPE
+
+open Cedar.Data
+open Cedar.Spec
+open Cedar.Validation
+
+
+def EntityLoader : Type := Set EntityUID → Map EntityUID PartialEntityData
```
emina: Why not use an `abbrev` here?

`abbrev EntityLoader := Set EntityUID → Map EntityUID PartialEntityData`

`abbrev`s are generally easier to work with since Lean can "see" their definition (and it becomes opaque when we use `def`).

cedar-lean/Cedar/TPE/BatchedEvaluator.lean
```diff
@@ -0,0 +1,123 @@
...
+def Residual.allLiteralUIDs (x : Residual) : Set EntityUID :=
+  match x with
+  | .val v _ty =>
+    match v with
+    | .prim (.entityUID uid) => Set.singleton uid
+    | _ => Set.empty
+  | .error _e =>
+    Set.empty
+  | .var v _ =>
+    -- these cases should not happen, since the request was fully concrete
```
emina: I don't understand this comment.

cedar-lean/Cedar/TPE/BatchedEvaluator.lean
```diff
@@ -0,0 +1,123 @@
...
+
+def Residual.allLiteralUIDs (x : Residual) : Set EntityUID :=
+  match x with
+  | .val v _ty =>
+    match v with
+    | .prim (.entityUID uid) => Set.singleton uid
+    | _ => Set.empty
+  | .error _e =>
+    Set.empty
+  | .var v _ =>
```
emina:
```suggestion
  | .var v _ => Set.empty
```

cedar-lean/Cedar/TPE/BatchedEvaluator.lean
```diff
@@ -0,0 +1,123 @@
...
+    -- these cases should not happen, since the request was fully concrete
+    match v with
+    | .principal => Set.empty
+    | .resource  => Set.empty
+    | .action    => Set.empty
+    | .context   => Set.empty
+  | .ite c t e _ => Residual.allLiteralUIDs c ∪ Residual.allLiteralUIDs t ∪ Residual.allLiteralUIDs e
```
emina: Naming convention:  we use `x` for "expressions".  So for `.ite`, the match pattern would be `| .ite x₁ x₂ x₃`.  See, e.g., https://github.com/cedar-policy/cedar-spec/blob/255b05588fd67a05e1b0c25c1279af35fec7bb05/cedar-lean/Cedar/Spec/Evaluator.lean#L93

cedar-lean/Cedar/TPE/BatchedEvaluator.lean
```diff
@@ -0,0 +1,123 @@
...
+-/
+partial def batchedEvalLoop
+  (residual : Residual)
+  (req : Request)
+  (loader : EntityLoader)
+  (store : PartialEntities)
+  : Result Value :=
+  let toLoad := (Residual.allLiteralUIDs residual).filter (λ uid => (store.find? uid).isNone)
+  let newEntities := loader toLoad
+  let newStore := newEntities.kvs.foldl (λ acc ed => acc.insert ed.1 ed.2) store
```
emina: Why do a fold here as opposed to just unioning `store` and `newEntities`?

```
let newStore := Map.make (newEntities.kvs ++ store.kvs)
```

This would be more efficient, and won't need to add the `insert` function in `Cedar.Data.Map` :)

cedar-lean/Cedar/TPE/BatchedEvaluator.lean
```diff
@@ -0,0 +1,123 @@
...
+  let newRes := Cedar.TPE.evaluate residual (Request.asPartialRequest req) newStore
+
+  match newRes with
+  | .val v _ty => .ok v
+  | _ =>
+    batchedEvalLoop newRes req loader newStore
```
emina: Nit: let's remove the line break here and put this expression on the same line as the pattern (it's short).

cedar-lean/Cedar/TPE/BatchedEvaluator.lean
```diff
@@ -0,0 +1,123 @@
...
+-/
+def batchedEvaluate
+  (x : TypedExpr)
+  (req : Request)
+  (loader : EntityLoader)
+  : Result Value :=
+  let emptyStore : PartialEntities := Map.mk []
+  let residual := TypedExpr.toResidual x
+  -- an initial partial evaluation, removing all variables
+  let residual₂ := Cedar.TPE.evaluate residual (Request.asPartialRequest req) emptyStore
```
emina: I'd simplify this a bit to remove the extra temporary variables:

```
let residual := Cedar.TPE.evaluate (TypedExpr.toResidual x) (Request.asPartialRequest req) Map.empty
batchedEvalLoop residual req loader Map.empty
```

cedar-lean/Cedar/TPE/BatchedEvaluator.lean
```diff
@@ -0,0 +1,123 @@
...
+  -- start the batched evaluation loop
+  batchedEvalLoop residual₂ req loader emptyStore
+
+def entityLoaderFor : (e: Entities) -> EntityLoader :=
```
emina: Nit:  we use `es` for "Entities" values.

emina: We can write this more simply like this (Lean supports currying for callers):

```
def entityLoaderFor (es: Entities) (uids : Set EntityUID) :  EntityLoader :=
 Map.make (uids.toList.map (λ uid =>
      match (e.find? uid) with
      | .some data =>
        (uid, EntityData.asPartial data)
      | .none =>
        (uid, PartialEntityData.MissingEntity)))
```

cedar-lean/Cedar/TPE/Input.lean
```diff
@@ -43,17 +43,41 @@ structure PartialRequest where
...
+
+
 -- We don't need type annotations here following the rationale above
-structure PartialEntityData where
-  attrs     : Option (Map Attr Value)
-  ancestors : Option (Set EntityUID)
-  tags      : Option (Map Attr Value)
+inductive PartialEntityData where
+  | present (attrs : Option (Map Attr Value)) (ancestors : Option (Set EntityUID)) (tags : Option (Map Attr Value))
+  | MissingEntity
```
emina: Nit: we use lowerCaseCamel for constructor names.  I would call this just `absent` (for symmetry with `present` above).

cedar-lean/Cedar/TPE/Input.lean
```diff
@@ -43,17 +43,41 @@ structure PartialRequest where
...
 
+def Entities.asPartial (entities: Entities) : PartialEntities :=
+  entities.mapOnValues EntityData.asPartial
+
 def PartialEntities.get (es : PartialEntities) (uid : EntityUID) (f : PartialEntityData → Option α) : Option α :=
   (es.find? uid).bind f
 
+def PartialEntityData.ancestors : PartialEntityData → Option (Set EntityUID)
+  | .present _ ancestors _ => ancestors
+  | .MissingEntity => .some Set.empty
```
emina: I don't follow this:  why is it `.some Set.empty` versus `.none`? 

Same question for the `PartialEntities` functions below.

cedar-lean/Cedar/TPE/BatchedEvaluator.lean
oflatt: Oops stray comment

cedar-lean/Cedar/TPE/BatchedEvaluator.lean
oflatt: Ah I've been writing too much rust clearly
Thanks!

cedar-lean/Cedar/TPE/Input.lean
oflatt: This is pretty subtle, and it's possible I've made an error (hoping this comes up in the proof).
Looking at the call sites, whenever we access `PartialEntityData.ancestors` and the entity isn't there in the first place, the code can safely return no ancestors. This is because normal evaluation returns false on ancestor checks for missing entities. 

Does that sound right to you? Perhaps I missed a case?

oflatt: Similarly, a missing entity can have no tags and no attributes (an error will be thrown because the tag won't be found in the case of get tag)

cedar-lean/Cedar/TPE/BatchedEvaluator.lean
oflatt: nice idea, I didn't know that!

cedar-lean/Cedar/TPE/Input.lean
```diff
@@ -53,7 +53,7 @@ def Request.asPartialRequest (req : Request) : PartialRequest :=
 -- We don't need type annotations here following the rationale above
 inductive PartialEntityData where
   | present (attrs : Option (Map Attr Value)) (ancestors : Option (Set EntityUID)) (tags : Option (Map Attr Value))
-  | MissingEntity
+  | abset
```
emina: Typo: "absent" :)

cedar-lean/Cedar/TPE/Input.lean
```diff
@@ -68,15 +68,15 @@ def PartialEntities.get (es : PartialEntities) (uid : EntityUID) (f : PartialEnt
...
 
 def PartialEntityData.tags : PartialEntityData → Option (Map Attr Value)
   | .present _ _ tags => tags
-  | .MissingEntity => .some Map.empty
+  | .abset => .some Map.empty
 
 def PartialEntityData.attrs : PartialEntityData → Option (Map Attr Value)
   | .present attrs _ _ => attrs
-  | .MissingEntity => .some Map.empty
+  | .abset => .some Map.empty
```
emina: Returning `.some Map.empty` for the `.absent` case is not always type-correct.  I think proofs will eventually fail here, because they will expected `PartialEntityData.attrs` to conform to the type of the attribute record specified in the schema.  If such a type contains any required fields, we can't return the empty record here.

cedar-lean/Cedar/TPE/Input.lean
```diff
@@ -68,15 +68,15 @@ def PartialEntities.get (es : PartialEntities) (uid : EntityUID) (f : PartialEnt
 
 def PartialEntityData.ancestors : PartialEntityData → Option (Set EntityUID)
```
emina, line 57: Why is the type here `Option T` rather than just `T`, given that neither of the cases ever return `none`?

cedar-lean/Cedar/TPE/Input.lean
```diff
@@ -68,15 +68,15 @@ def PartialEntities.get (es : PartialEntities) (uid : EntityUID) (f : PartialEnt
...
   | .present _ ancestors _ => ancestors
-  | .MissingEntity => .some Set.empty
+  | .abset => .some Set.empty
 
 def PartialEntityData.tags : PartialEntityData → Option (Map Attr Value)
```
emina: Same question as above:  why is the type `Option (Map Attr Value)` rather than `(Map Tag Value)`, given that `none` is never returned?

(Also note that we use `Tag` instead of `Attr` as the abbreviation for String when we're referring to tag keys.)

cedar-lean/Cedar/TPE/Input.lean
oflatt: Actually the first case sometimes returns none, since partial entities may not come with the ancestors

cedar-lean/Cedar/TPE/Input.lean
oflatt: Similarly, the tags might not be provided on present entities so the first case can return none

cedar-lean/Cedar/TPE/Input.lean
oflatt: That's a great point! I think this doesn't actually end up causing problems in the proof because the returned map is always used immediately in a computation that fails.
This is called by `getAttr` and `hasAttr`, both of which return a boolean. In other words, this map doesn't end up a part of the resulting residual. 

I was able to get the proof that PE produces a well typed residual to go through

cedar-lean/Cedar/TPE/Input.lean
oflatt: thanks

cedar-lean/Cedar/TPE/BatchedEvaluator.lean
```diff
@@ -0,0 +1,112 @@
...
+  | .getAttr x _ _ => Residual.allLiteralUIDs x
+  | .hasAttr x _ _ => Residual.allLiteralUIDs x
+  | .set x _ => x.mapUnion₁ (λ ⟨v, _⟩ => Residual.allLiteralUIDs v)
+  | .record x _ => x.mapUnion₂ (λ ⟨⟨_attr, v⟩, _⟩ => Residual.allLiteralUIDs v)
+  | .call _ x _ => x.mapUnion₁ (λ ⟨v, _⟩ => Residual.allLiteralUIDs v)
+termination_by sizeOf x
+decreasing_by
+  repeat case _ =>
+    simp [*]; try omega
+  . rename_i h
```
emina: I haven't tried this in the Lean IDE :) but it seems like we should be able to replace these try `. rename_i h` cases with something along the following lines:

```
all_goals {
  rename_i h
  simp at *
  omega 
}
```

cedar-lean/Cedar/TPE/BatchedEvaluator.lean
```diff
@@ -0,0 +1,112 @@
...
+open Cedar.Data
+open Cedar.Spec
+open Cedar.Validation
+
+
+abbrev EntityLoader := Set EntityUID → Map EntityUID PartialEntityData
+
+def Residual.allLiteralUIDs (x : Residual) : Set EntityUID :=
+  match x with
+  | .val v _ty =>
```
emina: A personal formatting preference:  I like to vertically align `=>` for each case.  It's easier to quickly scan the per-case code (for me at least).

cedar-lean/Cedar/TPE/BatchedEvaluator.lean
```diff
@@ -0,0 +1,112 @@
...
+Evaluate a cedar expression using an EntityLoader
+instead of a full Entities store.
+This algorithm minimizes the number of calls to the EntityLoader using partial evaluation.
+-/
+def batchedEvaluate
+  (x : TypedExpr)
+  (req : Request)
+  (loader : EntityLoader)
+  : Result Value :=
+  let emptyStore : PartialEntities := Map.mk []
```
emina: I think you should be able to delete this line, and change 104 to `batchedEvalLoop residual req loader Map.empty`.

cedar-lean/Cedar/TPE/BatchedEvaluator.lean
oflatt: Nice idea, I was able to get this to work

cedar-lean/Cedar/TPE/BatchedEvaluator.lean
oflatt: nice, done

cedar-lean/Cedar/TPE/BatchedEvaluator.lean
oflatt: yep!

cedar-lean/Cedar/TPE/BatchedEvaluator.lean
```diff
@@ -0,0 +1,112 @@
...
+abbrev EntityLoader := Set EntityUID → Map EntityUID PartialEntityData
+
+def Residual.allLiteralUIDs (x : Residual) : Set EntityUID :=
+  match x with
+  | .val (.prim (.entityUID uid)) _ty  => Set.singleton uid
+  | .val _ _                           => Set.empty
+  | .error _e                          => Set.empty
+  | .var _ _                           => Set.empty
+  | .ite x₁ x₂ x₃ _      =>
+    Residual.allLiteralUIDs x₁ ∪ Residual.allLiteralUIDs x₂ ∪ Residual.allLiteralUIDs x₃
```
cdisselkoen:
```suggestion
    x₁.allLiteralUIDs ∪ x₂.allLiteralUIDs ∪ x₃.allLiteralUIDs
```

cedar-lean/Cedar/TPE/BatchedEvaluator.lean
```diff
@@ -0,0 +1,112 @@
...
+-/
+partial def batchedEvalLoop
+  (residual : Residual)
+  (req : Request)
+  (loader : EntityLoader)
+  (store : PartialEntities)
+  : Result Value :=
+  let toLoad := (Residual.allLiteralUIDs residual).filter (λ uid => (store.find? uid).isNone)
+  let newEntities := loader toLoad
+  let newStore := Map.make (newEntities.kvs ++ store.kvs)
```
cdisselkoen: Does our `Map` type support the union operator?

cedar-lean/Cedar/TPE/BatchedEvaluator.lean
```diff
@@ -0,0 +1,112 @@
...
+  let newRes := Cedar.TPE.evaluate residual (Request.asPartialRequest req) newStore
```
cdisselkoen:
```suggestion
  let newRes := Cedar.TPE.evaluate residual req.asPartialRequest newStore
```

cedar-lean/Cedar/TPE/BatchedEvaluator.lean
```diff
@@ -0,0 +1,112 @@
...
+Evaluate a cedar expression using an EntityLoader
+instead of a full Entities store.
+This algorithm minimizes the number of calls to the EntityLoader using partial evaluation.
+-/
+def batchedEvaluate
+  (x : TypedExpr)
+  (req : Request)
+  (loader : EntityLoader)
+  : Result Value :=
+  let residual := Cedar.TPE.evaluate (TypedExpr.toResidual x) (Request.asPartialRequest req) Map.empty
```
cdisselkoen:
```suggestion
  let residual := Cedar.TPE.evaluate x.toResidual req.asPartialRequest Map.empty
```

cedar-lean/Cedar/TPE/BatchedEvaluator.lean
```diff
@@ -0,0 +1,112 @@
...
+  -- start the batched evaluation loop
+  batchedEvalLoop residual req loader Map.empty
+
+def entityLoaderFor (es: Entities) (uids : Set EntityUID) :=
```
cdisselkoen, line 77: Can we get a comment explaining this function?  I think `entityLoaderFor es` is an `EntityLoader` that just "loads" data from the given `es`?

cedar-lean/Cedar/TPE/Input.lean
```diff
@@ -138,7 +163,7 @@ where
       es₂.kvs.all λ (a₂, e₂) => match es₁.find? a₂ with
         | .some e₁ =>
           let ⟨attrs₁, ancestors₁, tags₁⟩ := e₁
-          let ⟨attrs₂, ancestors₂, tags₂⟩ := e₂
+          let (attrs₂, ancestors₂, tags₂) := (e₂.attrs, e₂.ancestors, e₂.tags)
```
cdisselkoen: is this `let` helping readability?  The newly bound names are essentially the same length/complexity as the old ones

cedar-lean/Cedar/Thm/Data/Map.lean
```diff
@@ -752,6 +752,19 @@ theorem in_kvs_in_mapOnValues [LT α] [DecidableLT α] [DecidableEq α] {f : β
   simp only [kvs, List.mem_map, Prod.mk.injEq]
   exists (k, v)
 
+
+/--
+  Given a hypothesis like:
+    h: Data.Map.find? es uid = none
+  Proves something like:
+    Data.Map.findOrErr es uid Error.entityDoesNotExist
```
cdisselkoen:
```suggestion
    Data.Map.findOrErr es uid Error.entityDoesNotExist = .error .entityDoesNotExist
```

cedar-lean/Cedar/Thm/Data/Map.lean
```diff
@@ -752,6 +752,19 @@ theorem in_kvs_in_mapOnValues [LT α] [DecidableLT α] [DecidableEq α] {f : β
...
+-/
+theorem find?_none_implies_findorErr_errors [LT α] [DecidableLT α] [DecidableEq α] {m : Map α β} {k : α} (e : Error) :
+  m.find? k = none → m.findOrErr k e = .error e
```
cdisselkoen: Should we go ahead and prove this as an iff?

cdisselkoen (approved):
Looks great! Mostly tiny syntax nits

cedar-lean/Cedar/TPE/BatchedEvaluator.lean
oflatt: I can't use dot notation here because `Request.asPartialRequest` isn't defined in the same namespace as Request
I can't move it over to the same namespace because it causes build cycle issues with partial requests

cedar-lean/Cedar/TPE/BatchedEvaluator.lean
oflatt: I don't think so- I think it would be werid because `Map.union a b != Map.union b a`

cedar-lean/Cedar/TPE/BatchedEvaluator.lean
oflatt: same here

cedar-lean/Cedar/TPE/BatchedEvaluator.lean
oflatt: Good idea
Yep!

cedar-lean/Cedar/TPE/Input.lean
oflatt: good point

cedar-lean/Cedar/TPE/BatchedEvaluator.lean
cdisselkoen: In that case should `Map` support the `++` operator directly?

cedar-lean/Cedar/TPE/BatchedEvaluator.lean
cdisselkoen: Can you get around that by putting it inside `namespace ... end` in its current file, without moving it to another file?  Would that fix the build cycle issues?

cedar-lean/Cedar/Thm/Data/Map.lean
oflatt: sure

emina (approved):

cedar-lean/Cedar/TPE/BatchedEvaluator.lean
oflatt: sure!

cedar-lean/Cedar/TPE/BatchedEvaluator.lean
oflatt: Oh that works, thanks for the tip

emina (approved):

PR #687, opened 2025-07-21
Update CI

shaobo-he-aws:
*Issue #, if available:*

*Description of changes:*

#685 added the following to `Cargo.toml` files of `cedar-lean-ffi` and `cedar-lean-cli` so that it is not needed in CI.

```
[patch.crates-io]
cedar-policy = { path = "../cedar/cedar-policy" }
```

In other words, both packages essentially depends on the version of `cedar-policy` on the main branch, as opposed to the one on crates.io.


chaluli (approved):
LGTM.

PR #683, opened 2025-07-21
`SymEnv.ofEnv` is strongly well-formed

zhengyao-lin:
This PR extends the results in #661 to `StronglyWellFormed` and `StronglyWellFormedFor*` (in particular with proofs for acyclicity, transitivity, and `SymEntities.Partitioned` of `SymEnv.ofEnv Γ`):
```
theorem ofEnv_is_swf
  {Γ : TypeEnv}
  (hwf : Γ.WellFormed) :
  (SymEnv.ofEnv Γ).StronglyWellFormed

theorem ofEnv_swf_for_expr
  {Γ : TypeEnv} {tx : TypedExpr}
  (hwf : Γ.WellFormed)
  (hwt : TypedExpr.WellTyped Γ tx) :
  (SymEnv.ofEnv Γ).StronglyWellFormedFor tx.toExpr

theorem ofEnv_swf_for_policy
  {Γ : TypeEnv} {p p' : Policy}
  (hwf : Γ.WellFormed)
  (hwt : wellTypedPolicy p Γ = .some p') :
  (SymEnv.ofEnv Γ).StronglyWellFormedForPolicy p'

theorem ofEnv_swf_for_policies
  {Γ : TypeEnv} {ps ps' : Policies}
  (hwf : Γ.WellFormed)
  (hwt : wellTypedPolicies ps Γ = .some ps') :
  (SymEnv.ofEnv Γ).StronglyWellFormedForPolicies ps'
```

Thanks for reviewing!

PR #682, opened 2025-07-17
`SymEnv.ofEnv` is sound

zhengyao-lin:
This PR proves the following soundness theorem for `SymEnv.ofEnv`
```
theorem ofEnv_soundness
  {Γ : TypeEnv} {env : Env}
  env.StronglyWellFormed →
  InstanceOfWellFormedEnvironment env.request env.entities Γ →
  env ∈ᵢ SymEnv.ofEnv Γ
```
i.e., for any sufficiently well-formed `env : Env`, we can construct a well-formed `Interpretation` of `SymEnv.ofEnv Γ` that corresponds to `env`.

This can be used to reduce the preconditions of some top-level soundness theorems in [`Cedar/Thm/Verification`](https://github.com/cedar-policy/cedar-spec/blob/ef53d7f049f8f1c965ba5b04d06fc136470d5ad9/cedar-lean/Cedar/Thm/Verification.lean) from `env ∈ᵢ εnv` to `InstanceOfWellFormedEnvironment env.request env.entities Γ`, which is a simpler condition (and has an executable version).

The `Interpretation` is constructed using the new `Env.symbolize?` function in `Cedar/SymCC/Symbolizer.lean`.
Since it's executable, it potentially enables a nice feature to specify a concrete entity store in verification requests in SymCC: one can specialize an `εnv : SymEnv` by `εnv.interpret env.symbolize?`. I'll leave this to a future PR.

This PR is a bit large; but the majority of new definitions are in `Cedar/SymCC/Symbolizer.lean`, and the rest is mostly only proofs.

Thank you!!

PR #661, opened 2025-07-07
Prove that well-formed `Environment` compiles to well-formed `SymEnv`

zhengyao-lin:
This PR proves these theorems (`Cedar/Thm/SymCC/Env/ofEnv.lean`):
```
theorem ofEnv_is_wf
  {Γ : Environment}
  (hwf : Γ.WellFormed) :
  (SymEnv.ofEnv Γ).WellFormed

theorem ofEnv_wf_for_expr
  {Γ : Environment} {tx : TypedExpr}
  (hwf : Γ.WellFormed)
  (hwt : TypedExpr.WellTyped Γ tx) :
  (SymEnv.ofEnv Γ).WellFormedFor tx.toExpr
```

and it also improves the theorem `compile_well_typed` from a previous PR #646 to the following statement (replacing `(SymEnv.ofEnv Γ).WellFormedFor (tx.toExpr)` by `Γ.WellFormed`)
```
theorem compile_well_typed {tx : TypedExpr} {Γ : Environment} :
  Γ.WellFormed →
  TypedExpr.WellTyped Γ tx →
  ∃ t : Term,
    compile tx.toExpr (SymEnv.ofEnv Γ) = .ok t ∧
    t.WellFormed (SymEnv.ofEnv Γ).entities ∧
    t.typeOf = .option (TermType.ofType tx.typeOf)
```

This PR is dependent on two previous unmerged PRs #658 and #659, so it's not ready for merging yet; but it is ready for review.

Thank you!

PR #688, opened 2025-07-22
SymCC: now 10% more sound

zhengyao-lin:
With `compile` verified to be well-typed (#646) and various soundness/well-formedness proofs of `SymEnv.ofEnv` (#661, #682, #683), we can now improve the soundness theorems of verifiers to only talk about concrete `TypeEnv` and `Env` and validation results.

As an example, previously we have
```
theorem verifyNeverErrors_is_sound  {p : Policy} {εnv : SymEnv} {asserts : Asserts} :
  εnv.StronglyWellFormedForPolicy p →
  verifyNeverErrors p εnv = .ok asserts →
  εnv ⊭ asserts →
  ∀ env,
    env ∈ᵢ εnv →
    env.StronglyWellFormedForPolicy p →
    (evaluate p.toExpr env.request env.entities).isOk
```

This PR adds improved counterparts like the following
```
theorem verifyNeverErrors_is_ok_and_sound {p p' : Policy} {Γ : TypeEnv} :
  Γ.WellFormed →
  wellTypedPolicy p Γ = .some p' →
  ∃ asserts,
    verifyNeverErrors p' (SymEnv.ofEnv Γ) = .ok asserts ∧
    SymEnv.ofEnv Γ ⊭ asserts →
      ∀ env : Env,
        InstanceOfWellFormedEnvironment env.request env.entities Γ →
        env.StronglyWellFormedForPolicy p →
        (evaluate p.toExpr env.request env.entities).isOk
```
In particular, the improvements are:
- We now only depend on `Γ.WellFormed` and policy validation in `wellTypedPolicy`, instead of assuming complex assumptions about the symbolic environment (`εnv.StronglyWellFormedForPolicy`)
- `verifyNeverErrors` (and other verification tasks) are guaranteed to succeed (due to [`compile_well_typed`](https://github.com/cedar-policy/cedar-spec/pull/646))
- In the postcondition, we can talk about all `Env` that is `InstanceOfWellFormedEnvironment`, instead of the more complex model-based condition `env ∈ᵢ εnv`.

All of these high-level theorems are in the new `Cedar/Thm/SymCC/WellTypedVerification.lean` file.

This PR is a bit long; thanks a lot for reviewing!


chaluli:
How do you get 10% specifically? Also, I don't see changes to the code portion of SymCC. As a pedant, I'd like to point out since you didn't change the code, it's just as sound as it was before. But now we have the proofs! :)

cedar-lean/Cedar/Thm/SymCC/Verifier/Basic.lean
```diff
@@ -0,0 +1,229 @@
+/-
```
emina: We reserve the name `Basic` for files that contain basic definitions (and usually no theorems). This file provides lemmas showing that well-typed policies lead to `.ok` results for the symbolic authorizer and compiler and their auxiliary functions.

Let's rename it to `WellTypedOk` or something like that to reflect its contents.

emina (changes requested):
Ah, let's rename `Basic` to `WellTypedOk` or something like that (see previous comment about using `Basic` for definitions).

cedar-lean/Cedar/Thm/SymCC/Verifier/Basic.lean
zhengyao-lin: makes sense. done!

chaluli (approved):

PR #689, opened 2025-07-22
Require that well-formed `TypeEnv` does not have singleton Bool types

zhengyao-lin:
An issue has come up in my attempt at completeness of `SymEnv.ofEnv`, where our current definition of `TypeEnv.WellFormed` does not enforce that `TypeEnv` contains no singleton Bool types like `.bool .tt` or `.bool .ff`.

This PR strengthens `TypeEnv.WellFormed` to require that `TypeEnv` does not have these singleton Bool types, and fixes the executable version of it (`TypeEnv.validateWellFormed`).

@shaobo-he-aws @chaluli please check if this makes sense as `TypeEnv.validateWellFormed` is called in both `cedar-lean-cli` and TPE (`isValidAndConsistent`) to sanity-check inputs.

### DIFF
--- a/cedar-lean/Cedar/Thm/Validation/EnvironmentValidation.lean
+++ b/cedar-lean/Cedar/Thm/Validation/EnvironmentValidation.lean
@@ -168,6 +168,52 @@ decreasing_by
 
 end
 
+theorem type_validate_lifted_is_sound
+  {ty : CedarType}
+  (hok : ty.validateLifted = .ok ()) :
+  CedarType.IsLifted ty
+:= by
+  cases ty with
+  | bool bty =>
+    cases bty
+    · constructor
+    · simp [CedarType.validateLifted] at hok
+    · simp [CedarType.validateLifted] at hok
+  | int | string | entity _ => constructor
+  | set ty' =>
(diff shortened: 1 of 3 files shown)


shaobo-he-aws:
I need to look into it further. `isValidAndConsistent` shouldn't check singleton types because TPE doesn't leverage type information explicitly during evaluation.

emina:
> I need to look into it further. `isValidAndConsistent` shouldn't check singleton types because TPE doesn't leverage type information explicitly during evaluation.

Regardless of TPE, we should strengthen the definition of a WF environment to disallow singleton types.  This is both required by the symbolic evaluator and enforced by the schema syntax.  That is, it's syntactically impossible for a Cedar schema to give rise to a type environment that includes tt/ff.

zhengyao-lin:
sorry scratch what I said; I think this might actually not be easily doable.

TPE depends on `well_typed_is_sound`, which needs to assume `env.WellFormed` since we have strenghtened the definition of `InstanceOfType`.

emina:
I think we should go with the new strengthened definition.  It's simply formalizing what's already enforced by schema syntax (that there is no such thing as types tt/ff in the schema).

PR #691, opened 2025-07-25
`SymEnv.ofEnv` is complete

zhengyao-lin:
new theorem has dropped :)
```
theorem ofEnv_completeness
  {Γ : TypeEnv} {env : Env}
  Γ.WellFormed →
  env.StronglyWellFormed →
  Env.EnumCompleteFor env (SymEnv.ofEnv Γ) →
  env ∈ᵢ SymEnv.ofEnv Γ →
  InstanceOfWellFormedEnvironment env.request env.entities Γ
```

This is the converse of #682.

A few caveats and tweaks:
- It turns out that the new condition `Env.EnumCompleteFor` is necessary, after a dicussion with @emina. This is because `SameEntities` does not enforce that all action entities exist in the concrete entity store, whereas `InstanceOfWellFormedEnvironment` does require that for typechecking soundness. However, while action and enum entities are distinguished at the level of `TypeEnv`, they are have similar representation at the level of `SymEnv`: so `SameEntities` cannot distinguish actions and enums. We decided to settle for this new precondition `Env.EnumCompleteFor` because other options (such as strengthening `InstanceOfWellFormedEnvironment` would break entity validation and potentially DRT).

  In the proof of completeness theorems of verification (e.g., `verifyEquivalent_is_complete`), the concretizer is used for constructing the witness `Env`, which means that it should satisfy `Env.EnumCompleteFor` (but I'll leave the proof for another PR).
- I added a new condition in `SameEntityData` for `members`; otherwise `SameEntities` may allow invalid action UIDs with a valid action entity type.
```
def SameEntityData (uid : EntityUID) (d : EntityData) (δ : SymEntityData) : Prop :=
  SameValues (.record d.attrs) (app δ.attrs uid) ∧
  (∀ anc, anc ∈ d.ancestors → InSymAncestors anc) ∧
  (∀ ancTy ancUF, δ.ancestors.find? ancTy = .some ancUF → InAncestors ancUF ancTy) ∧
  (∀ mems, δ.members = .some mems → uid.eid ∈ mems) ∧ /-- HERE --/
  SameTags uid d δ
```
Other proofs are fixed accordingly; @emina please check if this is ok.

### DIFF
--- a/cedar-lean/Cedar/Thm/SymCC/Env/Completeness.lean
+++ b/cedar-lean/Cedar/Thm/SymCC/Env/Completeness.lean
@@ -34,7 +34,7 @@ open Cedar.Data
 /--
 Inverse of `entity_uid_wf_implies_sym_entities_is_valid_entity_uid`
 -/
-theorem sym_entities_is_valid_entity_uid_implies_entity_uid_wf
+private theorem sym_entities_is_valid_entity_uid_implies_entity_uid_wf
   {Γ : TypeEnv} {uid : EntityUID}
   (hwf : Γ.WellFormed)
   (huid : (SymEnv.ofEnv Γ).entities.isValidEntityUID uid) :
@@ -93,7 +93,7 @@ theorem sym_entities_is_valid_entity_uid_implies_entity_uid_wf
   · contradiction
 
 /-- `TermPrim` case of `ofType_typeOf_pullback`. -/
-theorem ofType_typeOf_pullback_prim
+private theorem ofType_typeOf_pullback_prim
   {Γ : TypeEnv}
   {p : TermPrim} {ty : CedarType} {v : Value}
   (hwf_Γ : Γ.WellFormed)
@@ -165,7 +165,7 @@ theorem ofType_typeOf_pullback_prim
       constructor
       simp only [InstanceOfExtType]
 
-theorem value?_some_implies_typeOf_not_option
+private theorem value?_some_implies_typeOf_not_option
   {t : Term} {v : Value} {ty : TermType}
   (hsome : t.value? = .some v)
   (hopt : t.typeOf = .option ty) :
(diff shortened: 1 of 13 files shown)


cedar-lean/Cedar/Thm/SymCC/Env/Completeness.lean
```diff
@@ -0,0 +1,1308 @@
...
+    contradiction
+  | _ =>
+    try simp only [Term.value?] at hsome
+    try simp only [Term.typeOf] at hopt
+    contradiction
+
+/--
+If a term type is both the result of `TermType.ofType` and `Term.typeOf`
+for some well-formed `CedarType` and well-formed `Term`,
+then the if the the term is conretizable (via `value?`),
```
emina: Typo:  "then if the term is concretizable (via `value?`),"

cedar-lean/Cedar/Thm/SymCC/Term/ofType.lean
```diff
@@ -0,0 +1,33 @@
...
+
+import Cedar.Spec
+import Cedar.SymCC
+
+namespace Cedar.Thm
+
+open Validation SymCC Spec
+
+/-- Rephrases `ofRecordType` with map -/
+theorem ofRecordType_as_map (rty : List (Attr × QualifiedType)) :
```
emina, line 25: If this theorem is used in only one place, we should move it to the caller's file, unless we expect to write more theorems about `TermType.ofType`.

emina (changes requested):
Nice!

cedar-lean/Cedar/Thm/SymCC/Term/ofType.lean
zhengyao-lin: This is used in some other places too; but I haven't refactored them to use this theorem. Let me make that change.

zhengyao-lin: fixed!

chaluli (approved):
Looks great. Nice!

PR #699, opened 2025-07-31
Refactor partial evaluation to use `Residual` instead of `TypedExpr` as the input

oflatt:
This PR refactors partial evaluation to use a Residual instead of a TypedExpr. This allows partial evaluation to be run repeatedly, and also will enable proofs about the new batched evaluator (coming soon!)

An interesting part of this PR is the automation that turns calls to map1 into normal map calls. Normally this is a tedious process where you have to manually convert a closure into a slightly different one. This PR uses a Lean macro to solve the problem.


cedar-lean/Cedar/Thm/TPE/Conversion.lean
```diff
@@ -0,0 +1,622 @@
...
+          exact h₂
+        | toDays h₂ =>
+          apply ExtResidualWellTyped.toDays
+          rw [←conversion_preserves_typeof]
+          exact h₂
+termination_by sizeOf expr
+decreasing_by
+  all_goals (
+    simp
+    rename_i a b d d e f g h i j k l m n o
```
oflatt: This is horrible, I need advice for how to grab the hypothesis of the form `expr = ...` and use it 
Or we can manually write out all 19 goals

cedar-lean/Cedar/Thm/Data/List/Automation.lean
```diff
@@ -0,0 +1,120 @@
...
+  for termination proofs.
+
+  The tactic works by
+  1. Unfolding things like mapM₁, mapM₂, attach, ect
+  2. Finding an opportunity for simplification (a map on a pmap)
+  3. Rewritting using List.map_pmap_subtype and the simplified function.
+
+  If the tactic fails, you can try doing simplification first to expose these opportunities or using `find_pmap_func` to see what the computed function is.
+-/
+elab "auto_map₁_to_map" : tactic => do
```
oflatt: New automation-
Do we want this? I'm pretty happy with it but it's a complexity tradeoff.
If we want to keep it, I'll rewrite the current code everywhere to use it.

cedar-lean/Cedar/Thm/TPE/Conversion.lean
john-h-kastner-aws: sometimes I find using `first` works out nicer than `try` for this sort of pattern

oflatt: good idea, that's a little better (though the rename_i calls are still pretty ugly)

cdisselkoen: In other termination proofs in this codebase, we have the same problem and just use a lot of `_` for the ones the proof doesn't need to referemce

cdisselkoen: you can also look into using `rename` instead of `rename_i`, might or might not be better

cedar-lean/Cedar/Thm/Data/List/Automation.lean
zhengyao-lin: I agree that this is quite helpful!

Minor nit: maybe we want to name it as `auto_pmap_to_map` since it's handling more than just `map₁`?

Also does it automatically figure out the argument `f` of `mapM_pmap_subtype` in more complicated cases like here? https://github.com/cedar-policy/cedar-spec/blob/main/cedar-lean/Cedar/SymCC/Compiler.lean#L276

I sometimes have to specify `f` explicitly when using `simp` with `mapM_pmap_subtype`/`mapM₂_eq_mapM`, which is quite annoying and might affect proof stability; it'd be great if that can be automated!

oflatt: My hope is that people never have to think about pmap, since this macro desugars different maps then looks for pmap in order to do rewriting.

Yes it figure out what `f` is! That's the main reason it's helpful, since `f` is annoying to figure out.

zhengyao-lin: That's neat!

cedar-lean/Cedar/TPE/Residual.lean
```diff
@@ -64,6 +64,47 @@ def Residual.isError : Residual → Bool
...
+  | .unaryApp _ _ ty
+  | .binaryApp _ _ _ ty
+  | .getAttr _ _ ty
+  | .hasAttr _ _ ty
+  | .set _ ty
+  | .record _ ty
+  | .call _ _ ty
+  | .error ty => ty
+
+def Residual.toTypedExpr : Residual → TypedExpr
```
emina: Why do we need to convert Residuals to TypedExprs?  

The main motivation behind redefining TPE to work on Residuals was to avoid this conversion.

If we really do need to do this, then we should define this function to be of type `Residual → Option TypedExpr` rather than returning `true` when conversion is not possible.

This is the pattern we use in the rest of CedarLean to define what are effectively partial functions: they return either Result or Option.   

That does make proofs slightly less convenient but it makes the code more robust and easier to debug :)

cedar-lean/Cedar/Thm/Data/List/Automation.lean
emina: I'm going to vote against this automation :) 

One of our primary goals in the proofs is simplicity, ease of upgrades, and maintainability.  Automation that performs search, term transformations, etc, tends to be brittle and makes maintenance more difficult.  You'll see that in the rest of codebase, we limit automation to effectively proof templates (implemented with macros) and, of course, we use Lean's built-in tactics freely.

I understand the temptation to add more automation :). But in my experience, simplicity wins in the long term.

cedar-lean/Cedar/Thm/Data/List/Basic.lean
```diff
@@ -463,6 +463,18 @@ theorem mapM_key_id_sortedBy_key {α β : Type} [LT α] {ks : List α} {kvs : Li
 
     exact List.SortedBy.cons_cons hlt hs
 
+/-- if you use mapM on a list constructed using map
+    you can just do one mapM with a combined function
+    -/
+theorem mapM_then_map_combiner {α β γ ε} {f : α → β} {g : β → Except ε γ} {xs : List α} :
```
emina: Let's move this to Cedar.Thm.List.Lemmas, to the section containing the rest of `mapM` theorems: https://github.com/cedar-policy/cedar-spec/blob/ede2ca788bc24b6f290912f9025eeb9ccd178d7e/cedar-lean/Cedar/Thm/Data/List/Lemmas.lean#L347

cedar-lean/Cedar/Thm/TPE/Conversion.lean
```diff
@@ -0,0 +1,624 @@
...
+namespace Cedar.Thm
+
+open Cedar.TPE
+open Cedar.Spec
+open Cedar.Validation
+
+set_option pp.proofs true
+
+mutual
+theorem conversion_preserves_evaluation_forall2 :
```
emina, line 37: Similar comment to the earlier one:  why do we need to convert TypedExprs to Residuals?

cedar-lean/Cedar/TPE/Residual.lean
oflatt: Oops, I just left this in on accident.

oflatt: oops, I just left this in accidentally

cedar-lean/Cedar/Thm/TPE/Conversion.lean
oflatt: The `evaluatePolicy` entry point to partial evaluation produces a `TypedExpr` from a `Policy`, so we need to convert this to a `Residual` before calling the implementation of partial evaluation.

I think this will allow us to keep the same API in Rust but change the internals to be on `Residual`. We could also expose a new PE endpoint for residuals.

emina: Ah, sorry, I misread this as converting a Residual to TypedExpr!

cedar-lean/Cedar/Thm/Data/List/Automation.lean
oflatt: Okay! Removed the automation.

john-h-kastner-aws (approved):
looks good

PR #702, opened 2025-08-05
Modularize the generators

shaobo-he-aws:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cedar-policy-generators/src/schema_gen.rs
+++ b/cedar-policy-generators/src/schema_gen.rs
@@ -1,0 +1,35 @@
+use crate::{err::Result, schema::Schema};
+use arbitrary::Unstructured;
+use cedar_policy_core::{ast::EntityType, validator::ValidatorSchema as CoreSchema};
+
+pub trait SchemaGen: std::fmt::Debug {
+    fn arbitrary_entity_type(&self, u: &mut Unstructured<'_>) -> Result<EntityType>;
+}
+
+#[derive(Debug)]
+pub struct ValidatorSchema<'a> {
+    core_schema: &'a CoreSchema,
+    entity_types: Vec<EntityType>,
+}
+
+impl<'a> ValidatorSchema<'a> {
+    pub fn new(core_schema: &'a CoreSchema) -> Self {
+        let entity_types = core_schema.entity_type_names().cloned().collect();
+        Self {
+            core_schema,
+            entity_types,
+        }
+    }
+}
+
+impl SchemaGen for ValidatorSchema<'_> {
+    fn arbitrary_entity_type(&self, u: &mut Unstructured<'_>) -> Result<EntityType> {
+        Ok(u.choose(&self.entity_types)?.clone())
+    }
+}
+
+impl SchemaGen for Schema {
+    fn arbitrary_entity_type(&self, u: &mut Unstructured<'_>) -> Result<EntityType> {
+        Ok(u.choose(&self.entity_types)?.clone())
+    }
+}
--- a/cedar-policy-generators/src/expr.rs
+++ b/cedar-policy-generators/src/expr.rs
@@ -20,8 +20,9 @@ use crate::err::{while_doing, Error, Result};
 use crate::hierarchy::{generate_uid_with_type, Hierarchy};
 use crate::schema::{
     attrs_from_attrs_or_context, entity_type_name_to_schema_type, lookup_common_type,
-    uid_for_action_name, Schema,
+    uid_for_action_name,
 };
+use crate::schema_gen::SchemaGen;
 use crate::settings::ABACSettings;
 use crate::size_hint_utils::{size_hint_for_choose, size_hint_for_range, size_hint_for_ratio};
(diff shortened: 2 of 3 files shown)


cedar-policy-generators/src/hierarchy.rs
```diff
@@ -435,10 +435,9 @@ impl HierarchyGenerator<'_, '_> {
             .iter()
             .map(|name| {
                 let (name, uid_choices) = match &self.mode {
-                    HierarchyGeneratorMode::SchemaBased { schema } => (
-                        name.qualify_with(schema.namespace()),
-                        schema.get_uid_enum_choices(name),
-                    ),
+                    HierarchyGeneratorMode::SchemaBased { schema } => {
+                        (name.clone(), schema.get_uid_enum_choices(name))
+                    }
```
cdisselkoen, line 440: why do we no longer need `.qualify_with()`?

cedar-policy-generators/src/schema.rs
```diff
@@ -1619,33 +1583,23 @@ impl Schema {
...
+                        EntityType::from_normalized_str("Action")
+                            .unwrap()
+                            .qualify_with(self.namespace()),
+                        eid.clone(),
+                        None,
+                    )
+                })
+                .collect::<Vec<_>>(),
+        )?
+        .clone())
```
cdisselkoen, line 1602: Doesn't have to be addressed in this PR, but this is pretty inefficient, it builds all Action EntityUIDs and collects them all into a Vec only to clone one of its choice and throw away the whole Vec. More efficient would be to first `u.choose()` an EID and then only build the `EntityUID` for that particular chosen EID

cedar-policy-generators/src/schema_gen.rs
```diff
@@ -0,0 +1,606 @@
...
+        &self,
+        ety: &EntityType,
+    ) -> Option<Box<dyn Iterator<Item = EntityType> + '_>>;
+    /// Get attribute types and whether there are additional attributes
+    fn attribute_by_entity_type(
+        &self,
+        ety: &EntityType,
+    ) -> Option<(BTreeMap<SmolStr, QualifiedType>, bool)>;
+    /// Get tag type of an entity type
+    fn tag_type_by_entity_type(&self, ety: &EntityType) -> Option<Option<abac::Type>>;
```
cdisselkoen, line 122: can the doc string here specify what return values of `None` and `Some(None)` mean and how they are different?

cedar-policy-generators/src/schema_gen.rs
```diff
@@ -0,0 +1,606 @@
...
+                doing_what: format!("getting arbitrary attr for type {target_type:?}"),
+            }),
+        }
+    }
+    fn arbitrary_entity_type_with_tag_type(
+        &self,
+        target_type: &abac::Type,
+        u: &mut Unstructured<'_>,
+    ) -> Result<ast::EntityType> {
+        let candidates: Vec<ast::EntityType> = self
```
cdisselkoen: nit: can we collect into `Vec<&ast::EntityType>` in order to avoid some clones? or does that not work?

cedar-policy-generators/src/schema_gen.rs
```diff
@@ -0,0 +1,606 @@
...
+            &self
+                .core_schema
+                .get_entity_type(entity_type)
+                .expect("entity type should exist")
+                .attributes()
+                .iter()
+                .map(|a| a.0.clone())
+                .collect::<Vec<_>>(),
+        )?
+        .clone())
```
cdisselkoen, line 382: nit: similar inefficiency here to one of my above comments

cdisselkoen (approved):
Seems like a reasonable refactor to me

cedar-drt/fuzz/fuzz_targets/input-generation.rs
```diff
@@ -0,0 +1,126 @@
...
+            schema::Schema::arbitrary_request_size_hint(depth),
+            schema::Schema::arbitrary_request_size_hint(depth),
+            schema::Schema::arbitrary_request_size_hint(depth),
+            schema::Schema::arbitrary_request_size_hint(depth),
+            schema::Schema::arbitrary_request_size_hint(depth),
+        ]))
+    }
+}
+
+// Type-directed fuzzing of ABAC hierarchy/policy/requests.
```
john-h-kastner-aws: Can you elaborate a bit more on what this new target is testing?

shaobo-he-aws: This target actually is not used for testing. Instead, it's intended to generate requests/entities based on a schema and a policy.

shaobo-he-aws: Will put down more comments about it.

cedar-policy-generators/src/hierarchy.rs
shaobo-he-aws: This is because the `entity_types` method of trait `SchemaGen` should return an iterator of fully qualified entity type names.

PR #703, opened 2025-08-05
Added a target to DRT terms

shaobo-he-aws:
*Issue #, if available:*

*Description of changes:*


cedar-drt/fuzz/fuzz_targets/symcc-term-drt.rs
```diff
@@ -0,0 +1,137 @@
...
+                compile_rust_policy(&policy, &schema, &req_env),
+            ) {
+                (Ok(Ok(lean_asserts)), Ok(rust_asserts)) => {
+                    let lean_asserts = lean_asserts
+                        .into_iter()
+                        .map(|t| Term::try_from(t).expect("term conversion should succeed"))
+                        .collect::<BTreeSet<_>>();
+                    let rust_asserts =
+                        BTreeSet::from_iter(rust_asserts.as_ref().into_iter().cloned());
+                    assert_eq!(
```
john-h-kastner-aws: should use `similiar_asserts::assert_eq!` for nice diff

cedar-drt/fuzz/fuzz_targets/symcc-term-drt.rs
```diff
@@ -0,0 +1,137 @@
...
+                (Ok(Ok(_)), Err(err)) => {
+                    panic!("Rust errors while Lean does not: {err}")
+                }
+                (Err(err), Ok(_)) => {
+                    panic!("Lean errors while Rust does not: {err}")
+                }
+                (Ok(Err(err)), Ok(_)) => {
+                    panic!("Lean errors while Rust does not: {err}")
+                }
+                (_, _) => {}
```
john-h-kastner-aws: Can you add a comment to address what case were ignoring? (look like it's just `Err, Err`)

cedar-drt/fuzz/fuzz_targets/symcc-term-drt.rs
```diff
@@ -0,0 +1,137 @@
...
+            HierarchyGenerator::size_hint(depth),
+            schema::Schema::arbitrary_policy_size_hint(&SETTINGS, depth),
+            schema::Schema::arbitrary_request_size_hint(depth),
+            schema::Schema::arbitrary_request_size_hint(depth),
+            schema::Schema::arbitrary_request_size_hint(depth),
+            schema::Schema::arbitrary_request_size_hint(depth),
+            schema::Schema::arbitrary_request_size_hint(depth),
+            schema::Schema::arbitrary_request_size_hint(depth),
+            schema::Schema::arbitrary_request_size_hint(depth),
+            schema::Schema::arbitrary_request_size_hint(depth),
```
john-h-kastner-aws: Shouldn't need these

cedar-drt/fuzz/fuzz_targets/symcc-term-drt.rs
john-h-kastner-aws: or actually, it's `Err(_), Err(_)` and `Ok(Err(_)), Err(_)`. IIUC, `Err(_)` is an error in the ffi layer, so it should probably cause the target to fail. Someone who knows the lean ffi better should comment.

shaobo-he-aws: I thought so too but it turns out the FFI also validates the request environment and raises an error if it fails.

cdisselkoen: I agree it would be better to spell the case(s) out explicitly and comment on why we're ignoring them

chaluli: I'll double check. The outer error was meant to be an FFI error while the inner error was meant to be a SymCC error. I could have done validation before calling and that would also call an FFI error. If so, should probably make that a different error variant so it can be matched on.

shaobo-he-aws:
Closed in favor of #708

PR #710, opened 2025-08-12
Fix deserialization of bitvec

shaobo-he-aws:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cedar-lean-ffi/src/datatypes.rs
+++ b/cedar-lean-ffi/src/datatypes.rs
@@ -1,5 +1,6 @@
 use cedar_policy::{Decision, EntityId, EntityTypeName, EntityUid, PolicyId};
 use num_bigint::ParseBigIntError;
+use serde::de::{SeqAccess, Visitor};
 use serde::{Deserialize, Deserializer};
 use smol_str::SmolStr;
 use thiserror::Error;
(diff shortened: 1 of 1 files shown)


cedar-lean-ffi/src/datatypes.rs
```diff
@@ -313,8 +313,8 @@ pub enum Op {
 
 #[derive(Debug, Deserialize)]
 pub struct Bitvec {
-    pub width: u8,
-    pub val: String,
+    pub size: u8,
+    pub value: String,
```
cdisselkoen: Is there a difference between the meaning of `size` and `width`, or is this just a rename?  The Rust uses "width", and I think that is @lianah's preference?

cdisselkoen (approved):

cedar-lean-ffi/src/datatypes.rs
chaluli: You could keep the name width and add a #[serde(rename = "size")] to the width parameter so that it parses the Lean JSON message correctly.

shaobo-he-aws: Done.

chaluli (approved):
Looks great!

PR #712, opened 2025-08-14
Add a DRT target to DRT Term to SMT script translation

shaobo-he-aws:
*Issue #, if available:*

*Description of changes:*


cedar-drt/fuzz/fuzz_targets/symcc-smt-script-drt.rs
```diff
@@ -0,0 +1,144 @@
...
+        depth: usize,
+    ) -> std::result::Result<(usize, Option<usize>), MaxRecursionReached> {
+        Ok(arbitrary::size_hint::and_all(&[
+            schema::Schema::arbitrary_size_hint(depth)?,
+            HierarchyGenerator::size_hint(depth),
+        ]))
+    }
+}
+
+fn compile_rust_policy<'a>(
```
chaluli: Given multiple targets needs this function now. Would it be good to move it to `cedar-drt/fuzz/src/symcc.rs` or similar file so that it can be shared amongst multiple fuzz targets.

cedar-lean-ffi/src/datatypes.rs
```diff
@@ -703,3 +917,16 @@ mod deserialization {
             serde_json::from_value(json).expect("deserialization should succeed");
     }
 }
+
+#[cfg(test)]
```
chaluli, line 921: I appreciate the test for pattern. But is there a specific reason why we only test the conversion for pattern and not for other types?

Is it an especially hard version to get right?

shaobo-he-aws: Yes. It involves `char` to `u32` conversion and vice versa. So I wasn't sure.

cedar-drt/fuzz/fuzz_targets/symcc-smt-script-drt.rs
shaobo-he-aws: Done.

PR #714, opened 2025-08-16
Refactor FFI to use less unsafe code

shaobo-he-aws:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cedar-lean/CedarFFI/Main.lean
+++ b/cedar-lean/CedarFFI/Main.lean
@@ -280,25 +280,20 @@ def parseCheckAssertsReq (proto : ByteArray) : Except String (Cedar.SymCC.Assert
 /--
   Run `solver` on `vcs` without exposing the IO monad to the calling code
 -/
-private unsafe def unsafeSolve {α} (solver : IO Solver) (vcs : SolverM α) : Except String α := do
-    match unsafeIO solver with
-    | .error _ => .error "error creating solver"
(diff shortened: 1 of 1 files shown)


cedar-lean/CedarFFI/Main.lean
```diff
@@ -58,6 +60,12 @@ def runAndTimeIO (f : IO α) : IO (Timed α) := do
     duration := stop - start
   }
 
+unsafe def runFfiM {α : Type} [Lean.ToJson α] (m : FfiM α) : String :=
+  match unsafeIO m with
+  | .error s => toString (Lean.toJson ((.error s!"IO error: {s}") : Except String α))
+  | .ok (.error s) => toString (Lean.toJson ((.error s) : Except String α))
+  | .ok (.ok r) => toString (Lean.toJson (.ok r : Except String α))
```
cdisselkoen, line 67: @emina will disagree, but here is a great example of where I think the `$` operator would improve readability.  It's hard for me to read at a glance what the `: Except String α` applies to, I have to carefully count parens.  With `$` we could write the first line as something like
```
| .error s => toString $ Lean.toJson $ (.error s!"IO error: {s}") : Except String α
```

emina: Indeed, I personally find the parenthesized version easier to read :D

PR #718, opened 2025-08-20
Add check with cex functions to FFI

shaobo-he-aws:
*Issue #, if available:*

*Description of changes:*
We currently do not perform deserialization of `Env` because we will need to serialize `Env` properly, e.g., according to the entity format.

### DIFF
--- a/cedar-lean/CedarFFI/Main.lean
+++ b/cedar-lean/CedarFFI/Main.lean
@@ -261,7 +261,7 @@ opaque timedSolve {α} (solver : IO Solver) (vcs : SolverM α) : IO (Except Stri
   returns JSON encoded string that encodes
   1.) .error err_message if there was in error in parsing or running the solver
   2.) .ok { data := true, duration := <encode+solve_time> } if the solver could prove `req` holds
-  3.) .ok { data := true, duration := <encode+solve_time> } if the solver could prove `req` does not hold
+  3.) .ok { data := false, duration := <encode+solve_time> } if the solver could prove `req` does not hold
 -/
 @[export runCheckNeverErrors] unsafe def runCheckNeverErrors (req : ByteArray) : String :=
   runFfiM do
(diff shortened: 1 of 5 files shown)


chaluli (approved):
I think this is good as is. Though, I think we should perhaps add an issue to add a Cex type for properly deserializing CEXs and provide better ergonomics.

cedar-lean/CedarFFI/Main.lean
```diff
@@ -268,6 +268,19 @@ opaque timedSolve {α} (solver : IO Solver) (vcs : SolverM α) : IO (Except Stri
...
     timedSolve Solver.cvc5 (checkNeverErrors policy εnv)
 
+/--
+  `req`: binary protobuf for an `CheckPolicyRequest`
+
+  returns JSON encoded string that encodes
+  1.) .error err_message if there was in error in parsing or running the solver
+  2.) .ok { data := true, duration := <encode+solve_time> } if the solver could prove `req` holds
+  3.) .ok { data := true, duration := <encode+solve_time> } if the solver could prove `req` does not hold
+-/
```
cdisselkoen: where is the cex returned?

cedar-lean/CedarFFI/Main.lean
```diff
@@ -268,6 +268,19 @@ opaque timedSolve {α} (solver : IO Solver) (vcs : SolverM α) : IO (Except Stri
     let (policy, εnv) ← parseCheckPolicyReq req false
     timedSolve Solver.cvc5 (checkNeverErrors policy εnv)
 
+/--
+  `req`: binary protobuf for an `CheckPolicyRequest`
+
+  returns JSON encoded string that encodes
+  1.) .error err_message if there was in error in parsing or running the solver
+  2.) .ok { data := true, duration := <encode+solve_time> } if the solver could prove `req` holds
+  3.) .ok { data := true, duration := <encode+solve_time> } if the solver could prove `req` does not hold
```
chaluli: This comment doesn't seem to be correct.

shaobo-he-aws: Yep. Will fix it.

shaobo-he-aws: Done.

cedar-lean/CedarFFI/Main.lean
shaobo-he-aws: Fixed.

PR #719, opened 2025-08-21
Limit the number of iterations in target `symcc-*-drt`

shaobo-he-aws:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cedar-drt/fuzz/fuzz_targets/symcc-smt-script-drt.rs
+++ b/cedar-drt/fuzz/fuzz_targets/symcc-smt-script-drt.rs
@@ -43,6 +43,9 @@ static RUNTIME: LazyLock<tokio::runtime::Runtime> = LazyLock::new(|| {
         .unwrap()
 });
 
+/// Limit the number of request environments so that we don't produce slow units
+const MAX_REQ_ENV_NUM: usize = 128;
+
 /// Input expected by this fuzz target
 #[derive(Debug, Clone)]
 pub struct FuzzTargetInput {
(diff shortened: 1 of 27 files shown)


cedar-drt/fuzz/fuzz_targets/symcc-smt-script-drt.rs
```diff
@@ -43,6 +43,9 @@ static RUNTIME: LazyLock<tokio::runtime::Runtime> = LazyLock::new(|| {
         .unwrap()
 });
 
+/// Limit the number of request environments so that we don't produce slow units
+const MAX_REQ_ENV_NUM: usize = 128;
```
john-h-kastner-aws: Don't we already do this a different way in some validation targets?  We should have one mechanism to configure this

john-h-kastner-aws: found it. We could make [PER_ACTION_REQUEST_ENV_LIMIT](https://github.com/cedar-policy/cedar-spec/blob/e88ee623d5805730b9466ca2103fdd7b5f5ff894/cedar-policy-generators/src/schema.rs#L954) configurable in `ABACSettings` or something

chaluli: Agreed, I think it would be better to limit # of req_env than to not cover all of the req_env.

chaluli (changes requested):
Looks good. But I think we should definitely follow John's suggestion to re-use the mechanism used to limit request environments for some of the Validation targets.

cedar-drt/fuzz/src/symcc.rs
```diff
@@ -90,6 +90,11 @@ pub fn compile_policies<'a>(
     Ok(func(&well_typed_policies, &sym_env)?)
 }
 
+/// The limit on the total number of request envs specific to symcc
+pub const fn total_action_request_env_limit() -> usize {
```
john-h-kastner-aws, line 94: Any particular difference between using a `const` and a `const fn` for this? (I'm fine with either)

shaobo-he-aws: No. This is just to be consistent with `ABACSetting`'s `default_*` functions.

chaluli (approved):
Thanks for the changes!

PR #724, opened 2025-08-25
rename RequestType to ActionSignature

cdisselkoen:
Renames `RequestType` to `ActionSignature`, to align some names better between our Rust and Lean code.  Also renames related variables and field names (e.g. `reqty` to `sig`).

### DIFF
--- a/cedar-lean/Cedar/Thm/Validation/Typechecker/Types.lean
+++ b/cedar-lean/Cedar/Thm/Validation/Typechecker/Types.lean
@@ -76,11 +76,11 @@ inductive InstanceOfType (env : TypeEnv) : Value → CedarType → Prop where
       (h₁ : InstanceOfExtType x xty) :
       InstanceOfType env (.ext x) (.ext xty)
 
-def InstanceOfRequestType (request : Request) (env : TypeEnv) : Prop :=
-  InstanceOfEntityType request.principal env.reqty.principal env ∧
-  request.action = env.reqty.action ∧
-  InstanceOfEntityType request.resource env.reqty.resource env ∧
-  InstanceOfType env request.context (.record env.reqty.context)
+def InstanceOfTypeEnv (request : Request) (env : TypeEnv) : Prop :=
+  InstanceOfEntityType request.principal env.sig.principal env ∧
+  request.action = env.sig.action ∧
+  InstanceOfEntityType request.resource env.sig.resource env ∧
+  InstanceOfType env request.context (.record env.sig.context)
 
 def InstanceOfEntityTags (data : EntityData) (entry : EntitySchemaEntry) (env : TypeEnv) : Prop :=
   match entry.tags? with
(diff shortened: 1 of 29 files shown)


cedar-lean/Cedar/Thm/Validation/Typechecker/Types.lean
```diff
@@ -76,11 +76,11 @@ inductive InstanceOfType (env : TypeEnv) : Value → CedarType → Prop where
...
-def InstanceOfRequestType (request : Request) (env : TypeEnv) : Prop :=
-  InstanceOfEntityType request.principal env.reqty.principal env ∧
-  request.action = env.reqty.action ∧
-  InstanceOfEntityType request.resource env.reqty.resource env ∧
-  InstanceOfType env request.context (.record env.reqty.context)
+def InstanceOfTypeEnv (request : Request) (env : TypeEnv) : Prop :=
+  InstanceOfEntityType request.principal env.sig.principal env ∧
+  request.action = env.sig.action ∧
+  InstanceOfEntityType request.resource env.sig.resource env ∧
+  InstanceOfType env request.context (.record env.sig.context)
```
cdisselkoen, line 83: This was arguably misnamed previously -- it defines alignment with a `TypeEnv`, not just a `RequestType`.  I renamed to `InstanceOfTypeEnv` rather than `InstanceOfActionSignature`.  Note that `RequestType`/`ActionSignature` is not one of the parameters here

john-h-kastner-aws:
`ActionSignature` doesn't feel right since a single action signature doesn't completely describe an action. Some variant of `*Env` feels better to me since it's purpose is to provide type bindings for the scope variables.

For completeness, I see that the Rust also defines, [`RequestSchema`](https://github.com/cedar-policy/cedar/blob/86677b7eaf39c40f823be28467d1c1320f7ea05c/cedar-policy-core/src/ast/request.rs#L621),  [`RequestType`](https://github.com/cedar-policy/cedar/blob/86677b7eaf39c40f823be28467d1c1320f7ea05c/cedar-policy-core/src/ast/request.rs#L57), and [another `RequestType`](https://github.com/cedar-policy/cedar/blob/86677b7eaf39c40f823be28467d1c1320f7ea05c/cedar-policy-symcc/src/symcc/env.rs#L526).

cdisselkoen:
The `context` field is redundant because that information already exists in the schema given the other three fields.

emina:
That's true.  We keep it around in Lean because it would be pretty annoying to have to constantly lookup the `context` via schema (and always have schema around just to be able to look up `context`).

john-h-kastner-aws:
> Should we just leave it at RequestType?

I'm happy with this, but then we'll have three `RequestType`s in the Rust env :P. Hopefully we could unify those

emina:
> we'll have three `RequestType`s in the Rust env :P. Hopefully we could unify those

Yep, 1 > 3 in this case :))

PR #725, opened 2025-08-28
Prove batched evaluation (with iteration limit) is correct

oflatt:
This PR proves that batched evaluation is equivalent to normal evaluation


cedar-lean/Cedar/TPE/BatchedEvaluator.lean
```diff
@@ -34,22 +34,24 @@ abbrev EntityLoader := Set EntityUID → Map EntityUID PartialEntityData
...
   (residual : Residual)
   (req : Request)
   (loader : EntityLoader)
   (store : PartialEntities)
-  : Result Value :=
-  let toLoad := residual.allLiteralUIDs.filter (λ uid => (store.find? uid).isNone)
-  let newEntities := loader toLoad
-  let newStore := newEntities ++ store
-  let newRes := Cedar.TPE.evaluate residual req.asPartialRequest newStore
+  (iters : Nat)
```
emina: If the last argument is an inductive type, we usually write the function like this:

```
def batchedEvalLoop
  (residual : Residual)
  (req : Request)
  (loader : EntityLoader)
  (store : PartialEntities) 
  : Nat → Residual
  | 0     => residual
  | n + 1 => ...
```

cedar-lean/Cedar/Thm/Data/Map.lean
```diff
@@ -1306,160 +1306,124 @@ theorem wellFormed_correct {α β} [LT α] [StrictLT α] [DecidableLT α] {m : M
...
 
 
-
-theorem map_make_find_in_tail
-  [LT α] [StrictLT α] [DecidableEq α]
-  [DecidableLT α] [LawfulBEq α]
+/--
+Helper for make_find?_eq_list_find?
+-/
+private theorem map_make_find_in_tail
```
emina: Naming convention:  we keep the `?` in `find?` when it appears in theorem names.  Let's change the name of this theorem (and others in this file) to match the convention.

cedar-lean/Cedar/Thm/TPE/BatchedEvaluator.lean
```diff
@@ -0,0 +1,237 @@
...
+The main correctness theorem for batched evaluation:
+Batched evaluation with an entity loader produces the same result
+as normal evaluation with the complete entity store.
+-/
+theorem batched_eval_eq_evaluate
+  {x : TypedExpr}
+  {req : Request}
+  {es : Entities}
+  {env : TypeEnv} :
+  let loader := entityLoaderFor es
```
emina: We can simplify the code and this theorem by getting rid of the `loader` argument to `BatchedEvaluate`.  

Currently, this argument is always bound to the same loader function (`entityLoaderFor es`), and the theorem / proofs are done only with respect to that function.  

Alternatively, if we want to keep the more abstract concept of the loader, we'll want this theorem to be more general.  Specifically, it would need to define the concept of a "well-formed loader for `es`" and then show that all well-formed loader functions lead to correct results.

The main question is whether there is any practically interesting function other than `entityLoaderFor es` that works correctly with this `batchedEvaluate`.  

If not, I would suggest simplifying `batchedEvaluate` to not take the `loader` as an argument, and just call `loadEntities es uids` directly (where `loadEntities es uids` has the same body as `entityLoaderFor`).

oflatt: Good idea! I think we'll have to model the loader if we extend batched evaluation so you can load partial entities at a time (for example skipping ancestors)
But for now I'll inline it

oflatt: Actually, I realized that you can have a variety of other implementations of entity loader, including providing extra entities besides those requested. I changed the spec for the loader to say that your entities must refine the entities in the store. This actually makes the proof easier

It's also sound to provide partial entities (missing attributes or ancestors) but the algorithm might not converge to a value if you do that.

Perhaps we should change the algorithm to re-request entities that are still partial and appear in the program?

emina: Nice!  Yes, I think we should change the algorithm to support the most general (least constrained) loader possible that (a) leads to correct output and (b) converges.

cedar-lean/Cedar/Thm/TPE/BatchedEvaluator.lean
```diff
@@ -18,6 +18,10 @@ open Cedar.Thm
 open Cedar.Data
 
 
+inductive EntityLoader.WellBehaved (store: Entities) : EntityLoader → Prop
```
emina: Why make this an inductive type rather than a function?

```
def EntityLoader.WellBehaved (store: Entities) (loader: EntityLoader) : Prop := 
  ∀ s, EntitiesRefine store (loader s)
```

oflatt: good point

emina (approved):

cedar-lean/Cedar/TPE/BatchedEvaluator.lean
```diff
@@ -28,28 +28,43 @@ open Cedar.Spec
...
+  d.asPartial
+
 
 /--
 The batched evaluation loop
   1. Asks for any new entities referenced by the residual
   2. Partially evaluates now that new entities are loaded
-  3. Exits if a value has been found
+  3. Exits if a value has been found or iterations are exhausted
 -/
```
cdisselkoen: Might be worth documenting that this is a _maximum_ iterations, i.e., the loop won't always execute the full number of iterations.  Either here or on `batchedEvaluate`

cedar-lean/Cedar/Thm/Data/Map.lean
```diff
@@ -1306,160 +1306,124 @@ theorem wellFormed_correct {α β} [LT α] [StrictLT α] [DecidableLT α] {m : M
...
+  simp
   constructor
   case mp =>
-    exact list_find?_implies_make_find?
+    intro h
+    exists k
   case mpr =>
-    exact make_find?_implies_list_find?
+    intro h
+    rcases h with ⟨k₂, h⟩
```
cdisselkoen: this can be rewritten `replace ⟨k₂, h⟩ := h`. At some point I think our guidance was to use `rcases` only when the pattern contains `|`, i.e., only when it causes splits.  For simple destructuring it's overkill

cedar-lean/Cedar/Thm/Data/Map.lean
```diff
@@ -1700,4 +1646,22 @@ theorem toList_congr
...
+    rfl
+  rw [h_append_eq]
+  rw [make_find?_eq_list_find?]
+  rw [List.find?_append]
+  simp [Map.find?, Option.map, Option.or]
+  cases List.find? (fun x => x.fst == k) m₁.kvs
+  . simp
+    cases List.find? (fun x => x.fst == k) m₂.kvs
+    all_goals
+      simp
```
cdisselkoen:
```suggestion
    cases List.find? (fun x => x.fst == k) m₂.kvs <;> simp
```

cedar-lean/Cedar/Thm/TPE/BatchedEvaluator.lean
```diff
@@ -0,0 +1,186 @@
...
+open Cedar.Data
+
+
+/-- A well behaved entity loader
+1. Loads all the requested entities, returning none for missing
+entities
+2. Refines the backing entity store
+
+The first condition is required for convergence of
+batched evaluation, which has not be proven. It is unused
```
cdisselkoen:
```suggestion
batched evaluation, which has not been proven. It is unused
```

cedar-lean/Cedar/Thm/TPE/BatchedEvaluator.lean
```diff
@@ -0,0 +1,186 @@
...
+in the code base at the moment.
+-/
+abbrev EntityLoader.WellBehaved (store: Entities) (loader: EntityLoader) : Prop :=
+  ∀ s, s ⊆ (loader s).keys ∧
+       EntitiesRefine store ((loader s).mapOnValues EntityDataOption.asPartial)
+
+theorem as_partial_request_refines {req : Request} :
+  RequestRefines req (Request.asPartialRequest req) := by
```
cdisselkoen:
```suggestion
  RequestRefines req req.asPartialRequest := by
```

cedar-lean/Cedar/Thm/TPE/BatchedEvaluator.lean
```diff
@@ -0,0 +1,186 @@
...
+  intro h₁ h₂ h₃
+  have h₄ := (direct_request_and_entities_refine req es)
+
+  let first_partial := (TPE.evaluate (TypedExpr.toResidual x) (Request.asPartialRequest req) (Entities.asPartial (Data.Map.mk [])))
+  let h₅ : Residual.WellTyped env (TypedExpr.toResidual x) := by {
+    apply conversion_preserves_typedness
+    exact h₂
+  }
+  have conversion_sound := conversion_preserves_evaluation x req es
+  rw [conversion_sound]
```
cdisselkoen:
```suggestion
  rw [conversion_preserves_evaluation x req es]
```

cedar-lean/Cedar/Thm/TPE/BatchedEvaluator.lean
```diff
@@ -0,0 +1,186 @@
...
+  have partial_sound := partial_evaluate_is_sound h₅ h₃ h₄
+  rw [partial_sound]
```
cdisselkoen:
```suggestion
  rw [partial_evaluate_is_sound h₅ h₃ h₄]
```

cedar-lean/Cedar/Thm/TPE/BatchedEvaluator.lean
```diff
@@ -0,0 +1,186 @@
...
+
+  have h₆ : Residual.WellTyped env (TPE.evaluate x.toResidual req.asPartialRequest Map.empty) := by
+    apply partial_eval_preserves_well_typed h₃
```
cdisselkoen:
```suggestion
    apply partial_eval_preserves_well_typed h₃ _ h₅
```
then you'll be left with just the nontrivial case

cdisselkoen (approved):

cedar-lean/Cedar/Thm/TPE/BatchedEvaluator.lean
oflatt: neat

cedar-lean/Cedar/Thm/Data/Map.lean
oflatt: good idea

emina (approved):

PR #726, opened 2025-08-29
Set lower limits for the symcc term roundtrip target

adpaco-aws:
In #719 we set lower limits specific to symcc targets. However, the `symcc-term-roundtrip-abac-type-directed` target has resulted in multiple slow units over the last few days. My investigation has revealed that the time we spend on this target for each request environment is close to 350ms. This PR sets lower limits for total action/request environment pairs so that it's easier to stay within bounds. We could also update the symcc limits if we don't want to have too many ad-hoc limits.

### DIFF
--- a/cedar-drt/fuzz/fuzz_targets/symcc-term-roundtrip-abac-type-directed.rs
+++ b/cedar-drt/fuzz/fuzz_targets/symcc-term-roundtrip-abac-type-directed.rs
@@ -39,6 +39,9 @@ pub struct FuzzTargetInput {
     pub policy: ABACPolicy,
 }
 
+const SYMCC_TERM_ROUNDTRIP_PER_ACTION_REQUEST_ENV_LIMIT: usize = 4;
+const SYMCC_TERM_ROUNDTRIP_TOTAL_ACTION_REQUEST_ENV_LIMIT: usize = 24;
+
 /// settings for this fuzz target
 const SETTINGS: ABACSettings = ABACSettings {
     match_types: true,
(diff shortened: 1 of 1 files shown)


john-h-kastner-aws:
There must be something were doing wrong if it's ~100ms per FFI call. The authorization/validation DRT targets can't be spending so much time calling the Lean mode, right?

adpaco-aws:
Yes, we should be concerned about this performance. After spending some time looking at the FFI code, I think the main difference is that SymCC targets are generating/returning more complex data structures, whereas the authz/validation targets only do responses.

Unfortunately, I can't spend more time looking into this, but if I were to continue I'd probably compare the time we're spending in the model vs. Lean/Rust conversions. We should figure out how expensive it is to perform symbolic compilation, and determine if these proportions make sense.

Regarding these lower limits, it's probably better to keep these targets as they are in case they catch other issues. In other words, we can tolerate slow units for now if they're going to help us test more inputs. What do you think?

adpaco-aws:
Right, the up-to-3 FFI calls per environment should be noticeable w.r.t. other targets. One obvious optimization in this target to expose the conversion to object for `schema` and `req_env` so we only do it once, or cache them in memory so we don't actually do the conversion in the FFI layer when we see the same input.

shaobo-he-aws (approved):

adpaco-aws:
Closing this since the optimization in #732 is expected to result in a speed-up for FFI calls (from ~100ms down to ~50ms). This likely won't prevent slow units from happening: even with the speed-up, some of these tests would still qualify as slow units. But as mentioned earlier, that's acceptable for now if it allows us to test more complex inputs. If possible, we should continue improving the FFI layer as much as possible if we want to avoid them while keeping the current limits.

PR #728, opened 2025-09-03
Make the generators driven by Cedar types instead of json schema types

shaobo-he-aws:
*Issue #, if available:*

*Description of changes:*

This PR makes the generators driven by a `Type` used by the Lean validator.It moves the generators closer to being a standalone library because `Type` can be easily derived from a type used by the validator. In other words, users can easily provide custom Cedar schemas.

### DIFF
--- a/cedar-policy-generators/src/abac.rs
+++ b/cedar-policy-generators/src/abac.rs
@@ -31,7 +31,6 @@ use std::cell::RefCell;
 use std::collections::BTreeMap;
 use std::net::{Ipv4Addr, Ipv6Addr};
 use std::ops::{Deref, DerefMut};
-use std::sync::Arc;
 use thiserror::Error;
 
 // Mutate a hypothetically valid string (randomly).
@@ -699,7 +698,7 @@ impl AvailableExtensionFunctions {
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Arbitrary)]
 pub struct QualifiedType {
     /// Type
-    pub ty: Arc<Type>,
+    pub ty: Type,
     /// Qualification
     pub required: bool,
 }
@@ -810,7 +809,7 @@ impl Type {
             Type::string(),
             Type::set_of(Self::arbitrary_nonextension(u)?),
             Self::arbitrary_record_inner(u, None, |u| Ok(QualifiedType {
-                ty: Arc::new(Self::arbitrary_nonextension(u)?),
+                ty: Self::arbitrary_nonextension(u)?,
                 required: u.arbitrary()?
             }))?,
             Type::entity(u.arbitrary()?)
--- a/cedar-policy-generators/src/expr.rs
+++ b/cedar-policy-generators/src/expr.rs
@@ -29,7 +29,6 @@ use cedar_policy_core::ast;
 use cedar_policy_core::validator::json_schema::{EntityTypeKind, StandardEntityType};
 use smol_str::SmolStr;
 use std::collections::BTreeMap;
-use std::sync::Arc;
 
 /// Struct for generating expressions
 #[derive(Debug)]
@@ -1754,7 +1753,7 @@ fn record_type_with_attr(attr_name: SmolStr, attr_type: Type) -> Type {
     Type::Record(BTreeMap::from_iter([(
         attr_name,
         QualifiedType {
-            ty: Arc::new(attr_type),
+            ty: attr_type,
             required: true,
         },
     )]))
--- a/cedar-policy-generators/src/schema.rs
+++ b/cedar-policy-generators/src/schema.rs
@@ -40,7 +40,6 @@ use cedar_policy_core::validator::{
 };
 use smol_str::{SmolStr, ToSmolStr};
 use std::collections::BTreeMap;
-use std::sync::Arc;
 
 /// Contains the schema, but also pools of constants etc
 #[derive(Debug, Clone)]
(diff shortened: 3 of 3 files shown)


cedar-policy-generators/src/abac.rs
```diff
@@ -693,6 +694,15 @@ impl AvailableExtensionFunctions {
     }
 }
 
+/// A qualified type
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Arbitrary)]
+pub struct QualifiedType {
+    /// Type
+    pub ty: Box<Type>,
```
cdisselkoen: why the `Box`? does it deserve a comment explaining why?

cedar-policy-generators/src/abac.rs
```diff
@@ -706,14 +716,14 @@ pub enum Type {
     /// Set, with the given element type. Note that we only generate
     /// homogeneous sets.
     /// Set(None) means any set type. It will still be a homogeneous set.
-    Set(Option<Box<Type>>),
+    Set(Box<Type>),
```
cdisselkoen: Should the comment about `Set(None)` be removed?

cedar-policy-generators/src/abac.rs
```diff
@@ -706,14 +716,14 @@ pub enum Type {
...
     /// Note that for now we have only a single Record type: all records are the
     /// same type, no effort to generate records with particular attributes
-    Record,
+    Record(BTreeMap<SmolStr, QualifiedType>),
```
cdisselkoen: Should this comment be removed?

cedar-policy-generators/src/abac.rs
```diff
@@ -706,14 +716,14 @@ pub enum Type {
...
     /// Note that for now we have only a single Entity type: all entities are
     /// the same type, no effort to generate entities with particular
     /// attributes, or distinguish entities of different entity types
-    Entity,
+    Entity(EntityType),
```
cdisselkoen: Should this comment be removed?

cedar-policy-generators/src/abac.rs
```diff
@@ -739,20 +749,15 @@ impl Type {
...
-    /// Set type, without specifying the element type. (It will still be a
-    /// homogeneous set.)
-    pub fn any_set() -> Self {
-        Type::Set(None)
+        Type::Set(Box::new(element_ty))
     }
     /// Record type
-    pub fn record() -> Self {
-        Type::Record
+    pub fn record(m: impl Iterator<Item = (SmolStr, QualifiedType)>) -> Self {
```
cdisselkoen: I suggest taking `IntoIterator` instead of `Iterator`.  This will allow callers to, e.g., pass an entire map without converting it into an iterator themselves

cedar-policy-generators/src/expr.rs
```diff
@@ -2580,24 +1749,12 @@ impl ExprGenerator<'_> {
 
 /// internal helper function, get a [`json_schema::Type`] representing a Record
 /// with (at least) one attribute of the specified name and type.
```
cdisselkoen: should this comment be updated? It returns `Type`

cedar-policy-generators/src/schema.rs
```diff
@@ -1288,54 +1294,11 @@ impl Schema {
...
-        }
-        .map(|ty| json_schema::Type::Type { ty, loc: None }))
-    }
-
     /// get an attribute name and its `json_schema::Type`, from the schema
-    pub fn arbitrary_attr(
-        &self,
-        u: &mut Unstructured<'_>,
-    ) -> Result<&(SmolStr, json_schema::Type<ast::InternalName>)> {
+    pub fn arbitrary_attr(&self, u: &mut Unstructured<'_>) -> Result<SmolStr> {
```
cdisselkoen: Should this comment be updated? The function doesn't return a `json_schema::Type`

cedar-policy-generators/src/expr.rs
```diff
@@ -616,7 +617,7 @@ impl ExprGenerator<'_> {
                         // has expression on an entity, for an arbitrary attribute name
                         1 => Ok(ast::Expr::has_attr(
                             self.generate_expr_for_type(
-                                &Type::entity(),
+                                &Type::Entity(u.choose(self.schema.entity_types())?.clone()),
```
john-h-kastner-aws, line 620: This pattern comes up enough to possibly be worth pulling into a method

shaobo-he-aws:
> Looks good. In your comment you say that this should allow a user of the generators to provide a custom Cedar schema to the generator.
> 
> But I see no code to do this. I can confirm trying to convert an existing `ValidatorSchema` in cedar to a generator schema is near impossible.

The functionality is actually implemented in #702. I'm currently testing if #702 works.

cedar-policy-generators/src/abac.rs
shaobo-he-aws: Changed to `Arc`.

cedar-policy-generators/src/expr.rs
shaobo-he-aws: Yes, #702 implements such a method.

cedar-policy-generators/src/abac.rs
cdisselkoen: I guess my question was, why not just `Type`?

shaobo-he-aws: For cheap clones?

cdisselkoen: I guess?  It comes with a performance penalty when constructing `QualifiedType` from scratch though, or accessing the `ty` field.  I don't actually care that much, was just curious.  Maybe we should benchmark how expensive a clone of `Type` is currently, and whether it's worth it to make `Type` itself cheaper to clone by wrapping some things in Arcs

shaobo-he-aws: Got it. I changed it to owned `Type`. Let's merge it and see how it goes.

PR #731, opened 2025-09-04
refactor Lean FFI and ensure drops

cdisselkoen:
Adds an `OwnedLeanObject` safe wrapper, which most importantly ensures that we decrement the Lean reference count for all Lean objects after we're done using them.

### DIFF
--- a/cedar-lean-ffi/src/lean_ffi.rs
+++ b/cedar-lean-ffi/src/lean_ffi.rs
@@ -109,30 +109,52 @@ static START: Once = Once::new();
 /// A struct which will initialize the lean backend (and initialize a thread running the lean runtime)
 pub struct CedarLeanFfi {}
 
-/// Takes ownership of its argument -- do not access `lean_str_obj` after
-/// calling this function
-fn lean_obj_p_to_rust_string(lean_str_obj: *mut lean_object) -> String {
-    let lean_obj_p = unsafe { lean_string_cstr(lean_str_obj) };
-    let lean_obj_cstr = unsafe { CStr::from_ptr(lean_obj_p as *const c_char) };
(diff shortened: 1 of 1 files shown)


john-h-kastner-aws (approved):
very nice

chaluli (approved):
Looks great!

shaobo-he-aws (approved):
Looks great

cedar-lean-ffi/src/lean_ffi.rs
```diff
@@ -109,30 +108,84 @@ static START: Once = Once::new();
...
-            y.add(i).write(*bi)
+/// Safe wrapper around `*mut lean_object`, but unlike `OwnedLeanObject`,
+/// this one assumes that we are _not_ responsible for freeing the object
+/// when the Rust value is dropped.
+struct BorrowedLeanObject(*mut lean_object);
+
+impl BorrowedLeanObject {
+    /// Create a `BorrowedLeanObject` with `buf` as its contents.
+    ///
+    /// As of this writing, it's unclear whether we (Rust) are responsible for
```
john-h-kastner-aws: I think the right way to think about it is that a function exported from Lean takes ownership and decrements the reference count itself. So, if we never use this object in an FFI call, we would need to decrement it.

IIUC, we could have this return an `OwnedLeanObject`, and then `call_lean_ffi_function` calls `std::mem::forget(arg)` to not decrement the reference count twice.

cdisselkoen: Makes sense. Will adjust

chaluli (approved):
Great stuff. Looks to be working. This seems like it was a bit tricky because of lack of documentation. Great job to you and John for figuring it out.

PR #732, opened 2025-09-04
Faster `Set` and `Map` construction in protobuf schema parsing

john-h-kastner-aws:
Looking at a flamegraph I saw `toSchema` was spending a lot of time calling `insertCanonical` to build well formed sets and maps. We can instead sort and de-duplicate lists using fast lean library functions.

On the particular slow unit also discussed in #726, this change reduces the time per FFI call from ~100ms to ~50ms. The next largest contributor to the run time looks like it's string decoding in protobuf parsing without any low-hanging fruit to optimize it.

I've not proven that these implementations are equivalent to `Set.make` and `Map.make`. We don't currently have any proofs about the output of protobuf parsing (and I don't think we particularly want to), but this would make any future attempt a bit more difficult.


cdisselkoen:
In some places in the Protobuf parsing code we use standard set/map structures internally (e.g., Lean standard-library structures), and then convert to `Data.Set` / `Data.Map` once, at the end.  Would that be a better solution here?  Would both solutions work together?  Is this better, and we should adapt all of the Protobuf parsing code to use this solution?

cedar-lean/CedarProto/Schema.lean
```diff
@@ -34,6 +34,27 @@ deriving instance Hashable for Spec.Name
 deriving instance Hashable for Spec.EntityType
 deriving instance Hashable for Spec.EntityUID
 
+/-
+Fast variant of `Set.make`, assuming we start with starting with an array we can
```
cdisselkoen:
```suggestion
Fast variant of `Set.make`, assuming we start with an array we can
```

cdisselkoen (approved):

PR #734, opened 2025-09-06
Do not use depend on anyhow

shaobo-he-aws:
anyhow's backtracing functionality does not work well with Lean FFIs

*Issue #, if available:*

*Description of changes:*


john-h-kastner-aws (approved):
How exactly does this changes error reports?

I guess I'll trust that is an improvement, but can't tell what it'll do form the diff.

cdisselkoen:
My understanding is it will not improve error reports, but also won't meaningfully regress them.  The reason for the change given in the description above is that anyhow's backtracing functionality does not work well with the Lean FFI

PR #735, opened 2025-09-08
Fix README of cedar-lean

shaobo-he-aws:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cedar-lean/README.md
+++ b/cedar-lean/README.md
@@ -29,12 +29,7 @@ To run the unit tests:
 lake exe CedarUnitTests
 ```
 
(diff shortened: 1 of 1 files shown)


cdisselkoen:
The old instructions referred to `authorize`, but the README we're linking to doesn't list an `authorize` subcommand.  We should probably add an `authorize` subcommand to the cedar-lean-cli, or if it already exists, update the README there

john-h-kastner-aws (approved):
There's still a `lake exe Cli` still and it does something (probably just takes protobuf now). Can we delete it entirely? Otherwise we should document it's new usage.

PR #736, opened 2025-09-08
Simplify PE by removing `absent` case on partial entities 

oflatt:
*Description of changes:*

This PR simplifies our codebase by removing the absent case on partial entities. It turns out that an absent entity is equivalent to a present one with no attributes, tags, or ancestors.

### DIFF
--- a/cedar-lean/Cedar/TPE/BatchedEvaluator.lean
+++ b/cedar-lean/Cedar/TPE/BatchedEvaluator.lean
@@ -51,7 +51,7 @@ def batchedEvalLoop
   | 0 => residual
   | n + 1 =>
     let toLoad := residual.allLiteralUIDs.filter (λ uid => (store.find? uid).isNone)
-    let newEntities := ((loader toLoad).mapOnValues EntityOrMissing.asPartial)
+    let newEntities := ((loader toLoad).mapOnValues MaybeEntityData.asPartial)
     let newStore := newEntities ++ store
 
     match Cedar.TPE.evaluate residual req.asPartialRequest newStore with
--- a/cedar-lean/Cedar/TPE/Input.lean
+++ b/cedar-lean/Cedar/TPE/Input.lean
@@ -205,7 +205,7 @@ This is because
 
 This is a necessary condition for the soundness of batched entity loading.
 -/
-def EntityOrMissing.asPartial :
+def MaybeEntityData.asPartial :
   MaybeEntityData → PartialEntityData
 | none =>
   { attrs :=  (.some Map.empty)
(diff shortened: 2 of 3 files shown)


cedar-lean/Cedar/TPE/BatchedEvaluator.lean
```diff
@@ -36,14 +36,18 @@ See `EntityLoader.WellBehaved` for a formal definition.
 -/
 abbrev EntityLoader := Set EntityUID → Map EntityUID (Option EntityData)
```
shaobo-he-aws: Nit: add an abbreviation of `Option EntityData` and add a `asPartialEntityData` to it.

cedar-lean/Cedar/TPE/BatchedEvaluator.lean
```diff
@@ -36,14 +36,18 @@ See `EntityLoader.WellBehaved` for a formal definition.
...
 
+
+/-- subtle: a missing entity is the same as a concrete entity with empty attrs, tags, and ancestors.
```
shaobo-he-aws: Nit: expand the comment a little bit and elaborate on why they are the same.

cedar-lean/Cedar/TPE/Input.lean
```diff
@@ -45,33 +45,29 @@ structure PartialRequest where
...
-inductive PartialEntityData where
-  | present (attrs : Option (Map Attr Value)) (ancestors : Option (Set EntityUID)) (tags : Option (Map Attr Value))
-  | absent
+structure PartialEntityData where
+  attrs     : Option (Map Attr Value)
+  ancestors : Option (Set EntityUID)
+  tags      : Option (Map Attr Value)
 
-abbrev PartialEntities := Map EntityUID PartialEntityData
+abbrev EntityOrMissing := Option EntityData
```
emina: I'd expect `EntityOrMissing` to be an abbreviation for `Option Entity` rather than `Option EntityData`.  How about `OptionalEntityData` or `MaybeEntityData` or something like that?  A separate comment can explain that if a `uid` is mapped to `none` by `EntitiesWithMissing`, this means the UID is not present in the underlying entities. 

I would also rename `EntitiesWithMissing` to `SlicedEntities` or something like that.  This makes it clear that we need this extra marker (whether something is present or not) because we are extracting definitive information (present / absent) from an underlying store.

oflatt: Oh right, I got confused because I have another PR coming that refactors it to be an `Entity` instead of `EntityData`.

I'll fix this PR and then resolve conflicts later

PR #737, opened 2025-09-10
[in progress] Convergence of Batched Evaluation to a Value

oflatt:
*Issue #, if available:*

*Description of changes:*


oflatt:
I'm guessing we should hold off on this PR until batched evaluation is more stable- we might extend it to treat ancestors specially

PR #742, opened 2025-09-16
Add DRT for batched evaluation

shaobo-he-aws:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cedar-drt/fuzz/fuzz_targets/batched-evaluation-drt.rs
+++ b/cedar-drt/fuzz/fuzz_targets/batched-evaluation-drt.rs
@@ -1,0 +1,149 @@
+/*
+ * Copyright Cedar Contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#![no_main]
+use cedar_drt::CedarLeanEngine;
+use cedar_drt_inner::{fuzz_target, schemas};
+
+use cedar_policy::{Decision, Effect, Entities, Policy, PolicySet, Schema, TestEntityLoader};
+
+use cedar_policy_generators::{
+    abac::{ABACPolicy, ABACRequest},
+    err::Error,
+    hierarchy::HierarchyGenerator,
+    schema,
+    settings::ABACSettings,
+};
+use libfuzzer_sys::arbitrary::{self, Arbitrary, Unstructured};
+
+/// Input expected by this fuzz target
+#[derive(Debug, Clone)]
+pub struct FuzzTargetInput {
+    /// generated schema
+    pub schema: schema::Schema,
+    /// generated entity slice
+    pub entities: Entities,
+    /// generated policy
+    pub policy: ABACPolicy,
+    /// the requests to try for this hierarchy and policy. We try 8 requests per
+    /// policy/hierarchy
+    pub requests: [ABACRequest; 8],
+    pub iterations: u8,
+}
+
+/// settings for this fuzz target
+const SETTINGS: ABACSettings = ABACSettings {
+    match_types: true,
+    enable_extensions: true,
+    max_depth: 7,
+    max_width: 7,
+    enable_additional_attributes: true,
+    enable_like: true,
+    enable_action_groups_and_attrs: true,
+    enable_arbitrary_func_call: true,
+    enable_unknowns: false,
+    enable_action_in_constraints: true,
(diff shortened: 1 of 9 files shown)


cedar-drt/fuzz/fuzz_targets/batched-evaluation-drt.rs
```diff
@@ -0,0 +1,148 @@
...
+        let mut loader = TestEntityLoader::new(&input.entities);
+        log::debug!("policy: {policyset}");
+
+        for req in input.requests {
+            let req = req.into();
+            log::debug!("req: {req}");
+            // We need to start with an `Ok` result of Rust because the
+            // validation DRT property is if Rust validations, then Lean also
+            // does. `is_authorized_batched` validates policies/request/entities
+            if let Ok(rust_decision) =
```
oflatt, line 128: Shouldn't the rust and lean agree completely? So if one fails the other does and vice versa?

cedar-drt/fuzz/fuzz_targets/batched-evaluation-pbt.rs
```diff
@@ -0,0 +1,141 @@
...
+        let policy = Policy::from(input.policy);
+        let mut policyset = PolicySet::new();
+        policyset.add(policy.clone()).unwrap();
+        let mut loader = TestEntityLoader::new(&input.entities);
+        log::debug!("policy: {policyset}");
+
+        for req in input.requests {
+            let req = req.into();
+            log::debug!("req: {req}");
+            if let Ok(batched_decision) =
```
oflatt: Similarly, I think batched evaluation should be completely equivalent to normal evaluation. So it errors iff normal evaluation errors.

cedar-drt/fuzz/fuzz_targets/batched-evaluation-pbt.rs
```diff
@@ -0,0 +1,141 @@
...
+    enable_arbitrary_func_call: true,
+    enable_unknowns: false,
+    enable_action_in_constraints: true,
+    per_action_request_env_limit: ABACSettings::default_per_action_request_env_limit(),
+    total_action_request_env_limit: ABACSettings::default_total_action_request_env_limit(),
+};
+
+impl<'a> Arbitrary<'a> for FuzzTargetInput {
+    fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
+        let schema = schema::Schema::arbitrary(SETTINGS.clone(), u)?;
```
oflatt, line 65: Should we do some that are type aware? Is most of this going to be taken up by errors, since batched evaluation requires type-correct policies?

cedar-lean/CedarFFI/Main.lean
```diff
@@ -800,4 +803,59 @@ private def ignoreOutput (vc : SymEnv → Cedar.SymCC.Result Cedar.SymCC.Asserts
     let data := (String.fromUTF8? inner_buffer.data).getD ""
     return ({ data := data, duration := r.duration } : Timed String)
 
+def EntityLoader.mk (entities: Cedar.Spec.Entities) : EntityLoader :=
```
oflatt: I think the lean codebase already has a test entity loader constructor that does the same thing.

cedar-drt/fuzz/fuzz_targets/batched-evaluation-drt.rs
shaobo-he-aws: We unfortunately can't because Lean and Rust have different precision in validation.

cedar-drt/fuzz/fuzz_targets/batched-evaluation-pbt.rs
shaobo-he-aws: I set maximum iteration to a random number. Maybe I should use the API to get the upper bound.

cedar-lean/CedarFFI/Main.lean
shaobo-he-aws: Yep. Will remove it.

cedar-drt/fuzz/fuzz_targets/batched-evaluation-pbt.rs
shaobo-he-aws: I believe the generator is type-directed according to the settings (controlled by `match_types`).

cedar-drt/fuzz/fuzz_targets/batched-evaluation-pbt.rs
john-h-kastner-aws: What happens if the number of iteration is too low? If that's an `Err`, can we match on the error to just ignore that one?

oflatt: I think getting the upper bound (using level validation) is a good idea. level validation doesn't support literals, so you can fall back to the size of the program

oflatt: The actual errors might differ though

shaobo-he-aws: > What happens if the number of iteration is too low? If that's an Err, can we match on the error to just ignore that one?

I don't think we can do it elegantly without #743. Lean does not throw such an error as we do not have a model for partial authorization.

shaobo-he-aws: Used a fixed upper bound.

PR #745, opened 2025-09-24
Adjust level based entity slicing to be infallible

john-h-kastner-aws:
This procedure would previously return `none` if it dereferenced an entity which did not exist in the original entities. This PR adjust it to instead return an entity slice without the missing entity.

This change means that evaluating a request with the slice may result in an unknown entity error, but the soundness proofs shows that that this only happens when the error also occurred using the unsliced entities.

The top level theorem drops the precondition `(hs : .some slice = entities.sliceAtLevel request n)` leaving us with

```
theorem validate_with_level_is_sound {ps : Policies} {schema : Schema} {n : Nat} {request : Request} {entities : Entities}
  (hwf : schema.validateWellFormed = .ok ())
  (hr : validateRequest schema request = .ok ())
  (he : validateEntities schema entities = .ok ())
  (htl : validateWithLevel ps schema n = .ok ()) :
  isAuthorized request entities ps = isAuthorized request (entities.sliceAtLevel request n) ps
```

### DIFF
--- a/cedar-lean/Cedar/Spec/Slice.lean
+++ b/cedar-lean/Cedar/Spec/Slice.lean
@@ -40,14 +40,14 @@ def Request.sliceEUIDs (r : Request) : Set EntityUID :=
   Set.make [r.principal, r.action, r.resource] ∪
   (Value.record r.context).sliceEUIDs
 
-def Entities.sliceAtLevel (es : Entities) (r : Request) (level : Nat) : Option Entities := do
-  let slice ← sliceAtLevel r.sliceEUIDs level
-  let slice ← slice.elts.mapM λ e => do some (e, ←(es.find? e))
-  some (Map.make slice)
+def Entities.sliceAtLevel (es : Entities) (r : Request) (level : Nat) : Entities :=
+  let slice := sliceAtLevel r.sliceEUIDs level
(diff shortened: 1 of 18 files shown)


cedar-lean/Cedar/Spec/Slice.lean
```diff
@@ -40,14 +40,14 @@ def Request.sliceEUIDs (r : Request) : Set EntityUID :=
...
-    | 0 => some ∅
-    | level + 1 => do
-      let eds ← work.elts.mapM es.find?
-      let slice ← List.mapUnion id <$> eds.mapM (λ ed => sliceAtLevel ed.sliceEUIDs level)
-      some (work ∪ slice)
+  sliceAtLevel (work : Set EntityUID) : Nat → Set EntityUID
+    | 0 => ∅
+    | level + 1 =>
+      let eds := work.elts.filterMap es.find?
+      let slice := List.mapUnion id $ eds.map (λ ed => sliceAtLevel ed.sliceEUIDs level)
```
emina, line 52: Nit: I think this looks nicer without `$` :)

let slice := List.mapUnion id (eds.map λ ed => sliceAtLevel ed.sliceEUIDs level)

emina (approved):

PR #747, opened 2025-09-26
Pass schema object as pointer to symcc targets

### LINKED ISSUE #741: Optimize lean FFI for symcc
### Category

Other

Our fuzzing has been regularly finding excessively slow inputs to various symcc differential testing targets. Some basic profiling show that the slow down isn't in symbolic compilation (either in Rust or in Lean), but in parsing the protobuf structures to serialize data for FFI function calls from Rust into Lean.

We've tried to resolve this two ways:
* Setting a low limit for the number of request environments for these target (#719)
* Optimizing some parts of protobuf parsing in Lean (#720, #732)

But, we're still finding these slow inputs.

The difference between the symcc targets and the other differential test targets is that the symcc targets make O(n) FFI calls into Lean (at least one call for every request environment) while other targets only make O(1) FFI calls per input. This means inefficiencies in the FFI layer will be much more noticeable.

To fully resolve this issue, we should find a way to avoid repeating work between different request environments for the symcc targets. We might do this by

* Parsing constant parts of the request once, storing a pointer to the Lean object for the parsed request to reuse between FFI calls for different request environments
* Moving more of the testing logic into Lean so that these targets can also make a single FFI call, with structures being saved and reused within the Lean implementation as appropriate.

If we attempt these solution and find them insufficient or not practical, we could instead apply a change like #726, setting a dramatically lower limit on the number of request environments for symcc targets (after considering the impact this might have on effectiveness of our differential testing). 

.

### Additional context

_No response_

john-h-kastner-aws:
Adds and exports a new Lean function `loadProtobufSchema` which parses a schema from protobuf and returns a lean object for that schema. This object is then passed to all the symcc lean ffi function.

This avoid re-parsing the schema for every request environment.

Fixes #741. For the one slow unit I tested this on, run time is reduced from 6928ms to 562ms. Currently running effected fuzz targets locally and I haven't found any issues

### DIFF
--- a/cedar-lean-cli/src/analysis.rs
+++ b/cedar-lean-cli/src/analysis.rs
@@ -457,10 +457,11 @@ fn policyset_vacuous(
     let mut vr = Vec::new();
 
     let lean_context = CedarLeanFfi::new();
+    let schema = lean_context.load_lean_schema_object(schema)?;
     for req_env in req_envs {
-        if lean_context.run_check_always_allows(policyset, schema, req_env)? {
+        if lean_context.run_check_always_allows(policyset, schema.clone(), req_env)? {
             vr.push(VacuityResult::MatchesAll);
-        } else if lean_context.run_check_always_denies(policyset, schema, req_env)? {
+        } else if lean_context.run_check_always_denies(policyset, schema.clone(), req_env)? {
             vr.push(VacuityResult::MatchesNone);
         } else {
             vr.push(VacuityResult::MatchesSome);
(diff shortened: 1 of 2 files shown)


chaluli (approved):

chaluli (approved):

john-h-kastner-aws:
Currently have 3 hours running each of the fuzz impacted fuzz targets, so I'm pretty confident this won't break anything in the weekend test runs

PR #751, opened 2025-09-29
Added additional SymCC PBT targets

katherine-hough:
*Issue #, if available:*
N/A 

*Description of changes:*
* Added fuzzing targets that check that SymCC verification tasks performed on inputs that pass strict validation either timeout or produce a value.
* Fixed potential resource leak from unreaped child processes in `symcc-cex-pbt`.


shaobo-he-aws (approved):

katherine-hough:
Currently, these targets also let us check that CVC5 does not return an unknown. I am not sure if we would get that if those options are used.

PR #756, opened 2025-09-30
Generate wrong number of args for extension function calls

john-h-kastner-aws:
Allows us to test that cedar-policy/cedar#1863 actually disables variadic isInRange calls by sometimes generating extra arguments to an extension function call. The Lean model has not been updated to handle variadic calls, so it will return an error as expected. Differential testing will assert that Rust also errors.

### DIFF
--- a/cedar-policy-generators/src/expr.rs
+++ b/cedar-policy-generators/src/expr.rs
@@ -1315,11 +1315,21 @@ impl ExprGenerator<'_> {
     ) -> Result<ast::Expr> {
         let func = self.ext_funcs.arbitrary_for_type(target_type, u)?;
         assert_eq!(&func.return_ty, target_type);
-        let args = func
+        let mut args: Vec<ast::Expr> = func
             .parameter_types
             .iter()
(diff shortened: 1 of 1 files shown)


cedar-policy-generators/src/expr.rs
```diff
@@ -1315,11 +1315,21 @@ impl ExprGenerator<'_> {
     ) -> Result<ast::Expr> {
         let func = self.ext_funcs.arbitrary_for_type(target_type, u)?;
         assert_eq!(&func.return_ty, target_type);
-        let args = func
+        let mut args: Vec<ast::Expr> = func
             .parameter_types
             .iter()
             .map(|param_ty| self.generate_expr_for_type(param_ty, max_depth, u))
             .collect::<Result<_>>()?;
+        if self.settings.enable_arbitrary_func_call && u.ratio::<u8>(1, 10)? {
```
katherine-hough: I think 1/10 might be too high of a rate for something that isn't expected to pass validation.

katherine-hough (approved):

PR #759, opened 2025-10-02
Ignore encoding errors about unsupported unicode

shaobo-he-aws:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cedar-drt/fuzz/src/symcc.rs
+++ b/cedar-drt/fuzz/src/symcc.rs
@@ -16,8 +16,9 @@
 
 use cedar_policy::{Policy, PolicySet, RequestEnv, Schema};
 use cedar_policy_symcc::{
-    solver::LocalSolver, Asserts, CedarSymCompiler, SymEnv, WellFormedAsserts, WellTypedPolicies,
-    WellTypedPolicy,
+    err::{EncodeError, Error},
+    solver::LocalSolver,
+    Asserts, CedarSymCompiler, SymEnv, WellFormedAsserts, WellTypedPolicies, WellTypedPolicy,
 };
 
 /// Compile a well-typed policy set to `Asserts`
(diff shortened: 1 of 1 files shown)


john-h-kastner-aws (approved):
Looks good.

Do you know how much work it would be to disable generating these strings in the first place? Feels like it'd be pretty invasive, so I agree with just ignoring these errors, but wondering if you though about this more than I did.

shaobo-he-aws:
> Looks good.
> 
> Do you know how much work it would be to disable generating these strings in the first place? Feels like it'd be pretty invasive, so I agree with just ignoring these errors, but wondering if you thought about this more than I did.

Not really. It seems changing the generators just for certain targets is non-trivial. We probably need to think more about it.

john-h-kastner-aws:
also, do we need to (or does this already) fix the counter example generation target? IIUC, that also calls cvc5, so if the same issue could exist.

shaobo-he-aws:
> also, do we need to (or does this already) fix the counter example generation target?

I believe that target should be fixed by [cedar-policy/cedar#](https://github.com/cedar-policy/cedar/pull/1868)

john-h-kastner-aws:
> > also, do we need to (or does this already) fix the counter example generation target?
> 
> I believe that target should be fixed by [cedar-policy/cedar#1868](https://github.com/cedar-policy/cedar/pull/1868)

I mean, does this issue also apply to that target? It feel likely to me that we'll see it fail for this reason tonight since you've merged that PR.

chaluli:
> > > also, do we need to (or does this already) fix the counter example generation target?
> > 
> > 
> > I believe that target should be fixed by [cedar-policy/cedar#1868](https://github.com/cedar-policy/cedar/pull/1868)
> 
> I mean, does this issue also apply to that target? It feel likely to me that we'll see it fail for this reason tonight since you've merged that PR.

I don't believe the CEX target has the same issue. This target errors if the policies validate and the encoder returns an error. But the CEX one AFAIK will just skip the checks if it fails to encode.

PR #763, opened 2025-10-07
Create only one solver per fuzzer worker process

shaobo-he-aws:
*Issue #, if available:*

*Description of changes:*


chaluli (approved):

chaluli (commented):
Do the other fuzz targets that create a solver already use a single solver? If not, shouldn't those also be updated?

PR #782, opened 2025-11-05
Fix clippy warnings in Rust code

shaobo-he-aws:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cedar-lean-ffi/src/lean_object.rs
+++ b/cedar-lean-ffi/src/lean_object.rs
@@ -108,13 +108,14 @@ pub struct LeanObject<'a>(
     pub(crate) PhantomData<&'a OwnedLeanObject>,
 );
 
-/// Convert a Lean object to an owned object by incrementing the reference count.
-impl From<LeanObject<'_>> for OwnedLeanObject {
-    fn from(value: LeanObject) -> Self {
+impl<'a> LeanObject<'a> {
+    /// Convert this borrowed Lean object to an owned object by incrementing the reference count.
+    #[allow(clippy::wrong_self_convention)]
+    pub fn to_owned(&self) -> OwnedLeanObject {
         unsafe {
-            lean_inc(value.0);
+            lean_inc(self.0);
         }
-        OwnedLeanObject(value.0)
+        OwnedLeanObject(self.0)
     }
 }
 
--- a/cedar-lean-ffi/src/lean_ffi.rs
+++ b/cedar-lean-ffi/src/lean_ffi.rs
@@ -952,7 +952,7 @@ impl CedarLeanFfi {
             )
         };
         match lean_schema_object.as_borrowed().as_result()? {
-            Ok(lean_ok_obj) => Ok(LeanSchema(lean_ok_obj.into())),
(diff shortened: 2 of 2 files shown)


cedar-drt/fuzz/fuzz_targets/symcc-cex-pbt.rs
```diff
@@ -65,8 +68,8 @@ static SOLVER: LazyLock<Mutex<Solver>> = LazyLock::new(|| {
     })
 });
 
-async fn get_solver() -> std::sync::MutexGuard<'static, Solver> {
-    let mut guard = SOLVER.lock().unwrap();
+async fn get_solver() -> MutexGuard<'static, Solver> {
```
john-h-kastner-aws, line 71: What's this change?

cedar-lean-ffi/src/lean_object.rs
```diff
@@ -108,15 +108,17 @@ pub struct LeanObject<'a>(
     pub(crate) PhantomData<&'a OwnedLeanObject>,
 );
 
-impl<'a> LeanObject<'a> {
-    /// Convert this borrowed Lean object to an owned object by incrementing the reference count.
-    pub fn to_owned(&self) -> OwnedLeanObject {
+/// Convert a Lean object to an owned object by incrementing the reference count.
+impl From<LeanObject<'_>> for OwnedLeanObject {
```
john-h-kastner-aws: Why this change? IMO converting from a borrowed to owned value shouldn't be by an `Into` impl.

cedar-drt/fuzz/fuzz_targets/symcc-cex-pbt.rs
cdisselkoen: Looks like it's switching from the `std` MutexGuard to the `tokio` one. Which is _absolutely_ what needs to happen in async code. Using the `std` MutexGuard in async code can cause deadlocks (probably doesn't here only because there's not much contention? idk)

cedar-lean-ffi/src/lean_object.rs
shaobo-he-aws: The object that calls `to_owned` is not a reference. So I just implemented the `From` trait of that type.

john-h-kastner-aws: `LeanObject` is conceptually always a reference because it doesn't own the underlying lean object.

E.g., `Arc<T>` doesn't implement `Into<T>`

shaobo-he-aws: I meant the usage or the receiver of the `to_owned` is a Rust reference.

john-h-kastner-aws: The point is to go from a `LeanObject` (or `&LeanObject`) to an `OwnedLeanObject` by incrementing the reference count, so I think `to_owned` is an appropriate name for this function. I don't think a `From` implementation should be incrementing the reference count.

shaobo-he-aws: Reverted.

PR #783, opened 2025-11-07
Improve debugging for policy set equivalence failures

cdisselkoen:
Pretty-prints the two policy sets, rather than (only) printing the Debug representations which are hard to read.

### DIFF
--- a/cedar-drt/src/parsing_utils.rs
+++ b/cedar-drt/src/parsing_utils.rs
@@ -137,12 +137,12 @@ pub fn check_policy_set_equivalence(set1: &PolicySet, set2: &PolicySet) {
         .map(PreservedTemplate::from)
         .collect::<HashSet<PreservedTemplate>>();
     // Check that the sets of preserved templates are equal
-    assert_eq!(t1, t2, "{}", similar_asserts::SimpleDiff::from_str(
-        &set1.to_string(),
-        &set2.to_string(),
-        "LHS",
-        "RHS",
-    ));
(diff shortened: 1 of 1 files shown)


cedar-drt/src/parsing_utils.rs
```diff
@@ -136,8 +136,8 @@ pub fn check_policy_set_equivalence(set1: &PolicySet, set2: &PolicySet) {
         .all_templates()
         .map(PreservedTemplate::from)
         .collect::<HashSet<PreservedTemplate>>();
-    // Check that the sets are the same size
-    assert_eq!(t1, t2);
+    // Check that the sets of preserved templates are equal
+    assert_eq!(t1, t2, "LHS was:\n{set1}\n\nRHS was:\n{set2}\n");
```
john-h-kastner-aws: use `similar_asserts::assert_eq` here to get a pretty diff of debug reprs also

cdisselkoen: I'd like to get the diff of `set1` vs `set2` though, which implement `Display`, as opposed to the diff of `t1` and `t2`, which are what we're comparing for equality, but don't implement `Display`

john-h-kastner-aws: in that case you could use `similar_asserts::SimpleDiff::from_str` to get pretty diffs for arbitrary strings

john-h-kastner-aws (approved):

PR #787, opened 2025-11-12
Reuse compiled policies

cdisselkoen:
Allows users of SymCC to reuse compiled policies for different queries. Previously, to compare whether Policy A implies any of 10 different policies, you would have to recompile Policy A 10 times, even if all the queries used the same environment. Now, you can compile Policy A once (per environment), and reuse that for an unlimited number of `implies` or `equivalent` or etc queries.

Rather than change the existing SymCC, this is implemented as a separate module `SymCCOpt` which provides methods like `impliesOpt?` and `checkImpliesOpt` analogous to SymCC's `implies?` and `checkImplies`.  Tests are refactored to check SymCC behavior both through the existing unoptimized interface and the new optimized interface.

This PR does not yet include any proofs, but we plan to add proofs that the two interfaces have equivalent semantics.

### DIFF
--- a/cedar-lean/SymTest/WellTyped.lean
+++ b/cedar-lean/SymTest/WellTyped.lean
@@ -179,16 +179,26 @@ private def policyB : Policy := {
 
 /-- Returns two `TestCase`s, one which tests unoptimized SymCC, the other which tests SymCCOpt -/
 def testFailsOnIllTyped (p : Policy) : List (TestCase SolverM) :=
-  let desc := s!"verifyNeverErrors of {p.id} fails due to type errors"
   [
-    test (desc ++ " (unoptimized)")
+    test s!"on the unoptimized path, verifyNeverErrors of {p.id} fails due to type errors"
       -- on the unoptimized path, verifyNeverErrors fails with .typeError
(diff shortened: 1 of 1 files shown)


cedar-lean/Cedar/SymCC.lean
```diff
@@ -87,31 +117,32 @@ containing a counterexample interpretation `I`. The function `vc` is expected to
 produce a list of terms type .bool that are well-formed with respect to `εnv`
 according to `Cedar.SymCC.Term.WellFormed`. This call resets the solver.
 -/
-def checkSat (vc : SymEnv → Result Asserts) (εnv : SymEnv) : SolverM (Option Interpretation) := do
+def checkSat (vc : SymEnv → Result Asserts) (εnv : SymEnv) : SolverM (Option Interpretation) :=
```
emina: It doesn't look like `checkSat` is used anywhere after this refactoring?  Can we get rid of it?

cdisselkoen: You're correct that we can get rid of it (unless we want to keep it around because of the parallel with `checkUnsat`, which is still used).

cdisselkoen: Removed now

PR #797, opened 2025-11-21
add proofs of the remaining 6 functions in SymCCOpt.lean

cdisselkoen:
Continues #793 / #796 by adding equivalence proofs for the two SymCC interfaces for the remaining 6 functions in SymCCOpt.lean.  These are all straightforward to prove in terms of lemmas we've already proven in #793 and #796.

### DIFF
--- a/cedar-lean/Cedar/Thm/SymCC/Opt.lean
+++ b/cedar-lean/Cedar/Thm/SymCC/Opt.lean
@@ -506,3 +506,153 @@ theorem disjoint?_eqv_disjointOpt?_ok {ps₁ ps₂ : Policies} {cps₁ cps₂ : 
   simp [compiled_policies_eq_wtps hcps₁ hwps₁, compiled_policies_eq_wtps hcps₂ hwps₂]
   simp [verifyDisjointOpt_eqv_verifyDisjoint_ok hcps₁ hcps₂ hwps₁ hwps₂] at h₁
   subst asserts ; rfl
+
+/--
+This theorem covers the "happy path" -- showing that if optimized policy
+compilation succeeds, then `wellTypedPolicies` succeeds and `checkNeverErrors` and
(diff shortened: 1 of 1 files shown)


cedar-lean/Cedar/Thm/SymCC/Opt.lean
```diff
@@ -506,3 +506,153 @@ theorem disjoint?_eqv_disjointOpt?_ok {ps₁ ps₂ : Policies} {cps₁ cps₂ :
...
+`checkNeverErrorsOpt?` are equivalent.
+-/
+theorem checkNeverErrors_eqv_checkNeverErrorsOpt_ok {p : Policy} {cp : CompiledPolicy} {Γ : Validation.TypeEnv} :
+  Γ.WellFormed →
+  CompiledPolicy.compile p Γ = .ok cp →
+  ∃ wp,
+    wellTypedPolicy p Γ = .ok wp ∧
+    checkNeverErrorsOpt cp = checkNeverErrors wp (SymEnv.ofTypeEnv Γ)
+:= by
+  simp [checkNeverErrors, checkNeverErrorsOpt]
```
shaobo-he-aws, line 522: Nit: these `simp` tactics can be merged.

shaobo-he-aws (approved):

cedar-lean/Cedar/Thm/SymCC/Opt.lean
chaluli: Note, it's not actually true that all consecutive simp tactics can be merged.

This is because the simp tactic is not monotonic (i.e., more inputs = more things that can be proved). The tactic can get stuck and fail if there are inputs that make it cycle between unfolding / folding, etc.

chaluli (approved):
Looks good to me.

PR #798, opened 2025-11-24
full equivalence proofs for the optimized SymCC interface

cdisselkoen:
Completes the proofs in #793 / #796 / #797 that the optimized SymCC interface is equivalent to the unoptimized one. Previous PRs only covered the happy path, when compilation does not error.  In this PR, we prove that the two interfaces are equivalent down to returning the same errors in the same conditions.


chaluli (approved):
Looks great to me. Nice job!

PR #799, opened 2025-11-25
Fix parsing of decimal literals

chaluli:
Fixes parsing of small negative decimal literals

### DIFF
--- a/cedar-lean/UnitTest/Decimal.lean
+++ b/cedar-lean/UnitTest/Decimal.lean
@@ -66,8 +66,25 @@ def testsForInvalidStrings :=
     testInvalid "-922337203685477.5809" "overflow"
   ]
 
-def tests := [testsForValidStrings, testsForInvalidStrings]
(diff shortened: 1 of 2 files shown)


cedar-lean/UnitTest/Decimal.lean
```diff
@@ -66,8 +66,25 @@ def testsForInvalidStrings :=
     testInvalid "-922337203685477.5809" "overflow"
   ]
 
-def tests := [testsForValidStrings, testsForInvalidStrings]
+private def testOp (str : String) (op: Cedar.Spec.Ext.Decimal → Cedar.Spec.Ext.Decimal → Bool) (lhs: String) (rhs: String) (expected: Bool) : TestCase IO :=
+  test str ⟨λ _ => checkEq parseAndApplyOp (.some expected)⟩
+where parseAndApplyOp :=
+  match parse lhs, parse rhs with
+    | .some lhs, .some rhs => Option.some (op lhs rhs)
+    | _, _ => .none
```
cdisselkoen, line 74: This could be written more concisely as 
```
where parseAndApplyOp := do some (op (← parse lhs) (← parse rhs))
```
depending on taste

PR #800, opened 2025-12-02
Add SymCC operations for alwaysMatches and neverMatches

cdisselkoen:
New operations that apply to a single policy and tell whether it alwaysMatches or neverMatches.  This differs from alwaysAllows/alwaysDenies in the case of `forbid` policies, and exposing these primitives directly allows reusing the same `CompiledPolicy` whereas the workaround implementing this operation in terms of alwaysAllows/alwaysDenies would require recompiling the policy.

This PR includes full soundness and completeness proofs for the operations.  Kiro helped substantially with the proofs (but there were a couple places it failed to finish the proofs on its own).

### DIFF
--- a/cedar-lean/Cedar/SymCC/Verifier.lean
+++ b/cedar-lean/Cedar/SymCC/Verifier.lean
@@ -76,7 +76,7 @@ If the asserts are satisfiable, then there is some input in `εnv` which `p`
 does match.
 -/
 def verifyNeverMatches (p : Policy) (εnv : SymEnv) : Result Asserts :=
-  -- always false _or_ error, i.e., just never equal to ⊙true
+  -- always evaluates to ⊙false _or_ error; i.e., never evaluates to ⊙true
   verifyEvaluate (λ t => not (eq t (⊙true))) p εnv
 
 /--


cedar-lean/Cedar/SymCC/Verifier.lean
```diff
@@ -61,6 +61,24 @@ which `p` errors.
...
+  -- never errors, _and_ is always true
+  verifyEvaluate (eq · (⊙true)) p εnv
+
+/--
+Returns asserts that are unsatisfiable iff `p` matches no inputs in `εnv`.
+If the asserts are satisfiable, then there is some input in `εnv` which `p`
+does match.
+-/
+def verifyNeverMatches (p : Policy) (εnv : SymEnv) : Result Asserts :=
+  -- always false _or_ error, i.e., just never equal to ⊙true
```
chaluli: Maybe a better comment is:

```
-- always evaluates to ⊙false _or_ error, i.e., never evaluates to ⊙true
```

The comment on first glance can be confusing because it might lead a reader to think "is syntactically false or an error".

chaluli (approved):
Looks great to me!

I think these in general are just better primitives to work with for analyses.

cedar-lean/Cedar/SymCC/Verifier.lean
cdisselkoen: Thanks, made this change.

PR #808, opened 2025-12-17
add new SymCC primitives for comparing the matches-behavior of individual policies

cdisselkoen:
It's useful to compare the matches-behavior of individual policies, particularly in the presence of `forbid` policies. Previously this would require recompiling the `forbid` policy as a `permit` policy.

Kiro wrote a fair amount of these proofs, looking at the existing proofs as guides. I had to make a number of manual tweaks to finish them off, but nonetheless Kiro accelerated the process greatly.

### DIFF
--- a/cedar-lean/SymTest.lean
+++ b/cedar-lean/SymTest.lean
@@ -14,5 +14,4 @@
  limitations under the License.
 -/
 
-import SymTest.Arith
 import SymTest.Main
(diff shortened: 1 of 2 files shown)


chaluli (approved):
Nice, good assist by Kiro! Also, good to have these as primitives.

chaluli (approved):

john-h-kastner-aws (approved):

PR #812, opened 2025-12-30
Optimized compile function computes footprint in one pass

cdisselkoen:
This new optimized `compile` function computes not only the `Term`, but also the `footprint`, in one pass.  This is a significant optimization compared to the status quo, where the unoptimized `footprint` function invokes `compile` on _every subexpression_, resulting in a huge amount of repeated work compiling.

This PR comes with full adjustments to the proofs, showing that the optimized path and unoptimized path remain equivalent. No changes were necessary to the toplevel theorem statements.

### DIFF
--- a/cedar-lean/Cedar/SymCCOpt/Compiler.lean
+++ b/cedar-lean/Cedar/SymCCOpt/Compiler.lean
@@ -138,7 +138,11 @@ def compileApp₂ (op₂ : BinaryOp) (arg₁ arg₂ : CompileResult) (εs : SymE
   -- the binaryOp term (if any) (particularly relevant for `getTag`)
   let argsFootprint := arg₁.footprint ∪ arg₂.footprint
   -- mimicking the behavior of the unoptimized compiler in how the direct
-  -- footprint for the binaryOp term itself is computed
+  -- footprint for the binaryOp term itself is computed.
+  -- Assuming everything is well-typed, this will be ∅ in all cases except `getTag` (because none of
+  -- the other binary operators can have entity type), but we still add this in every case because:
+  -- (1) we want the two compilers to agree even on non-well-typed inputs, for ease of proofs; and
+  -- (2) the unoptimized compiler adds this
   let binaryOpFootprint := λ term => directFootprint (ifSome arg₁.term (ifSome arg₂.term term))
   -- for the rest of this function, we consider only the `option.get`-ed args.
   -- see detailed note in the toplevel optimized `compile` function below.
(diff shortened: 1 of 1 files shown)


cedar-lean/Cedar/SymCCOpt/Compiler.lean
```diff
@@ -0,0 +1,399 @@
...
+        -- essentially inlining our `directFootprint` function, no need to
+        -- compute `.typeOf`, we know the term is option-entity type
+        let term := ⊙uid
+        .ok { term, footprint := Set.singleton term }
+      else
+        .error .typeError
+
+def compileVar (v : Var) (req : SymRequest) : Result CompileResult :=
+  match v with
+  | .principal =>
```
chaluli, line 95: Are principal, action, and resource not guaranteed to be entity typed?

cedar-lean/Cedar/SymCCOpt/Compiler.lean
```diff
@@ -0,0 +1,399 @@
...
+  | .resource  =>
+      if req.resource.typeOf.isEntityType then
+        -- essentially inlining our `directFootprint` function, no need to
+        -- compute `.typeOf`, we know the term is option-entity type
+        let term := ⊙req.resource
+        .ok { term, footprint := Set.singleton term }
+      else
+        .error .typeError
+  | .context   =>
+      if req.context.typeOf.isRecordType then
```
chaluli, line 120: similarly, is context not always record type?

chaluli (approved):

cedar-lean/Cedar/SymCCOpt/Compiler.lean
cdisselkoen: This guarantee exists in principle, but not in the Lean type system (`req.principal` is simply a `Term`). So we need to make this check to ensure that the optimized compiler agrees with the unoptimized compiler (which performs this check). For ease of proofs, we guarantee that the optimized and unoptimized compilers have the same output even on non-well-typed inputs, i.e., with minimal preconditions.

cedar-lean/Cedar/SymCCOpt/Compiler.lean
cdisselkoen: Similarly, the guarantee exists in principle, but not in the Lean type system -- `req.context` is simply a `Term`, and the unoptimized compiler performs this check.

cedar-lean/Cedar/SymCCOpt/Compiler.lean
cdisselkoen: In this particular case, the unoptimized compiler returns a type error if `req.principal` is a `Term` of the wrong type, and it's easiest for the proofs if the optimized compiler does the same, rather than having to argue that this case can never happen because we only apply `compile` to well-typed policies and requests.

cedar-lean/Cedar/SymCCOpt/Compiler.lean
chaluli: That's fair.

cedar-lean/Cedar/SymCCOpt/Compiler.lean
```diff
@@ -0,0 +1,399 @@
...
+  -- mimicking the behavior of the unoptimized compiler in how the direct
+  -- footprint for the binaryOp term itself is computed
+  let binaryOpFootprint := λ term => directFootprint (ifSome arg₁.term (ifSome arg₂.term term))
+  -- for the rest of this function, we consider only the `option.get`-ed args.
+  -- see detailed note in the toplevel optimized `compile` function below.
+  let t₁ := option.get arg₁.term
+  let t₂ := option.get arg₂.term
+  match op₂, t₁.typeOf, t₂.typeOf with
+  | .eq, ty₁, ty₂                           => let term := if (← reducibleEq ty₁ ty₂) then ⊙eq t₁ t₂ else ⊙false
+                                               .ok { term, footprint := argsFootprint ∪ binaryOpFootprint term } -- assuming everything is well-typed, the `binaryOpFootprint` term here will be ∅, but: (1) we want the two compilers to agree even on non-well-typed inputs, for ease of proofs; and (2) the unoptimized compiler adds this term
```
john-h-kastner-aws: perhaps not worth restating this in full every time.

cdisselkoen: I'll revise

PR #817, opened 2026-01-02
Move Lean corpus and integration tests to ffi crate

john-h-kastner-aws:
Makes more sense to store it there IMO

requires https://github.com/cedar-policy/cedar/pull/2077

### DIFF
--- a/.github/workflows/run_integration_tests_reusable.yml
+++ b/.github/workflows/run_integration_tests_reusable.yml
@@ -65,4 +65,4 @@ jobs:
       - name: Run integration tests
         working-directory: ./cedar-spec/cedar-lean-ffi
         shell: bash
-        run: source ~/.profile && source set_env_vars.sh && cargo test --features "integration-testing" -- --nocapture
+        run: source ~/.profile && source set_env_vars.sh && cargo test --features "integration-testing" -- --nocapture integration_tests corpus_tests


cedar-lean-ffi/Cargo.toml
```diff
@@ -8,12 +8,12 @@ publish = false
 cedar-policy = { version = "*", path = "../cedar/cedar-policy", features = ["protobufs"] }
 cedar-policy-core = { version = "*", path = "../cedar/cedar-policy-core" }
 cedar-policy-symcc = { version = "*", path = "../cedar/cedar-policy-symcc" }
+cedar-testing = { version = "*", path = "../cedar/cedar-testing"}
```
cdisselkoen, line 11: should this be an optional dep under the `integration-testing` feature?

cedar-lean-ffi/Cargo.toml
```diff
@@ -24,5 +24,9 @@ cargo_metadata = "0.20.0"
 
 [dev-dependencies]
 cool_asserts = "2.0.3"
+rstest = "0.26.1"
```
cdisselkoen, line 27: should this be an optional dependency under the `integration-testing` feature?

cedar-lean-ffi/Cargo.toml
```diff
@@ -24,5 +24,9 @@ cargo_metadata = "0.20.0"
...
 
 [profile.release]
+
+[features]
+integration-testing = []
```
cdisselkoen, line 32: do we need to adjust CI to run tests with this feature enabled?

john-h-kastner-aws: I think the change to run_integration_tests_reusable.yml takes care of this

cedar-lean-ffi/Cargo.toml
john-h-kastner-aws: I don't think you can have optional dev dependencies

cedar-lean-ffi/Cargo.toml
cdisselkoen: Yeah, on further examination, looks good

cedar-lean-ffi/Cargo.toml
john-h-kastner-aws: this could be, but I don't think it should because multiple `cedar-drt` fuzz target use the `impl CedarTestImplementation for CedarLeanFfi` as their main entry point

cdisselkoen: Cool

PR #818, opened 2026-01-05
more datastructure lemmas

john-h-kastner-aws:
*Issue #, if available:*

*Description of changes:*


cedar-lean/Cedar/Thm/Data/List/Lemmas.lean
```diff
@@ -1885,6 +1892,39 @@ theorem mem_of_map_implies_exists_unmapped
         assumption
       . rw [h₃]
 
+theorem mem_map_iff_find? [BEq β] [LawfulBEq β] {k : β} {f : α → β} {kvs : List α} :
```
cdisselkoen, line 1895: could we simplify this proof by using `mem_implies_find?` (immediately below) as a lemma?

john-h-kastner-aws: not that I can find

john-h-kastner-aws: lol kiro found a much simpler proof (using some other list lemmas)

PR #819, opened 2026-01-06
Evaluation Theorems

john-h-kastner-aws:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cedar-lean/Cedar/Thm/Authorization/PolicySlice.lean
+++ b/cedar-lean/Cedar/Thm/Authorization/PolicySlice.lean
@@ -35,6 +35,51 @@ def IsSoundPolicySlice (req : Request) (entities : Entities) (slice policies : P
     policy ∉ slice →
     ¬ satisfied policy req entities ∧ ¬ hasError policy req entities
 
+theorem sound_slice_transitive :
+  IsSoundPolicySlice r es slice₁ ps →
+  IsSoundPolicySlice r es slice₂ slice₁ →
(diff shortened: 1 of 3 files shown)


cedar-lean/Cedar/Thm/Authorization/PolicySlice.lean
```diff
@@ -35,6 +35,51 @@ def IsSoundPolicySlice (req : Request) (entities : Entities) (slice policies : P
...
+does evaluate to `false`.
+-/
+def IsSoundPolicySliceFalseDef (req : Request) (entities : Entities) (slice policies : Policies) : Prop :=
+  slice ⊆ policies ∧
+  ∀ policy ∈ policies,
+    policy ∉ slice →
+    (evaluate policy.toExpr req entities) = .ok false
+
+theorem is_sound_policy_slice_def_equiv {req : Request} {entities : Entities} {slice policies : Policies} :
+  IsSoundPolicySlice req entities slice policies ↔ IsSoundPolicySliceFalseDef req entities slice policies
```
lianah, line 65: Out of curiosity, why do you still need the previous definition? It seems more convoluted. Is it easier for some proofs?

john-h-kastner-aws: Looking at the other proofs, I think the original definition is easier for the proofs about the authorization algorithm since that's written in terms of `satisifed` and `hasError`.  The alternate definition is less convoluted when proving slicing soundness, so I think it makes sense to have both

PR #820, opened 2026-01-08
Refactor TypedExpr WellTyped soundness proof to use Residual proof

john-h-kastner-aws:
Also depends on `conversion_preserves_evaluation`, `conversion_preserves_typedness`, and `conversion_preserves_type_of`, so i did some other reorganization to make it easier to avoid some cycles in imports I was running into.


cdisselkoen (approved):
Removes a huge amount of redundant proof code, looks great

PR #827, opened 2026-01-14
Models and proofs of correctness for permissions queries algorithms

john-h-kastner-aws:
A permissions query algorithm is an procedure for enumerating resources with a specific type that a principal is authorized to access.

For each algorithm, this PR proves that a resources is listed if and only if it has the expected type and access is in fact allowed.

Specifically, the two main theorems are

```lean
theorem resource_scan_correct (n : Nat) (schema : Schema) (req : ResourcesForPrincipalRequest) (entities : Entities) (policies : Policies) (r : EntityUID) :
    validateWithLevel policies schema n = .ok () →
    schema.validateWellFormed = .ok () →
    validateRequest schema (req.req r) = .ok () →
    validateEntities schema entities = .ok () →
    (r ∈ scanResourcesForPrincipal n schema req entities policies ↔
    (r ∈ entities ∧ r.ty = req.resourceType ∧ (isAuthorized (req.req r) entities policies).decision = .allow))

theorem discretionary_resource_for_principal_correct :
  arePoliciesDiscretionary ps = true →
    (r ∈ discretionaryResourcesForPrincipal pq es ps ↔ (r.ty = pq.resourceType ∧ (isAuthorized (pq.req r) es ps).decision = .allow))
```

### DIFF
--- a/cedar-lean/Cedar/PQ/Discretionary.lean
+++ b/cedar-lean/Cedar/PQ/Discretionary.lean
@@ -1,0 +1,106 @@
+/-
+ Copyright Cedar Contributors
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+      https://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-/
+
+import Cedar.PQ.ResourcesForPrincipal
+import Cedar.Validation
+
+/-!
+This file defines a permission query algorithm for "discretionary" Cedar policies.
+Discretionary policies are those with specific constraints on principal, action,
+and resource scopes, with no additional conditions.
+-/
+
+namespace Cedar.PQ
+
+open Cedar.Data
+open Cedar.Spec
+open Cedar.Validation
+
+/--
+Determines if a policy is discretionary. A policy is discretionary if:
+- Principal scope is eq, mem, or isMem
+- Action scope is eq
+- Resource scope is eq
+- No `when` or `unless` conditions
+-/
+def isPolicyDiscretionary (p : Policy) : Bool :=
+  (match p.principalScope.scope with
+  | .eq _ | .mem _ | .isMem _ _ => true
+  | _ => false) &&
+  (match p.actionScope with
+  | .actionScope (.eq _) => true
+  | _ => false) &&
+  (match p.resourceScope.scope with
+  | .eq _ => true
+  | _ => false) &&
+  p.condition == []
+
+/-- Checks if all policies in a list are discretionary. -/
+def arePoliciesDiscretionary (policies : Policies) : Bool :=
+  policies.all isPolicyDiscretionary
+
+/--
+Filters policies to those that apply to a specific principal and resource type.
+Returns policies where:
+- The resource scope matches the given resource type
+- The principal scope includes the given principal (directly or through groups)
+-/
+def policiesForPrincipalAndResourceType
+  (ps : Policies)
+  (principal : EntityUID)
+  (principal_groups : Set EntityUID)
+  (rty : EntityType) :
+  Policies
+:= ps.filter λ p =>
+  (match p.resourceScope.scope with
+  | .eq e => e.ty == rty
+  | _ => false) &&
+  (match p.principalScope.scope.bound with
+  | .some b => b == principal || principal_groups.contains b
+  | .none => false)
+
+/--
+Creates a map from resources to the policies that apply to them.
+Only considers policies with concrete resource scopes (eq).
+-/
+def policiesByResource (ps : Policies) : Map EntityUID Policies :=
+  match ps with
+  | [] => Map.empty
+  | p :: ps' =>
+    let m := policiesByResource ps'
+    match p.resourceScope.scope with
+    | .eq resource =>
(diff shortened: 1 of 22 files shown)


cedar-lean/Cedar/PQ/Discretionary.lean
```diff
@@ -0,0 +1,106 @@
...
+def arePoliciesDiscretionary (policies : Policies) : Bool :=
+  policies.all isPolicyDiscretionary
+
+/--
+Filters policies to those that apply to a specific principal and resource type.
+Returns policies where:
+- The resource scope matches the given resource type
+- The principal scope includes the given principal (directly or through groups)
+-/
+def policiesForPrincipalAndResourceType
```
cdisselkoen: This implementation is incorrect for arbitrary policies (e.g., misses policies with `.is` in the resourceScope, or `.any` in the principalScope), so it might be prudent to add a comment that this is only intended for discretionary policies. or even change the name

cedar-lean/Cedar/PQ/Discretionary.lean
```diff
@@ -0,0 +1,106 @@
...
+  | p :: ps' =>
+    let m := policiesByResource ps'
+    match p.resourceScope.scope with
+    | .eq resource =>
+      let existing := m.find? resource |>.getD []
+      m.filter (λ r _ => r != resource) ++ Map.make [(resource, p :: existing)]
+    | _ => m
+
+/--
+Computes the set of resources that a principal can access for discretionary policies.
```
cdisselkoen: what does this precisely mean? Are `ps` required to be only discretionary policies? if `ps` contain some non-discretionary policies, will the non-discretionary policies be ignored? or will the results possibly be incorrect?

cedar-lean/Cedar/PQ/ResourceScan.lean
```diff
@@ -0,0 +1,54 @@
...
+ limitations under the License.
+-/
+
+import Cedar.PQ.ResourcesForPrincipal
+import Cedar.PQ.ResourceScan.PolicySlice
+import Cedar.PQ.ResourceScan.ResourceCandidates
+
+/-!
+This file defines a permission query algorithm for Cedar policy by iterating
+over the possible resource.
```
cdisselkoen:
```suggestion
This file defines a permission query algorithm for Cedar policies by iterating
over the possible resources.
```

cedar-lean/Cedar/PQ/ResourceScan.lean
```diff
@@ -0,0 +1,54 @@
...
+
+The algorithm includes three optimizations:
+1. Policy slicing is used to evaluate only relevant policies given the principal
+   and resource type.
+2. Resources candidates are selected based on what could possibly be authorized
```
cdisselkoen:
```suggestion
2. Resource candidates are selected based on what could possibly be authorized
```

cedar-lean/Cedar/PQ/ResourceScan.lean
```diff
@@ -0,0 +1,54 @@
...
+   given the resources in the policy scopes. This optimization only applies when
+   all policy constrain the resource. Otherwise, it fall back to checking every
```
cdisselkoen, line 30:
```suggestion
   all policies constrain the resource. Otherwise, it falls back to checking every
```

cedar-lean/Cedar/PQ/ResourceScan/PolicySlice.lean
```diff
@@ -0,0 +1,59 @@
...
+
+def isPrincipalPolicyUnqualifiedResource (principal : EntityUID) (p : Policy) : Bool :=
+  p.principalScope.scope.bound == some principal &&
+  p.resourceScope.scope.bound == none
+
+def isUnqualifiedPrincipalResourceTypePolicy (resourceType : EntityType) (p : Policy) : Bool :=
+  p.principalScope.scope.bound == none &&
+  match p.resourceScope.scope.bound with
+  | some e => e.ty == resourceType
+  | none => true
```
cdisselkoen, line 38: these three lines, which check whether the policy could possibly apply to `resourceType`, could

- be factored out into a separate `def` since they're repeated three times in this file
- be further optimized to consider policies with `resource is` in the scope (my understanding is these have `.bound none`, but we could exclude some of them)

cedar-lean/Cedar/PQ/ResourceScan/ResourceCandidates.lean
```diff
@@ -0,0 +1,74 @@
...
+import Cedar.PQ.ResourcesForPrincipal
+
+namespace Cedar.PQ
+
+open Cedar.Data
+open Cedar.Spec
+open Cedar.Validation
+
+/--
+  If this is an RBAC policy, i.e., it constraints the resource to be `in` or
```
cdisselkoen:
```suggestion
  If this is an RBAC policy, i.e., it constrains the resource to be `in` or
```

cedar-lean/Cedar/PQ/ResourceScan/ResourceCandidates.lean
```diff
@@ -0,0 +1,74 @@
...
+  `==` to a concrete entity in the policy scope, return a list of entities which
+  it could apply to. It may or may not actually apply to those resources
+  depending on the policy condition and the exact authorization request made.
+  Returns `None` for non-rbac policies, indicating that all resources might be
+  authorized.
```
cdisselkoen: I don't think this is our typical definition of RBAC

cdisselkoen (approved):

cedar-lean/Cedar/PQ/ResourceScan/PolicySlice.lean
john-h-kastner-aws: I think the current code is the clearest way to implement the intended algorithm

PR #828, opened 2026-01-20
Update TPE proof for `<error-free> && false` reduction

john-h-kastner-aws:
Model and proof for simplifications in https://github.com/cedar-policy/cedar/pull/2091

Update to TPE algorithm is in cedar-lean/Cedar/TPE/Evaluator.lean and cedar-lean/Cedar/TPE/Residual.lean.

Main lemma is

```lean
theorem error_free_evaluate_ok {r : Residual} :
  InstanceOfWellFormedEnvironment req es env →
  Residual.WellTyped env r →
  r.ErrorFree →
  (r.evaluate req es).isOk
```

showing that the error-free predicate guarantees a well-typed residual cannot error.

### DIFF
--- a/cedar-lean/Cedar/Thm/TPE/Soundness.lean
+++ b/cedar-lean/Cedar/Thm/TPE/Soundness.lean
@@ -226,7 +226,7 @@ theorem partial_evaluate_is_sound_and
 {pes : PartialEntities}
 {env : TypeEnv}
 (h₂ : InstanceOfWellFormedEnvironment req es env)
-(h₄ : RequestAndEntitiesRefine req es preq pes)
+(h₃ : RequestAndEntitiesRefine req es preq pes)
 (hᵢ₁ : Residual.WellTyped env x₁)
 (hᵢ₂ : Residual.WellTyped env x₂)
 (hᵢ₃ : x₁.typeOf = CedarType.bool BoolType.anyBool)
@@ -309,7 +309,7 @@ theorem partial_evaluate_is_sound_and
         subst h₇
         rw [Residual.error_free_spec] at h₆
         have h₇ : Residual.WellTyped env (TPE.evaluate x₁ preq pes) :=
-          partial_eval_preserves_well_typed h₂ h₄ hᵢ₁
+          partial_eval_preserves_well_typed h₂ h₃ hᵢ₁
         have h₈ := error_free_evaluate_ok h₂ h₇ h₆
         simp [Except.isOk, Except.toBool] at h₈
         split at h₈ <;> try contradiction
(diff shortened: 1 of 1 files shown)


cedar-lean/Cedar/Thm/TPE/PreservesTypeOf.lean
```diff
@@ -95,33 +97,41 @@ private theorem partial_eval_preserves_typeof_or {env : TypeEnv} {a b : Residual
...
+    all_goals
+      split at heq
+    any_goals
       split at heq
 
-    any_goals contradiction
+    any_goals
+      contradiction
 
     any_goals
```
cdisselkoen: nit: instead of using `any_goals` so much, we could be more explicit which goal we're intending to apply this to, by using `case h_whatever => ...` or `case' h_whatever => ...`

cedar-lean/Cedar/Thm/TPE/Soundness.lean
```diff
@@ -224,6 +226,7 @@ theorem partial_evaluate_is_sound_and
 {pes : PartialEntities}
 {env : TypeEnv}
 (h₂ : InstanceOfWellFormedEnvironment req es env)
+(h₄ : RequestAndEntitiesRefine req es preq pes)
```
cdisselkoen: nit: skipped h3

cedar-lean/Cedar/Thm/TPE/PreservesTypeOf.lean
john-h-kastner-aws: 91 goals unfortunately. I'll see if there's a more substantial simplification to this proofs

john-h-kastner-aws: yeah, this whole file could be simplified a lot

cedar-lean/Cedar/Thm/TPE/PreservesTypeOf.lean
```diff
@@ -46,45 +46,28 @@ private theorem partial_eval_preserves_typeof_and {env : TypeEnv} {a b : Residua
...
+    replace ih_b := ih_b h_wf h_ref h_bwt
+    rw [h_bty] at ih_b
+    exact ih_b
+  · simp only [Residual.typeOf]
+    cases h_wt
+    rfl
+  · simp [Residual.typeOf]
+  · cases h_wt
+    rename_i h_awt h_aty _ _
+    replace ih_a := ih_a h_wf h_ref h_awt
```
cdisselkoen, line 62: just in case you aren't aware of this tactic: this could be written as as a call to `specialize`. Same with some other `replace` calls in this file

PR #831, opened 2026-01-21
use the optimized SymCC interface in FFI calls

cdisselkoen:
Should be a significant performance improvement for DRT targets, if the symbolic compilation was the bottleneck (and not the protobuf stuff, the actual solver calls, etc).

### DIFF
--- a/cedar-lean/CedarFFI/Main.lean
+++ b/cedar-lean/CedarFFI/Main.lean
@@ -600,7 +600,8 @@ private def ignoreOutput (asserts : Cedar.SymCC.Asserts) (εnv : SymEnv) : Solve
 @[export assertsOfCheckNeverErrors] unsafe def assertsOfCheckNeverErrors (schema : Cedar.Validation.Schema) (req: ByteArray) : String :=
   runFfiM do
     let (_, cp) ← parseCheckPolicyReq schema req
-    runAndTime (λ () => verifyNeverErrorsOpt cp)
+    -- for backwards compatibility, we provide a λ that returns Result, even though verify*Opt functions return Asserts directly
+    runAndTime (λ () => (.ok $ verifyNeverErrorsOpt cp : Cedar.SymCC.Result Cedar.SymCC.Asserts))
 
 /--
   `req`: binary protobuf for an `CheckPolicyRequest`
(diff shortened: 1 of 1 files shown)


cedar-lean/CedarFFI/Main.lean
```diff
@@ -570,18 +599,22 @@ private def ignoreOutput (vc : SymEnv → Cedar.SymCC.Result Cedar.SymCC.Asserts
...
+    runAndTime (λ () => verifyNeverErrorsOpt cp)
 
 /--
   `req`: binary protobuf for an `CheckPolicyRequest`
 
   returns JSON encoded of the term generated by `verifyNeverErrors` on the deserialized policy
+
+  this DOES NOT use the optimized interface (which requires you to compile the
+  typechecked policy, not the original deserialized policy), so it is
+  inefficient. Callers should migrate to `assertsOfCheckNeverErrors` if/when possible
```
john-h-kastner-aws, line 613: Are there callers using this?

cdisselkoen: One caller in cedar-spec, [here](https://github.com/cedar-policy/cedar-spec/blob/2e9aa6b3e7c631ed1ef2f9090d9f44cfa7d62707/cedar-drt/fuzz/fuzz_targets/symcc-term-drt.rs#L101).

cdisselkoen: Actually that's not for this specific function, it's for the AlwaysAllows primitive, but we might/should expand that DRT target to cover all primitives

cdisselkoen: I'm not sure why that caller needs the OnOriginal variant -- might look into that after this PR

john-h-kastner-aws:
> I am confused how these changes are causing the deserialization error we're seeing in CI. I can't see how anything about the structure of the output from Lean should be changing with this PR

I haven't though through why it might or might not matter, but I see that the `*Opt` functions return `Asserts` instead of `Result Asserts`. That could change the serialization.

cedar-lean/CedarFFI/Main.lean
lianah: That seems to be the target that checks that the terms generated by Lean and Rust are equivalent. The following comment would suggest that minor discrepancies between the validator in Rust/Lean cause the enforcer to generate terms that don't match:
```
               // We use `asserts_of_check_always_allows_on_original` instead
                // of `asserts_of_check_always_allows` because there are
                // discrepancies between the policies generated by Lean and Rust
                // validators. And the enforcer is sensitive to such
                // discrepancies.
```

PR #835, opened 2026-01-23
Expand CedarLean SymCC FFI to cover all SymCC primitives equally

cdisselkoen:
This PR expands the CedarLean FFI to cover all SymCC primitives equally -- notably including the same functions for the more-recently-introduced primitives (alwaysMatches, matchesEquivalent, etc) as exist for the original primitives.

This PR also includes a refactor in this file to improve code duplication and conciseness, now that we have 11 or 12 (depending on how you count) SymCC primitives. The refactor is potentially separable -- if folks find the refactor controversial we can take just the first Git commit from this PR (which adds the new primitives "the long way", without the refactor) and review the refactor separately.

The other commit in this PR (again potentially separable) adjust the `open`s in this file. The existing `open`s cause name collisions for lots of `Spec`/`SymCC` and `Proto` types, resulting in the need to qualify lots of types at their use sites. I've adjusted this to privilege the Spec/SymCC ones and not `open` any `Proto` namespaces, with the result that `Proto` types need to be qualified at their use sites, but `Spec`/`SymCC` types can be used unqualified.


cedar-lean-ffi/src/lean_ffi.rs
```diff
@@ -13,6 +13,12 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
+#![expect(
+    dead_code,
+    reason = "declares a broad array of FFI functions, some of which are not currently used in any fuzz targets but are declared for completeness"
```
lianah, line 19: Is there a plan to add these to the fuzz targets?

cedar-lean-ffi/src/lean_ffi.rs
```diff
@@ -640,6 +829,54 @@ impl CedarLeanFfi {
...
+    comparePolicies_func!(
+        asserts_of_check_matches_equivalent_timed,
+        asserts_of_check_matches_equivalent,
+        assertsOfCheckMatchesEquivalent,
+        ResultDef::to_result,
+        Result<Vec<Term>, String>
+    );
+
+    comparePolicies_func!(
+        asserts_of_check_matches_equivalent_on_original_timed,
```
lianah, line 841: Do we still need the `on_original` functions now that the validator gives the same output?

cedar-lean-ffi/src/lean_ffi.rs
cdisselkoen: No. For instance, the fuzz target `symcc-smt-script-drt` uses Rust to generate `Term`s for an AlwaysAllows job, then checks that Rust and Lean produce the same SMTLIB script given those `Term`s. This involves serializing/deserializing the `Term`s themselves, rather than the policies, to send them from Rust to Lean. The point of this target is to differentially test the Rust and Lean Encoders, and as long as AlwaysAllows jobs are capable of producing basically all the interesting combinations of `Term`s, we don’t need to use any other primitives. This target uses the FFI function `smtlib_of_check_asserts` (only), which maps `Term`s -> SMTLIB scripts. For completeness we also have FFI functions for `smtlib_of_*` for all of the other SymCC primitives, but they are currently unused and we have no plans to use them.

cdisselkoen: to clarify: the other `smtlib_of_*` FFI functions perform both symbolic compilation and encoding. we don't currently have any fuzz targets that difftest both symbolic compilation and encoding -- we difftest those separately.

cedar-lean-ffi/src/lean_ffi.rs
cdisselkoen: I have a followup PR to remove those, but the plan was to leave them for a few days and make sure no concerning DRT failures. If you would rather I squash the followup PR into this one, I'm happy to do that.

lianah: No, that's fine. I think this PR has enough functionality as is :)

cdisselkoen: Followup is https://github.com/cedar-policy/cedar-spec/pull/843

PR #836, opened 2026-01-23
Add model of partial-auth and re-auth for TPE

john-h-kastner-aws:
Prove correctness of re-authorization. Note that reauthorization is defined (as it is in Rust) as always concretely evaluating all policy residuals without short-circuiting based on any concrete permits/forbids after partial evaluation.

This keeps the proofs simple, but it does not prove correctness of the decision returned by partial isAuthorized. We still need to prove that if the partial isAuthorized function returns a decision, then that decision is correct.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cedar-lean/Cedar/Thm/TPE.lean
+++ b/cedar-lean/Cedar/Thm/TPE.lean
@@ -31,7 +31,8 @@ import Cedar.Thm.Authorization
 /-!
 This file defines the main theorem of TPE soundness for authorization.
 
-Significant lemmas proving soundness of policy and expression evaluation are in `Cedar.Thm.TPE.Soundness`.
+Significant lemmas proving soundness of policy and expression evaluation are in
+`Cedar.Thm.TPE.Policy` and `Cedar.Thm.TPE.Soundness`.
 -/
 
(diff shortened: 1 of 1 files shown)


cedar-lean/Cedar/Thm/TPE.lean
```diff
@@ -169,4 +170,95 @@ theorem partial_evaluate_policy_is_sound
...
+  simp only [TPE.isAuthorized] at h_auth
+  cases h_auth₁ : List.mapM (fun p => ResidualPolicy.mk p.id p.effect <$> evaluatePolicy schema p preq pes) policies <;> simp [h_auth₁] at h_auth
+  rename_i residuals
+  subst h_auth
+
+  rw [List.mapM_ok_iff_forall₂] at h_auth₁
+  have h_auth₂ := List.forall₂_implies_all_left h_auth₁
+  replace h_auth₁ := List.forall₂_implies_all_right h_auth₁
+
+  have equiv_satisfied : ∀ effect,
```
cdisselkoen: worth pulling out as its own lemma?

lianah (approved):

PR #842, opened 2026-01-27
change to edition 2024 in most crates, and re- cargo fmt

cdisselkoen:
Changing to edition 2024 is a nonbreaking change (2021 crates can still depend on us, and we can still depend on 2021 crates), but it does apparently change how `cargo fmt` formats some code. This PR moves us to edition 2024 in all Rust crates in this repo, except `cedar-policy-generators`, which uses `gen` as a module name (which became a reserved keyword in edition 2024).

### DIFF
--- a/cedar-lean-ffi/src/lean_ffi.rs
+++ b/cedar-lean-ffi/src/lean_ffi.rs
@@ -979,7 +979,7 @@ mod test {
     #[link(name = "CedarProto", kind = "static")]
     #[link(name = "Batteries", kind = "static")]
     #[link(name = "CedarFFI", kind = "static")]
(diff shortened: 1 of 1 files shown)


cedar-drt/tests/benchmark.rs
```diff
@@ -164,10 +164,10 @@ fn print_summary(auth_times: HashMap<&str, Vec<f64>>, val_times: HashMap<&str, V
 
 #[test]
 #[ignore] // doesn't test anything, just prints timing results -- run this on
```
john-h-kastner-aws: nit: might add a new line before the comment after the ignore.

PR #844, opened 2026-01-28
Add more SymCC term-drt targets

cdisselkoen:
Replaces the single `symcc-term-drt` target (which only tested AlwaysAllows), with a set of 11 targets, one per SymCC primitive. For the primitives that take policyset inputs (as opposed to single-policy inputs), also adds multipolicy variants: the standard targets (like `symcc-term-drt-always-allows`) use policysets containing only a single policy, following the example of the Cedar DRT, while the multipolicy targets (like `symcc-term-drt-always-allows-multipolicy`) use policysets containing an arbitrary number of policies. The latter is a more thorough test, in particular testing the policy-combining logic in SymCC, but because generating multiple policies is slower and potentially more fallible, we wanted to also have singleton-policyset versions of these targets to align with our current practice elsewhere.

### DIFF
--- a/cedar-drt/fuzz/fuzz_targets/symcc-cex-pbt.rs
+++ b/cedar-drt/fuzz/fuzz_targets/symcc-cex-pbt.rs
@@ -20,18 +20,16 @@ use cedar_drt_inner::{
     fuzz_target,
     symcc::{RUNTIME, SinglePolicyFuzzTargetInput, local_solver},
 };
-
 use cedar_policy::{Authorizer, Decision, PolicySet};
-use cedar_policy_symcc::{
-    CedarSymCompiler, CompiledPolicySet, Env, WellFormedAsserts, always_allows_asserts,
-    always_denies_asserts, err::SolverError, solver::LocalSolver,
-};
-
-use std::sync::LazyLock;
+use cedar_policy_symcc::{CedarSymCompiler, CompiledPolicySet, Env, solver::LocalSolver};
+use std::{future::Future, sync::LazyLock};
 use tokio::{
     sync::{Mutex, MutexGuard},
-    time::{Duration, error::Elapsed, timeout},
+    time::{Duration, timeout},
 };
+
+/// Solver timeout used for this target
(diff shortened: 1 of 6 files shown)


cedar-drt/fuzz/fuzz_targets/symcc-term-drt-always-allows.rs
```diff
@@ -0,0 +1,67 @@
...
+    let policy: Policy = input.policy.into();
+    debug!("Schema: {}\n", input.schema.schemafile_string());
+    debug!("Policies: {policy}\n");
+    policyset.add(policy).unwrap();
+
+    if let Ok(schema) = Schema::try_from(input.schema) {
+        let lean_schema = lean_ffi.load_lean_schema_object(&schema).unwrap();
+        for req_env in schema.request_envs() {
+            // The validator DRT property we've been testing is that
+            // rust_passes_validation => lean_passes_validation
```
john-h-kastner-aws: I don't think this changes how we'll write the fuzz target, but this comment is out of date.

cedar-drt/fuzz/fuzz_targets/symcc-term-drt-equivalent.rs
john-h-kastner-aws, line 1: These targets do share a lot of their core logic, but  I still have to ask if more could be abstracted into a utility.

john-h-kastner-aws (approved):

katherine-hough (approved):

cedar-drt/fuzz/fuzz_targets/symcc-term-drt-equivalent.rs
cdisselkoen: Factored out another few helper functions in the most recent commits.

cedar-drt/fuzz/fuzz_targets/symcc-term-drt-always-allows.rs
cdisselkoen: Removed all occurrences of this comment.

PR #845, opened 2026-01-30
Update to Lean 4.27.0

john-h-kastner-aws:
I've also committed `cedar-lean/lake-manifest.json` which I think should help avoid some build errors. After pulling previous updates I've seen errors building batteries which I think happen when the version in `.lake/packages/` is still for the previous lean release. I'm hoping committing `lake-manifest.json` will cause lake to fetch the new batteries version automatically.


cdisselkoen:
I don't know what's different between my local environment and CI. It's the same commit of `cedar`; the same Rust version (1.93.1); the same commands AFAICT. CI is using Ubuntu 24.04 and I'm also on Linux (a different distro, AL2023).

PR #772, opened 2025-10-20
Add memory limit to run CVC5 and periodically restart solvers

shaobo-he-aws:
*Issue #, if available:*

*Description of changes:*

PR #848, opened 2026-02-02
add target for DRT of SymCC counterexamples

cdisselkoen:
This new target performs DRT on the Decoder, Concretizer, Extractor components -- everything involved in turning solver model output into valid `Request`/`Entities`. 

The implementation in this PR uses the Rust side to generate a query for an arbitrary `Implies` operation and run it through the solver. If the solver produces a counterexample, we ship the raw model text from the solver, as well as the original policies and schema, over to the Lean. This is sufficient information for Lean to run its own decoder/concretizer/extractor without having to run a solver itself (and ensuring that Rust and Lean are starting from exactly the same model). Then we require the produced `Request`/`Entities` to be exactly identical.

### DIFF
--- a/cedar-drt/fuzz/src/symcc.rs
+++ b/cedar-drt/fuzz/src/symcc.rs
@@ -406,6 +406,9 @@ impl solver::Solver for WrappedLocalSolver {
 
 pub struct SymCCWithUsageLimit {
     pub symcc: CedarSymCompiler<WrappedLocalSolver>,
+    /// Maximum number of `CedarSymCompiler` calls (more specifically, calls of
+    /// `get_solver()`, not all of which are actual solver interactions) before
+    /// we kill the solver process and start a new one
     usage_count: usize,
 }
 


cedar-drt/fuzz/src/symcc.rs
```diff
@@ -367,6 +371,110 @@ impl<'a> Arbitrary<'a> for TwoPolicySetFuzzTargetInput {
...
+    }
+    async fn get_model(&mut self) -> Result<Option<String>, SolverError> {
+        self.raw_model = self.solver.get_model().await?;
+        Ok(self.raw_model.clone())
+    }
+}
+
+pub struct SymCCWithUsageLimit {
+    pub symcc: CedarSymCompiler<WrappedLocalSolver>,
+    usage_count: usize,
```
john-h-kastner-aws, line 412: Could use some docs on why we have this limit/what it's limiting

john-h-kastner-aws (approved):

cedar-drt/fuzz/src/symcc.rs
cdisselkoen: This limit was added in #772, with no PR description but I'm guessing to avoid some kind of OOM for long-running (much-reused) solver processes. I'll add a comment for _what_ it's limiting, at least.

PR #849, opened 2026-02-02
better algorithm for avoiding duplicate policy IDs in multipolicy fuzz targets

cdisselkoen:
This one should deterministically avoid _all_ duplicates, not just with high probability


cedar-drt/fuzz/src/symcc.rs
```diff
@@ -195,21 +199,25 @@ fn arbitrary_policies(
...
-            // If we find a policy with a duplicate ID, append `_{idx}` to its ID.
-            // This is highly likely, but not guaranteed, to be a unique ID --
-            // we could get very unlucky, and the new ID could be a duplicate of a
-            // _different_ policy's ID.
-            updates.push((other_idx, PolicyID::from_string(format!("{id}_{idx}"))));
+        if !seen.insert(id.clone()) {
+            // If we've seen this ID already, we need to change it.
+            // We'll do this by appending `_{idx}` to the ID.
+            let mut new_id = format!("{id}_{idx}");
+            while !seen.insert(PolicyID::from_string(new_id.clone())) {
```
katherine-hough, line 216: An alternative to this that we use in [policy_set.rs](https://github.com/cedar-policy/cedar-spec/blob/main/cedar-policy-generators/src/policy_set.rs#L48) is to just skip duplicate policy ids. What you have here should be fine. I was originally worried about the amount of time spent in this loop if a lot of the policy ids are duplicates which is not uncommon with a guided fuzzer, but because you have appended the index this should be relatively infrequent (even if the fuzzer is slightly biased towards additional iterations of this loop because of the hit count feedback).

john-h-kastner-aws, line 216: I think `seen` could store `&str` to avoid cloning the ids strings

cdisselkoen: Tried this, but at this line, we can't store a reference to `new_id` which will be dropped at the end of the loop iteration

cdisselkoen: I suppose we could put `Cow` in the hashset

john-h-kastner-aws: presumably doesn't matter too much

john-h-kastner-aws: `PolicyId::from_string` doesn't parse right? If it does then it could matter, but I'm pretty sure it doesn't

cdisselkoen: It [does not](https://github.com/cedar-policy/cedar/blob/b1611dcff61d3ba0a54f15c2f02bdfe4c9cfe190/cedar-policy-core/src/ast/policy.rs#L1941). And under the hood it is `SmolStr`, so the clone is cheap anyway

PR #896, opened 2026-03-04
Simplifications to partial_eval_preserves_typeof

john-h-kastner-aws:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cedar-lean/Cedar/Thm/TPE/PreservesTypeOf.lean
+++ b/cedar-lean/Cedar/Thm/TPE/PreservesTypeOf.lean
@@ -254,12 +254,12 @@ a residual with the same type.
 theorem partial_eval_preserves_typeof :
   ∀ res, PEPreservesTypeOf res
 := by
-  intro res env h_wt
+  intro res env
   cases res with
   | val v ty =>
     simp [TPE.evaluate, Residual.typeOf]
   | var v ty =>
-    intro preq pes
+    intro _ preq
     simp only [Residual.typeOf, TPE.evaluate]
     unfold varₚ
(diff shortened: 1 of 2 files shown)


cedar-lean/Cedar/Thm/TPE/PreservesTypeOf.lean
```diff
@@ -28,29 +28,22 @@ open Cedar.Spec
...
-  (res : Residual)
-  (req : Request)
-  (preq : PartialRequest)
-  (es : Entities)
-  (pes : PartialEntities) : Prop :=
-  InstanceOfWellFormedEnvironment req es env →
-  RequestAndEntitiesRefine req es preq pes →
-  Residual.WellTyped env res →
-  (TPE.evaluate res preq pes).typeOf = res.typeOf
+abbrev PEPreservesTypeOf (res : Residual) : Prop :=
```
cdisselkoen, line 31: Not related to this PR, but -- curious why this is `abbrev` instead of `def`?

cdisselkoen (approved):
Great simplification. I assume you checked whether we can remove any hypotheses from the proofs up the stack, similar to how we're removing unneeded hypotheses from this definition?

victornicolet (approved):

cedar-lean/Cedar/Thm/TPE/PreservesTypeOf.lean
john-h-kastner-aws: It compiles with `def`, so idk. Quick search says Lean is more willing to automatically unfold abbrevs.

PR #898, opened 2026-03-05
Split TPE proofs into seperate files per expression kind

john-h-kastner-aws:
massive diff, but there is no change to content.

I wanted to do this split before trying to update the proofs.


cdisselkoen (approved):
Good call for making this a separate PR

victornicolet (approved):

PR #908, opened 2026-03-13
Generate `context` variable for type-directed targets

john-h-kastner-aws:
The type-directed generator will currently never generate the `context` variable, leaving it untested specifically in the symcc targets. It is generated by the non-type-directed generators.

### DIFF
--- a/cedar-policy-generators/src/expr.rs
+++ b/cedar-policy-generators/src/expr.rs
@@ -666,8 +666,14 @@ impl ExprGenerator<'_> {
                 }
                 Type::Record(m) => {
                     if max_depth == 0 || u.len() < 10 {
-                        // no recursion allowed
-                        Err(Error::TooDeep)
+                        Ok(uniform!(
+                            u,
+                            ast::Expr::var(ast::Var::Context),
+                            ast::Expr::record(m.iter().filter_map(|(k, qty)| {
+                                qty.required.then(|| (k.clone(), ast::Expr::val(false)))
+                            }))
+                            .unwrap()
+                        ))
                     } else {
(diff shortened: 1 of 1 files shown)


cedar-policy-generators/src/expr.rs
```diff
@@ -666,8 +666,14 @@ impl ExprGenerator<'_> {
...
-                        // no recursion allowed
-                        Err(Error::TooDeep)
+                        Ok(uniform!(
+                            u,
+                            ast::Expr::var(ast::Var::Context),
+                            ast::Expr::record(m.iter().filter_map(|(k, qty)| {
+                                qty.required.then(|| (k.clone(), ast::Expr::val(false)))
+                            }))
+                            .unwrap()
+                        ))
```
cdisselkoen: is it OK that this won't always be type-correct, even though this is the type-directed code path?

john-h-kastner-aws: For the `ast::Expr::var(ast::Var::Context)` arm, it will often not be well-typed, but it mirrors the behavior for principal and resource variables on entity types. For the `record` case, I figure at least getting all the required attribute present would go a long way to generating type-correct expressions.

cdisselkoen (approved):

cdisselkoen (approved):

cedar-policy-generators/src/expr.rs
john-h-kastner-aws: pushed a larger change to generate will-typed record literals. Should also improve the set case by generating non-empty sets

katherine-hough (approved):

PR #913, opened 2026-03-19
Implement PST and EST/AST roundtrip fuzzing after desugaring

victornicolet:
Implements a first version of fuzzing for PST with:
- generators for `pst::Template` and all subtypes of the template.
- roundtrip testing for `pst -> est -> pst`.
- roundtrip testing for `ast -> pst -> ast -> pst`. Going directly from `pst -> ast` would cause desugaring of pst's operators.

This was mostly implemented by Claude Opus 4.6 with some suggestions on how to organize the generators and clarifications on edges cases. 
One of the current edge cases is that some extension methods `.lessThan` and `.greaterThan` are converted into operators (in PST, everything is an operator). 

This PR requires the fix in  https://github.com/cedar-policy/cedar/pull/2242 to be merged first.


cedar-drt/fuzz/fuzz_targets/pst-roundtrip.rs
```diff
@@ -0,0 +1,133 @@
...
+        Ok(arbitrary::size_hint::and_all(&[
+            Schema::arbitrary_size_hint(depth)?,
+            HierarchyGenerator::size_hint(depth),
+            Schema::arbitrary_policy_size_hint(&SETTINGS, depth),
+        ]))
+    }
+}
+
+/// Compare two PST templates for equivalence, ignoring policy IDs.
+fn check_pst_equivalence(original: &pst::Template, roundtripped: &pst::Template) {
```
katherine-hough: Could you pull out the code shared between this and pst_est_roundtrip::check_pst_equivalence to avoid duplicating it?

cedar-drt/fuzz/fuzz_targets/pst-roundtrip.rs
```diff
@@ -0,0 +1,133 @@
...
+        original, roundtripped
+    );
+    assert_eq!(
+        original.clauses(),
+        roundtripped.clauses(),
+        "Clauses mismatch.\nOriginal: {:?}\nRoundtripped: {:?}",
+        original,
+        roundtripped
+    );
+    // Ignore annotations: AST→PST→AST→PST may reorder or normalize annotation keys
```
katherine-hough: Could we do a partial check for the portion of the annotations that are preserved?

cedar-drt/fuzz/fuzz_targets/pst-roundtrip.rs
```diff
@@ -0,0 +1,133 @@
...
+}
+
+// AST → PST → AST → PST roundtrip, comparing at the PST level.
+// PST → AST is lossy (desugars >=, >, !=), so we go through AST twice
+// and compare the two PST representations which should be stable.
```
katherine-hough: I'm confused about why the second pass is necessary if we have started from an AST template. What sort of input would cause there to be a difference between pst1 and pst2?

victornicolet: The PST → AST might have caused some unwanted changes in representation or might fail. For example, if someone in the future forgets to properly update the `pst::BinaryOp` representation to match exactly the extensions in the AST, it might be possible to properly convert the AST to PST (operator name parsing had been updated properly) but fail in the other direction with the same PST (operator name printing hasn't been updated). There is a unit test checking this, but some other places could also have similar mismatches.

katherine-hough: Wouldn't we want the fuzzing target to fail in this situation?

cedar-drt/fuzz/fuzz_targets/pst-roundtrip.rs
victornicolet: Yes, will implement that.

cedar-drt/fuzz/fuzz_targets/pst-roundtrip.rs
john-h-kastner-aws: I think I agree with @katherine-hough. `AST -> PST -> AST` feels like it should be stable, the `PST -> AST` translation shouldn't be lossy if we're starting from an AST in the first place.

cedar-drt/fuzz/fuzz_targets/pst-roundtrip.rs
```diff
@@ -0,0 +1,133 @@
...
+    initialize_log();
+    let p: StaticPolicy = input.policy.into();
+    let ast_template: Arc<Template> = p.into();
+
+    debug!("Running AST→PST→AST→PST roundtrip on: {:?}", ast_template);
+
+    // AST → PST (first conversion)
+    let pst1: pst::Template = match ast_template.as_ref().clone().try_into() {
+        Ok(t) => t,
+        Err(_) => return, // skip if conversion fails
```
john-h-kastner-aws: Can we limit this to some set of expected errors?

cedar-drt/fuzz/fuzz_targets/pst-roundtrip.rs
```diff
@@ -0,0 +1,133 @@
...
+use cedar_policy_core::pst;
+use cedar_policy_generators::schema_gen::SchemaGen;
+use cedar_policy_generators::{
+    abac::ABACPolicy, hierarchy::HierarchyGenerator, schema::Schema, settings::ABACSettings,
+};
+use libfuzzer_sys::arbitrary::{self, Arbitrary, Unstructured};
+use log::debug;
+use std::sync::Arc;
+
+/// Fuzz target input: a policy generated via the standard ABAC generators.
```
john-h-kastner-aws: why not use the pst generator?

cedar-drt/fuzz/fuzz_targets/pst-roundtrip.rs
victornicolet: Yes it's now stable with https://github.com/cedar-policy/cedar/pull/2243 merged.

cedar-drt/fuzz/fuzz_targets/pst-roundtrip.rs
victornicolet: With the recent PRs in cedar, this should never fail anymore.

cedar-drt/fuzz/fuzz_targets/pst-roundtrip.rs
victornicolet: Changed to use the PST generator.

john-h-kastner-aws: I suppose the other question is if there's any AST that can't be obtained from the first `PST -> AST` step. I don't think there should be, but if there is it could make sense to _also_ test with the AST generators. But I don't think so .

katherine-hough (approved):

cedar-drt/fuzz/fuzz_targets/pst-roundtrip.rs
victornicolet: I don't think so, other than the `tolerant-ast` featured AST. But I think in general we will add more fuzzing targets for the PST, and one starting from the ast (and/or text) would make sense.

john-h-kastner-aws (approved):

PR #923, opened 2026-04-06
Add Cedar type generator respecting max_width and max_depth

john-h-kastner-aws:
This change limits the depth (and width) of generated types. This should avoid a recent failure we saw in differential testing where the serialized form a symcc Term (whose depth should be roughly equal to the expression it was compiled from) exceeded the recursion limit of serde, causing an error when deserializing the Term computed in the Lean model.

We already limit depth when generating expressions, but we don't impose any limit on the size of intermediate types used inside the generator. For example, to create an `==` expression, we choose a type for the operands, and then generate expressions of that type with `max_depth - 1`. If the depth of the intermediate type was greater than the renaming expression depth, we generate the expression by structural recursion on the type rather than on `max_depth` (prior to https://github.com/cedar-policy/cedar-spec/pull/908 the generator would instead fail to generate any expression in this case). This still works to avoid infinite recursion , but it makes the actual maximum depth limited only by whatever recursion guard the `arbitrary` crate imposes. With this change the depth of expression should actually respect `max_depth`. 

Record attributes name are now pull from the constant pool and entity types are take from the schema, which should make it easier to generate well-typed expressions. 

Also removes the `enable_extensions` flag. This was always set to `true`, and if I kept it I would need to handle it properly in the new type generator which is too much trouble for something we don't use.

### DIFF
--- a/cedar-policy-generators/src/expr.rs
+++ b/cedar-policy-generators/src/expr.rs
@@ -22,6 +22,7 @@ use crate::hierarchy::{generate_uid_with_type, Hierarchy};
 use crate::schema_gen::SchemaGen;
 use crate::settings::ABACSettings;
 use crate::size_hint_utils::{size_hint_for_choose, size_hint_for_range, size_hint_for_ratio};
+use crate::types::TypeGenerator;
 use crate::{accum, gen, gen_inner, uniform};
 use arbitrary::{Arbitrary, MaxRecursionReached, Unstructured};
 use cedar_policy_core::ast;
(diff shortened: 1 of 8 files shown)


cedar-policy-generators/src/expr.rs
```diff
@@ -291,16 +299,16 @@ impl ExprGenerator<'_> {
                         2 => Ok(ast::Expr::val(u.arbitrary::<bool>()?)),
                         // == expression, where types on both sides match
                         5 => {
-                            let ty: Type = u.arbitrary()?;
+                            let ty = self.as_type_gen().generate_type(max_depth, u)?;
```
john-h-kastner-aws, line 302: We could make this call with `self.settings.max_depth` without unbounded recursion, but I think it makes sense to limit type depth based on the remaining expression depth.

victornicolet (approved):

katherine-hough (approved):

Issue #284, opened 2024-04-12
Partial Eval DRT


aaronjeline:
### Category

DRT target(s)

### Describe the feature you'd like to request

Once partial evaluation has been formalized, we need to add DRT against the rust impl

### Describe alternatives you've considered

N/A

### Is this something that you'd be interested in working on?

- [X] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

john-h-kastner-aws:
Issue is still relevant, now applying to Type aware partial eval. We have DRT on batched evaluation (which uses partial eval), and property tests on TPE, but we don't have a direct DRT target on TPE

victornicolet:
We now have a DRT target that tests the equivalence between Lean and Rust's API for `is_authorized` TPE (#927 )

john-h-kastner-aws:
I think we can call this resolved

PR #927, opened 2026-04-09
Add FFI + DRT target for TPE is_authorized API

victornicolet:
Adds:
- Lean FFI for the `is_authorized` TPE interface.
- DRT target for TPE's `is_authorized` comparing decisions and residuals.

This is waiting on https://github.com/cedar-policy/cedar/pull/2295 to be merged.

### DIFF
--- a/cedar-drt/fuzz/fuzz_targets/tpe-is-authorized-drt.rs
+++ b/cedar-drt/fuzz/fuzz_targets/tpe-is-authorized-drt.rs
@@ -21,19 +21,19 @@
 use cedar_drt::logger::initialize_log;
 use cedar_drt_inner::{
     fuzz_target,
-    tpe::{TpeFuzzTargetInput, passes_request_validation, passes_validation},
+    tpe::{TpeFuzzTargetInput, passes_policyset_validation, passes_request_validation},
 };
 use cedar_lean_ffi::CedarLeanFfi;
 use cedar_policy::{
     PartialEntities, PartialRequest, Policy, PolicyId, PolicySet, Request, Schema, Validator,
+    pst::{Clause, Expr, UnaryOp},
 };
-use cedar_policy_core::pst;
 use log::debug;
-use std::collections::HashSet;
 use std::convert::TryFrom;
+use std::{collections::HashSet, sync::Arc};
 
 /// Compare Rust and Lean TPE outputs for a single partial request.
-fn compare_tpe_outputs(
+fn test_tpe_is_authorized_equiv(
     ffi: &CedarLeanFfi,
     schema: &Schema,
     policies: &PolicySet,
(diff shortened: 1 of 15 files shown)


cedar-lean-ffi/src/datatypes.rs
```diff
@@ -359,19 +363,10 @@ impl From<Decimal> for RestrictedExpression {
...
-            format!(".000{right}")
-        } else if right < 100 {
-            format!(".00{right}")
-        } else if right < 1000 {
-            format!(".0{right}")
-        } else {
-            format!(".{right}")
-        };
-        RestrictedExpression::new_decimal(format!("{neg}{left}{right}"))
+        RestrictedExpression::new_decimal(format!("{neg}{left}.{right:04}"))
```
victornicolet, line 369: This shouldn't be a change in representation.

cedar-lean-ffi/src/datatypes.rs
```diff
@@ -393,34 +388,24 @@ impl From<IpAddr> for RestrictedExpression {
...
             IpAddr::V4(cidr) => {
                 let addr = cidr.addr.as_u64();
-                let mut addr = format!(
-                    "{a0}.{a1}.{a2}.{a3}",
+                let prefix = cidr.prefix.map_or(32, |p| p.as_u64());
+                // Canonical reprsentation always formats with prefix,
+                // even if IP address is single host /32
+                let addr = format!(
+                    "{a0}.{a1}.{a2}.{a3}/{prefix}",
                     a0 = (addr >> 24) & 0xFF,
```
victornicolet, line 396: This is a change in a path that is used by symcc DRT. It doesn't change the semantics of what's being represented and I didn't get any errors running a few symcc drt targets. But someone who knows that code in more detail might have a better intuition.

cedar-lean-ffi/src/datatypes.rs
```diff
@@ -435,20 +420,17 @@ pub struct Datetime {
 impl TryFrom<Datetime> for RestrictedExpression {
     type Error = cedar_policy_core::parser::err::ParseErrors;
     fn try_from(dt: Datetime) -> Result<Self, Self::Error> {
-        // note this assumes the Lean represents datetimes as milliseconds since Unix epoch
-        let epoch = cedar_policy_core::ast::RestrictedExpr::call_extension_fn(
-            Name::parse_unqualified_name("datetime")?,
-            [cedar_policy_core::ast::RestrictedExpr::val("1970-01-01")],
+        // Match Rust's DateTime canonical form: datetime("1970-01-01").offset(duration("Nms"))
+        let epoch = ast::RestrictedExpr::call_extension_fn(
```
victornicolet, line 424: No change in representation here, the comment just notes that we have a canonical form for `datetime` in Rust, and this representation matches it.

cedar-drt/fuzz/fuzz_targets/tpe-is-authorized-drt.rs
```diff
@@ -0,0 +1,185 @@
...
+            let pst = rp
+                .to_pst()
+                .expect("policy->pst conversion should succeeed for residuals");
+            // Residual should only have one clause
+            let clause = match pst.body().clauses().as_slice() {
+                [Clause::When(x)] => x.clone(),
+                [Clause::Unless(x)] => Arc::new(Expr::UnaryOp {
+                    op: UnaryOp::Not,
+                    expr: x.clone(),
+                }),
```
john-h-kastner-aws, line 129: this case probably can't happen

cedar-lean-ffi/src/datatypes.rs
john-h-kastner-aws: This looks fine to me. I think if we directly compared restricted expressions anywhere this could be a problem, bu this looks like it's only reached through the `impl TryFrom<Env> for cedar_policy_symcc::Env` which calls other cedar library functions that end up evaluating the restricted expressions to values.

john-h-kastner-aws (approved):

cedar-drt/fuzz/fuzz_targets/tpe-is-authorized-drt.rs
victornicolet: Yeah, I just wanted to handle both enum cases where there's a single clause coming out of the conversion to the PST.

katherine-hough (approved):

victornicolet:
Noting that this added a TPE DRT target, as requested in #284 .

Issue #851, opened 2026-02-02
Lean model and proofs for batched-authorization


john-h-kastner-aws:
### Category

Lean formalization

### Describe the feature you'd like to request

#836 added the model and proofs for the partial authorization algorithm and #725 added proofs for batched evaluation.

The Rust library also has has a batched authorization algorithm which is not modeled or proven correct.

https://github.com/cedar-policy/cedar/blob/b1611dcff61d3ba0a54f15c2f02bdfe4c9cfe190/cedar-policy-core/src/batched_evaluator.rs#L87

We should add a Lean model of this algorithm and prove that it is correct. Correctness theorem will be the same as for evaluation but showing that the authorization decision is the same rather than the result of evaluation. 

### Describe alternatives you've considered

.

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

john-h-kastner-aws:
Added in #932

PR #932, opened 2026-04-20
Model and proofs for batched authorization

john-h-kastner-aws:
The existing Proofs only showed equivalence for evaluating a single expression, but the Rust implementation does authorization for sets of policies. This PR adds a batched authorization algorithm and proves it equivalent to concrete authorization, resolving #851

The main theorem is

```lean
theorem batched_authorize_decision_eq_authorize
  {schema : Schema} {policies : List Policy} {req : Request}
  {es : Entities} {response : TPE.Response} {d : Decision} :
  EntityLoader.WellBehaved es loader →
  schema.validateWellFormed = .ok () →
  validateEntities schema es = .ok () →
  batchedAuthorize schema policies req loader iters = .ok response →
  response.decision = some d →
  (Spec.isAuthorized req es policies).decision = d
```

### DIFF
--- a/cedar-lean/Cedar/TPE/BatchedEvaluator.lean
+++ b/cedar-lean/Cedar/TPE/BatchedEvaluator.lean
@@ -82,7 +82,7 @@ def batchedAuthorizeLoop
   (store : PartialEntities) (n : Nat)
   : Response
 :=
-  let resp := isAuthorized.isAuthorizedFromResiduals residuals
+  let resp := isAuthorizedFromResiduals residuals
   if resp.decision.isSome then
     resp
   else match n with
--- a/cedar-lean/Cedar/TPE/BatchedEvaluator.lean
+++ b/cedar-lean/Cedar/TPE/BatchedEvaluator.lean
@@ -75,10 +75,7 @@ def batchedEvaluate
   batchedEvaluateLoop residual req loader Map.empty iters
 
 /--
-The batched evaluation loop for authorization over a list of policies.
-  1. Asks for any new entities referenced by the residual
-  2. Partially evaluates now that new entities are loaded
-  3. Exits if a value has been found or it hits the maximum iteration limit
+The batched authorization loop for authorization over a list of policies.
 -/
 def batchedAuthorizeLoop
   (residuals : List ResidualPolicy) (req : Request) (loader : EntityLoader)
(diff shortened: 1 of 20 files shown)


cedar-lean/Cedar/TPE/Input.lean
```diff
@@ -121,7 +121,7 @@ where
     let (uid, _) := p
     match es.find? uid with
     | .some entry₁ => actionEntityIsValid uid entry₁
-    | _            => false
+    | _            => true
```
john-h-kastner-aws, line 124: This change lets TPE/batched authorization work without providing all actions in partial entities. We require actions be present when validating concrete entities so that `in` correctly handles ancestors of actions defined in the schema. For partial evaluation we can safely treat the ancestors of a missing action as unknown, so we don't need this condition.

See some prior discussion here https://github.com/cedar-policy/cedar-spec/pull/922#discussion_r3034185764

cedar-lean/CedarFFI/Main.lean
```diff
@@ -804,55 +804,29 @@ Note that the time is only the encoder and _not_ policy compilation to Term
...
-        .ok te
-      else
-        .error s!"invalid request"
-    | .none => .error s!"invalid environment derived from request"
-  return (expr, request, entities, iteration)
+  return (schema, policySet, request, entities, iteration)
 
 /--
-  `req`: binary protobuf for an `CheckPolicySetRequest`
+  `req`: binary protobuf for a `BatchedEvaluationRequest`
```
victornicolet: BatchedAuthorizationRequest I think.

cedar-lean/CedarFFI/Main.lean
```diff
@@ -804,55 +804,29 @@ Note that the time is only the encoder and _not_ policy compilation to Term
 
 /--
   `req`: binary protobuf for an `BatchedEvaluationRequest`
```
victornicolet: `BatchedAuth...`?

cedar-lean/Cedar/TPE/BatchedEvaluator.lean
```diff
@@ -72,8 +72,45 @@ def batchedEvaluate
...
+      let newStore := newEntities ++ store
+
+      let residuals : List ResidualPolicy := residuals.map λ rp =>
+        ⟨rp.id, rp.effect, Cedar.TPE.evaluate rp.residual req.asPartialRequest newStore⟩
+      batchedAuthorizeLoop residuals req loader newStore n
+
+/--
+Evaluate an authorization request using an EntityLoader instead of a full Entities store.
+
+Performs a maximum of `iter` number of calls to `loader`, but may perform fewer when an authorization decisions is reached.
```
victornicolet: nit `decisions` -> `decision`

victornicolet (approved):

cedar-lean/Cedar/TPE/Input.lean
lianah: Very minor: is it worth adding a code comment here?

lianah (approved):

PR #933, opened 2026-04-20
Generate templates in fuzz targets

### LINKED ISSUE #36: Additional fuzz testing for templates
### Category

New DRT target

Currently, `rbac` is the only fuzzing target that generates templates with slots (i.e., `?principal`, `?resource`). To improve coverage, we should consider adding slot generation to other targets as well.

Templates with slots are currently not generated by any targets that test parsing (e.g., `pp`, `formatter`) or validation (e.g., `validation-drt`, `validation-pbt`).

N/A

### Additional context

_No response_

john-h-kastner-aws:
Enhances policy generation to generate template linked policies instead of only generating static policies. Impacts a lot of fuzz targets. I've run a few targets locally to check that it works (links are generated, and it doesn't immediately crash), but this touches a lot of targets, so we might see something go wrong on the first nightly test run.

On the lean side, the bindings for template slots get substituted before evaluation. For Rust side, we construct a policy set containing a single template with a single link against that template. Most targets were already written using a `PolicySet` containing a single policy, so this didn't require major changes to fuzz targets.

Some fuzz targets still generate only static policies. This includes all symcc targets (the symcc implementations does not support links) and some formatting/roundtrip targets (can't support links if they go through cedar policy text, but should be extended to fuzz unlinked templates)  

Fixes #36

### DIFF
--- a/cedar-drt/fuzz/fuzz_targets/batched-evaluation-drt.rs
+++ b/cedar-drt/fuzz/fuzz_targets/batched-evaluation-drt.rs
@@ -19,8 +19,8 @@ use cedar_drt::logger::initialize_log;
 use cedar_drt_inner::{abac::FuzzTargetInput, fuzz_target};
 
 use cedar_lean_ffi::CedarLeanFfi;
-use cedar_policy::{Policy, PolicySet, Schema, TestEntityLoader};
-use cedar_policy_generators::{abac::ABACPolicy, policy::GeneratedPolicy};
+use cedar_policy::{Schema, TestEntityLoader};
+use cedar_policy_core::batched_evaluator::err::BatchedEvalError;
 
 // This target tests a property that batched evaluation, if succeeds, should
 // produce the same authorization decision based on the Lean model output
@@ -29,15 +29,7 @@ fuzz_target!(|input: FuzzTargetInput<true>| {
     initialize_log();
 
     if let Ok(schema) = Schema::try_from(input.schema) {
-        let ABACPolicy(GeneratedPolicy::Static(policy)) = input.policy else {
-            panic!(
-                "batched-evaluation-drt needs #932 to support linked policies\n{:?}",
-                input.policy
-            )
-        };
-        let policy = Policy::from(policy);
-        let mut policyset = PolicySet::new();
-        policyset.add(policy).unwrap();
+        let policyset = input.policy.into_policy_set();
         let mut loader = TestEntityLoader::new(&input.entities);
         log::debug!("policy: {policyset}");
         let iteration = (FuzzTargetInput::<true>::settings().max_depth + 1) as u32;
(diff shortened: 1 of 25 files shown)


cedar-drt/fuzz/fuzz_targets/batched-evaluation-drt.rs
```diff
@@ -28,7 +29,13 @@ fuzz_target!(|input: FuzzTargetInput<true>| {
     initialize_log();
 
     if let Ok(schema) = Schema::try_from(input.schema) {
-        let policy = Policy::from(input.policy);
+        let ABACPolicy(GeneratedPolicy::Static(policy)) = input.policy else {
+            panic!(
+                "batched-evaluation-drt needs #932 to support linked policies\n{:?}",
+                input.policy
+            )
+        };
```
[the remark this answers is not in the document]
john-h-kastner-aws: See #932

cedar-drt/fuzz/fuzz_targets/roundtrip.rs
```diff
@@ -67,8 +69,8 @@ impl<'a> Arbitrary<'a> for FuzzTargetInput {
 // AST --> text --> CST --> AST
 // Print a policy to a string and parse it back using the standard AST parser.
 // Panics if parsing fails.
-fn round_trip_ast(p: &StaticPolicy) -> StaticPolicy {
-    parse_policy(None, &p.to_string()).unwrap_or_else(|err| {
+fn round_trip_ast(p: &Policy) -> StaticPolicy {
+    parse_policy(None, &p.to_cedar().unwrap()).unwrap_or_else(|err| {
```
victornicolet, line 73: I don't think`to_string` and `to_cedar` are strictly equivalent (they may be in this case where the policy comes from ast), but `to_cedar` probably matches the intent of the target. Just checking this is intentional.

john-h-kastner-aws: Yes, `to_cedar` is intentional here so that we get cedar source regardless of how `Policy` was constructed. I'm not sure exactly how `to_string` behaves on a generated `cedar_policy::Policy`, it probably gives the cedar text representation via est, but `to_cedar` feels clearer.

cedar-policy-generators/src/policy.rs
```diff
@@ -26,19 +26,103 @@ use cedar_policy_core::ast::{
...
+pub enum GeneratedPolicy {
+    /// Generated static policy. The `GeneratedTemplate` will not have any slots
+    Static(GeneratedTemplate),
+    /// Generated linked policy
+    Link {
+        /// Generated template which will have at least one slot
+        template: GeneratedTemplate,
+        /// Bindings for slots in `template`
+        link: GeneratedLinkedPolicy,
+    },
```
victornicolet, line 45: nit: I'd call this variant `Linked` or `TemplateLinked` and rename the link type to `GeneratedLink`? The pre-existing `GeneratedLinkedPolicy` naming confuses me.

victornicolet (approved):

katherine-hough (approved):

victornicolet (approved):

PR #935, opened 2026-04-22
Roundtrip through JSON in cedar-to-json fuzzing target

luxas:
*Description of changes:*

I realized the `cedar-to-json` fuzz target actually doesn't go through the serde serialize-deserialize roundtrip. It should be lossless, but we do apply optimizations to the EST deserializer, so might be worth covering it. We do cover that roundtrip indirectly in `policy-set-roundtrip.rs`, but that starts from valid ASTs, not any Cedar policy string like here.

### DIFF
--- a/cedar-drt/fuzz/fuzz_targets/convert-policy-cedar-to-json.rs
+++ b/cedar-drt/fuzz/fuzz_targets/convert-policy-cedar-to-json.rs
@@ -20,6 +20,7 @@ use thiserror::Error;
 use cedar_drt::{check_for_internal_errors, check_policy_set_equivalence};
 use cedar_drt_inner::fuzz_target;
 
+use cedar_policy_core::ast;
 use cedar_policy_core::est;
 use cedar_policy_core::parser;
 
@@ -31,6 +32,8 @@ enum ESTParseError {
     #[error(transparent)]
     #[diagnostic(transparent)]
     ESTToAST(#[from] Box<est::PolicySetFromJsonError>),
+    #[error(transparent)]
+    Serde(#[from] Box<serde_json::error::Error>),
 }
 
 // Given some Cedar source, assert that parsing it directly (parsing to CST,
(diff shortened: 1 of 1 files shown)


cedar-drt/fuzz/fuzz_targets/convert-policy-cedar-to-json.rs
```diff
@@ -46,22 +49,33 @@ fuzz_target!(|src: String| {
...
-                    est_result.and_then(|est: est::PolicySet| {
+                
+                let ast_from_est_result: Result<ast::PolicySet, ESTParseError> = cst_node
+                    .try_into() // CST -> EST
+                    .map_err(|e: parser::err::ParseErrors| ESTParseError::from(Box::new(e)))
+                    .and_then(|est: est::PolicySet| {
+                        // EST -> JSON
+                        // Make this extra roundtrip through JSON, as the user of cedar-to-json would actually see the JSON string,
+                        // and first then parse it. This ensures the manual serde::Deserialize implementations roundtrip for valid Cedar strings.
+                        serde_json::to_value(est).map_err(|e| ESTParseError::from(Box::new(e)))
```
john-h-kastner-aws: IIUC, this is going to `serde_json::Value, not text, right? I think that's what we want.

john-h-kastner-aws (approved):

cedar-drt/fuzz/fuzz_targets/convert-policy-cedar-to-json.rs
luxas: Oh yeah, that's what I meant, but seems I mistyped in the hurry 🤦 will update

cedar-drt/fuzz/fuzz_targets/convert-policy-cedar-to-json.rs
```diff
@@ -46,22 +49,33 @@ fuzz_target!(|src: String| {
...
-                let ast_from_est_result: Result<_, ESTParseError> =
-                    est_result.and_then(|est: est::PolicySet| {
+
+                let ast_from_est_result: Result<ast::PolicySet, ESTParseError> = cst_node
+                    .try_into() // CST -> EST
+                    .map_err(|e: parser::err::ParseErrors| ESTParseError::from(Box::new(e)))
+                    .and_then(|est: est::PolicySet| {
+                        // EST -> JSON
+                        // Make this extra roundtrip through JSON, as the user of cedar-to-json would actually see the JSON string,
+                        // and first then parse it. This ensures the manual serde::Deserialize implementations roundtrip for valid Cedar strings.
```
victornicolet: nit: "actually see the JSON string first, and then parse it" maybe

victornicolet (approved):

cedar-drt/fuzz/fuzz_targets/convert-policy-cedar-to-json.rs
luxas: fixed

luxas:
Fixed the comment, should be ready to merge 👍

victornicolet (approved):

PR #939, opened 2026-04-30
fix for recursion limit reached in DRT code

victornicolet:
`serde_json`'s recursion limit is easily reached in the DRT code. This is one easy solution, but it might cause stack overflow instead of recursion limit exceeded.

### DIFF
--- a/cedar-drt/Cargo.toml
+++ b/cedar-drt/Cargo.toml
@@ -8,14 +8,20 @@ publish = false
 libfuzzer-sys = "0.4"
 cedar-lean-ffi = { path = "../cedar-lean-ffi", version = "4.4.0" }
 cedar-policy = "4.4.0"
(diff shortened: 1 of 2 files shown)


cedar-drt/Cargo.toml
```diff
@@ -8,14 +8,20 @@ publish = false
...
 cedar-testing = { path = "../cedar/cedar-testing", version = "4.4.0" }
-cedar-policy-generators = { path = "../cedar-policy-generators", version = "4.0.0", features = ["cedar-policy"] }
+cedar-policy-generators = { path = "../cedar-policy-generators", version = "4.0.0", features = [
+    "cedar-policy",
+] }
 env_logger = "0.11"
 log = "0.4"
 miette = "7.1.0"
-serde_json = "1.0.140"
+serde_json = { version = "1.0", features = ["unbounded_depth"] }
```
victornicolet, line 24: This is the only real change. Other changes are because my formatter ran on Cargo.toml.

PR #940, opened 2026-04-30
fix for recursion limit reached in DRT code

victornicolet:
A possible fix to recursion limit reached in the fuzzing code. We have the risk of getting stack overflows, but this at least allows testing for larger policies.

### DIFF
--- a/cedar-drt/Cargo.toml
+++ b/cedar-drt/Cargo.toml
@@ -8,14 +8,20 @@ publish = false
 libfuzzer-sys = "0.4"
 cedar-lean-ffi = { path = "../cedar-lean-ffi", version = "4.4.0" }
 cedar-policy = "4.4.0"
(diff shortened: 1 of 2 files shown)


katherine-hough:
I would much rather get an error than a stack overflow. The documentation for [`disable_recursion_depth`](https://docs.rs/serde_json/latest/serde_json/struct.Deserializer.html#method.disable_recursion_limit) indicates that we can guard against stack overflows by "wrapping your Deserializer in the dynamically growing stack adapter provided by the serde_stacker crate". But I assume if we did this we would then risk getting the overflow on the EST to AST conversion. Since this is an unstructured target, and we cannot control the depth with the generator, we might just want to ignore inputs that fail with serde's `RecursionLimitExceeded`? We could also use serde_stacker and then implement an iterative depth check on the EST before trying to convert into the AST?

PR #944, opened 2026-05-01
Fixed `convert-policy-cedar-to-json` failing due to hitting serde's recursion limit

### LINKED ISSUE #942: Fix recursion limit being hit in policy-cedar-to-json fuzzing target
The property based testing target `policy-cedar-to-json` can hit recursion limits in the testing code as opposed to the code being tested.

Some comment on a potential fix:

I would much rather get an error than a stack overflow. The documentation for [`disable_recursion_depth`](https://docs.rs/serde_json/latest/serde_json/struct.Deserializer.html#method.disable_recursion_limit) indicates that we can guard against stack overflows by "wrapping your Deserializer in the dynamically growing stack adapter provided by the serde_stacker crate". But I assume if we did this we would then risk getting the overflow on the EST to AST conversion. Since this is an unstructured target, and we cannot control the depth with the generator, we might just want to ignore inputs that fail with serde's `RecursionLimitExceeded`? We could also use serde_stacker and then implement an iterative depth check on the EST before trying to convert into the AST?

_Originally posted by @katherine-hough in https://github.com/cedar-policy/cedar-spec/issues/940#issuecomment-4353732207_

katherine-hough:
Fixed `convert-policy-cedar-to-json` failing due to hitting serde's recursion limit

Resolves: https://github.com/cedar-policy/cedar-spec/issues/942
Requires: https://github.com/cedar-policy/cedar/pull/2332

### DIFF
--- a/cedar-drt/fuzz/fuzz_targets/convert-policy-cedar-to-json.rs
+++ b/cedar-drt/fuzz/fuzz_targets/convert-policy-cedar-to-json.rs
@@ -16,6 +16,7 @@
 
 #![no_main]
 use serde::Deserialize as _;
+use serde::Serialize as _;
 use thiserror::Error;
 
 use cedar_drt::{check_for_internal_errors, check_policy_set_equivalence};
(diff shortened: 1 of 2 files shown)


cedar-drt/fuzz/fuzz_targets/convert-policy-cedar-to-json.rs
victornicolet: I thought this would be where the guard on expression height would be placed, to prevent the stack overflow in the next step.

john-h-kastner-aws (approved):

victornicolet (approved):

PR #948, opened 2026-05-06
make symcc fuzz targets use linked policies

victornicolet:
Make symcc fuzz targets use the `ABACPolicies` instead of the `Static-` version.

The corresponding PR on the cedar repo is https://github.com/cedar-policy/cedar/pull/2338

Ran the DRT for ~1 hour on 4 of the targets without issues.


cedar-drt/fuzz/src/symcc.rs
```diff
@@ -277,24 +277,37 @@ fn arbitrary_policies(
     Ok(policies)
 }
 
+fn abac_policies_to_policy_set(policies: Vec<ABACPolicy>) -> PolicySet {
+    let mut ps = PolicySet::new();
+    for p in policies {
+        let single = p.into_policy_set();
+        for policy in single.policies() {
+            ps.add(policy.clone()).unwrap_or(());
```
john-h-kastner-aws, line 285: probably don't want to ignore errors here

Specifically, this might be silently ignoring template links since `PolicySet::add` errors on non-static policies.

Issue #950, opened 2026-05-11
cedar-lean-cli analyze crashes with resource exhausted after leaking cvc5 subprocesses


michaels-zenity:
### Before opening, please confirm:

- [x] I have [searched for duplicate or closed issues](https://github.com/cedar-policy/cedar/issues?q=is%3Aissue+).
- [x] I have read the guide for [submitting bug reports](https://github.com/cedar-policy/cedar/blob/main/CONTRIBUTING.md#bug-reports).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### Bug Category

Other

### Describe the bug

`cedar-lean-cli analyze policies` crashes with `IO error: resource exhausted (error code: 35, resource temporarily unavailable)` once the policy set is large enough to drive a few thousand SMT queries in a single invocation.

`Solver.spawn` in `cedar-lean/Cedar/SymCC/Solver.lean` spawns a fresh cvc5 subprocess per `Solver.cvc5` instance and drops the resulting `IO.Process.Child` immediately, keeping only the stdin/stdout streams. cvc5 exits promptly when its work is done, but because the parent never `wait`s, every query leaves a zombie behind. The zombies accumulate until the per-user PID budget (`RLIMIT_NPROC` / `kern.maxprocperuid`) is full, after which every subsequent `posix_spawn` returns `EAGAIN`.

At any sampled instant only 0–2 cvc5 children are **live** — what saturates is the **zombie** count.

This is distinct from #753, which only addresses the error path. The leak here happens on the success path of every query.

### Expected behavior

`cedar-lean-cli analyze policies` should complete (or fail with a real SMT-level error) regardless of how many internal SMT queries it issues, without leaking subprocesses into the OS process table.

### Reproduction steps

1. Build `cedar-lean-cli` from current `main` per `cedar-lean-cli/README.md`.
2. Generate a 200-rule synthetic policy set with the helper below.
3. Run `cedar-lean-cli analyze policies policies.cedar schema.cedarschema -j`.
4. Observe exit code 2 and `error code: 35` on stderr. Smaller N (e.g. 100, 150) succeeds.
5. While the analyzer runs, sample `ps -o pid,stat,comm -U $(id -u) | awk '$2 ~ /Z/'` — the zombie cvc5 count grows monotonically until the kernel limit is hit.

### Code Snippet

`schema.cedarschema`:

```
entity User;
entity Resource;

action "act0", "act1", "act2", "act3", "act4"
[...]

PR #951, opened 2026-05-11
Solver: reap cvc5 subprocesses to prevent zombie accumulation

michaels-zenity:
*Issue #, if available:* #950

*Description of changes:*
`Solver.spawn` dropped the spawned `IO.Process.Child` immediately, keeping only the stdin/stdout streams. cvc5 exited promptly per query, but with no `wait` from the parent every query left a zombie in the process table. Workloads that drive thousands of SMT queries (e.g. `analyze policies` on a 200+-rule policyset) eventually saturated `RLIMIT_NPROC` / `kern.maxprocperuid`, surfaced as `IO error: resource exhausted (error code: 35)`.

Add a `cleanup : IO Unit` field on `Solver`. `Solver.spawn` populates it with a best-effort `(exit)` send and `proc.wait`; the writer and dummy constructors default it to a no-op. `SolverM.run` invokes `cleanup` in a `try/finally` so the child is reaped on both success and failure paths.

### DIFF
--- a/cedar-lean/Cedar/SymCC/Solver.lean
+++ b/cedar-lean/Cedar/SymCC/Solver.lean
@@ -56,10 +56,22 @@ public instance : Lean.ToJson Decision where
 public structure Solver where
   smtLibInput : IO.FS.Stream
   smtLibOutput : Option IO.FS.Stream
+  /--
+    Tear-down action that releases any process/file resources backing the
+    solver. For solvers that spawn an external SMT process this reaps the
+    child to prevent zombie accumulation; for in-memory writer/dummy
+    solvers this is a no-op. Run by `SolverM.run` after the user's
+    `SolverM` action completes (success or failure).
+  -/
+  cleanup : IO Unit := pure ()
 
 public abbrev SolverM (α) := ReaderT Solver IO α
 
-public def SolverM.run (solver : Solver) (x : SolverM α) : IO α := ReaderT.run x solver
+public def SolverM.run (solver : Solver) (x : SolverM α) : IO α := do
+  try
+    ReaderT.run x solver
+  finally
+    solver.cleanup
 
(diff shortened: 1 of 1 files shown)


john-h-kastner-aws (approved):
looks good, thanks!

cedar-lean/Cedar/SymCC/Solver.lean
```diff
@@ -76,7 +88,24 @@ public def spawn (path : String) (args : Array String) : IO Solver := do
...
+    -- becomes a zombie and the analyzer eventually trips
+    -- `RLIMIT_NPROC`/`kern.maxprocperuid` with `EAGAIN` from `posix_spawn`,
+    -- surfaced as `IO error: resource exhausted (error code: 35)`.
+    cleanup := do
+      try
+        stdinStream.putStr "(exit)\n"
+        stdinStream.flush
+      catch _ => pure ()
+      try
+        let _ ← proc.wait
```
john-h-kastner-aws, line 112: one concern on further thought: do we reuse the solver process across multiple `SolverM.run`s anywhere?

michaels-zenity: According to my checks no path runs SolverM.run twice against the same Solver value, so the cleanup can't tear down a process a later run still needs.

victornicolet: Maybe we should add some documentation for future uses that a solver value can only be run once?

victornicolet (approved):

cedar-lean/Cedar/SymCC/Solver.lean
michaels-zenity: Good idea, thank you, added comment

victornicolet (approved):

PR #952, opened 2026-05-12
fuzz target for transitive closure algorithm

victornicolet:
Fuzz target that tests equivalence of using `compute_tc` from scratch vs `repair_tc` (the incremental approach).

### DIFF
--- a/cedar-drt/fuzz/fuzz_targets/tc-repair-equivalence.rs
+++ b/cedar-drt/fuzz/fuzz_targets/tc-repair-equivalence.rs
@@ -73,27 +73,83 @@ impl<'a> Arbitrary<'a> for FuzzTargetInput {
     }
 }
 
-fn make_uid(index: u8) -> EntityUID {
-    EntityUID::with_eid_and_type("Node", &index.to_string()).unwrap()
+struct TestData {
+    // uids cache
+    uids: HashMap<u8, EntityUID>,
 }
 
-/// Build a HashMap of entities with only direct parent edges (no TC computed).
-fn build_entities(num_nodes: u8, edges: &[Edge]) -> HashMap<EntityUID, Arc<Entity>> {
-    let mut parents_map: Vec<HashSet<EntityUID>> = vec![HashSet::new(); num_nodes as usize];
-    for edge in edges {
-        parents_map[edge.child as usize].insert(make_uid(edge.parent));
+impl TestData {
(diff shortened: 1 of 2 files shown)


cedar-drt/fuzz/fuzz_targets/tc-repair-equivalence.rs
```diff
@@ -0,0 +1,207 @@
...
+        Ok(Self {
+            num_nodes,
+            pre_existing_edges,
+            new_edges,
+        })
+    }
+}
+
+fn make_uid(index: u8) -> EntityUID {
+    EntityUID::with_eid_and_type("Node", &index.to_string()).unwrap()
```
john-h-kastner-aws: nit: could be worth caching these so we don't spend time parsing

cedar-drt/fuzz/fuzz_targets/tc-repair-equivalence.rs
```diff
@@ -0,0 +1,207 @@
...
+            "repair_tc result differs from compute_tc (DAG path)"
+        );
+        // Independent verification: enforce_tc_and_dag must pass on repair_tc output.
+        enforce_tc_and_dag(&incremental_entities)
+            .expect("enforce_tc_and_dag failed on repair_tc output");
+    }
+
+    // --- Test with enforce_dag=false: TC equivalence on acyclic graphs ---
+    // Only run this if the graph is actually a DAG (already confirmed above).
+    if expected_result.is_ok() {
```
john-h-kastner-aws, line 220: Why do we need this condition. If we're not enforcing DAG could we run this checks on graphs with cycles?

We use the same TC code for computing the closure of entity type membership in the schema where don't enforce acyclicity. We never have to repair TC for that graph, but still feels worth testing for robustness.

john-h-kastner-aws (approved):

cedar-drt/fuzz/fuzz_targets/tc-repair-equivalence.rs
victornicolet: If the graph is not a DAG and `enforce_dag` is false, the implementation of `repair_tc` *will* return incorrect results for the TC (and different from compute_tc).
Maybe we should remove the option to enforce_dag in `repair_tc` since it's always set to true in non-testing code.

katherine-hough (approved):

Issue #298, opened 2024-04-29
Improve the consistency of naming and definitions in `Cedar.Data` and `Cedar.Thm.Data`


emina:
### Category

Lean formalization

### Describe the feature you'd like to request

We accumulated some redundancies and stylistic inconsistencies across the shared data structure definitions and lemmas.  Let's fix these by consolidating the definitions:

Definitions in `Cedar.Data`:

* `Map` has both `.kvs` and `.toList`, which are synonyms. Drop one.
* The definition of `Membership.mem` for `Map`s is inconsistent with `contains` and `find?`.  Let's make them consistent.
* `Set` has both `.elts` and `.toList`, which are synonyms. Drop one. 

Theorems in `Cedar.Thm.Data`:

* We are inconsistent about when we make arguments to lemmas implicit vs explicit.  Let's adopt a consistent style in line with Lean's Std and MathLib:  an argument is explicit only when it can't be inferred from the preconditions. 
* We are inconsistent about expressing lemmas as a single `iff` lemma versus two lemmas for both directions.  Let's use a single `iff` lemma when both directions have the same preconditions.

### Describe alternatives you've considered

N/A

### Is this something that you'd be interested in working on?

- [X] 👋 I may be able to implement this feature request
- [X] ⚠️ This feature might incur a breaking change

cdisselkoen:
Another tiny one: argument order.  For theorems that take both a Map and a key (or value or both), consistent ordering of those arguments.

emina:
Added

> Another tiny one: argument order. For theorems that take both a Map and a key (or value or both), consistent ordering of those arguments.

Added to the issue description.

cdisselkoen:
Another one we discussed: renaming `make_mem_list_mem`

MavenRain:
I'll try and make forward progress on this.

PR #954, opened 2026-05-16
Rename Map.make_mem_list_mem to mem_make_mem_list

MavenRain:
Addresses one item from #298:

  > Rename `make_mem_list_mem` because `mem` comes first in the statement of the
   theorem.

  The forward direction
  x ∈ (Map.make xs).toList → x ∈ xs
  is renamed to `mem_make_mem_list`, mirroring the existing converse
  `mem_list_mem_make` (which carries the LHS pattern first).


john-h-kastner-aws (approved):
Makes sense to me.

victornicolet (approved):

PR #955, opened 2026-05-16
Promote simp_do_let to shared Cedar.Thm.Tactics module

MavenRain:
Addresses #279 along the line suggested by @john-h-kastner-aws in the issue
  thread:

  > `simp_do_let` is in main and used by the symcc proofs.  We could adopt in
  the rest of our proofs.

  This is intentionally **not** the path that the closed #311 ("Introduce
  `csimp` tactic") took.  @emina's objection on that PR was that a generic `simp
   only` wrapper does not abstract a specific pattern; her stated rule of thumb
  is that custom tactics should either (a) abstract a specific repeating
  pattern, or (b) be general-purpose decision procedures.  `simp_do_let` clears
  bar (a): it captures the exact `cases h : e <;> simp only [h, Except.bind_err,
   Except.bind_ok, reduceCtorEq]` shape that recurs throughout the proofs.

  ## Changes

  - New file `cedar-lean/Cedar/Thm/Tactics.lean` containing only the
  `simp_do_let` macro and its imports (`Cedar.Thm.Data.Control` for the
  `Except.bind_*` lemmas, plus `Lean.Elab.Tactic.Basic`).
  - `cedar-lean/Cedar/Thm/SymCC/Tactics.lean` `public import`s the new module,
  so existing symcc call sites continue to compile unchanged.  The
  symcc-specific `clean_i` tactic stays where it is.
  - `cedar-lean/Cedar/Thm/Validation/Slice/Reachable/BinaryApp.lean` adopts
  `simp_do_let` at five exact-match sites.

  ## Adoption pattern note (for follow-up PRs)

  When replacing `cases hN : e <;> simp only [hN, Except.bind_err,
  Except.bind_ok, reduceCtorEq] at h` with `simp_do_let e at h`, the macro's
  `cases h' : e` uses a hygienic identifier rather than `hN`.  So the surviving
  branch has one more anonymous hypothesis than the original `cases hN :` form,
  which can silently shift any downstream `rename_i name1 _ _` pattern.  The
  right idiom is to always pair the macro with `rename_i hN` immediately after,
  even when `hN` is not referenced downstream, so the named/anonymous balance
  matches the original.  All five BinaryApp sites in this PR follow that idiom.

### DIFF
--- a/cedar-lean/Cedar/Thm.lean
+++ b/cedar-lean/Cedar/Thm.lean
@@ -15,6 +15,7 @@
 -/
 
 import Cedar.Thm.Data
+import Cedar.Thm.Tactics
 import Cedar.Thm.Authorization
 import Cedar.Thm.PolicySlice
 import Cedar.Thm.SymCC.Opt
--- a/cedar-lean/Cedar/Thm/Tactics.lean
+++ b/cedar-lean/Cedar/Thm/Tactics.lean
@@ -16,21 +16,24 @@
 
 module
 
-public import Cedar.Thm.Tactics
-import Cedar.Thm.SymCC.Data
-public meta import Lean.Meta.Tactic.Clear
+public import Cedar.Thm.Data.Control
+public import Lean.Elab.Tactic.Basic
 
 /-!
-This file contains basic utility tactics specific to symbolic compilation.
-General-purpose tactics live in `Cedar.Thm.Tactics`.
+This file contains general-purpose proof tactics that are shared across
+the Cedar theorems and not tied to any specific subsystem.
 --/
 
 namespace Cedar.Thm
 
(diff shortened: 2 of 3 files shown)


john-h-kastner-aws (approved):
nice. looks good to me

john-h-kastner-aws:
> The right idiom is to always pair the macro with `rename_i hN` immediately after, even when `hN` is not referenced downstream

I'm wondering if we could make this nicer by making the tactic something like `simp_do_let e as hN at h`. I'm not an expert on Lean tactics, but it'd be a nice improvement if it's easy

victornicolet (approved):

john-h-kastner-aws (approved):

PR #957, opened 2026-05-26
fuzz targets testing protobuf & other formats validation agreement

victornicolet:
This PR adds fuzzing targets that test that validation of inputs from decoding protobufs is the same as validation on other inputs. 
Those targets should succeed when we add validation of `models::*`. 

The matching PR in cedar/ : https://github.com/cedar-policy/cedar/pull/2372 

The next pair of PRs will deal with protobuf decoding for schemas.

### DIFF
--- a/cedar-drt/fuzz/src/props.rs
+++ b/cedar-drt/fuzz/src/props.rs
@@ -25,7 +25,7 @@ use cedar_policy::{
 /// The [`Entity`] gets converted to a singleton [`Entities`].
 pub fn entity_to_json_roundtrips(original: Entity) {
     // Wrap in Entities for JSON serialization
-    let entities = Entities::from_entities([original.clone()], None)
+    let entities = Entities::from_entities([original], None)
         .expect("Failed to create Entities from single entity");
     entities_to_json_roundtrips(entities);
 }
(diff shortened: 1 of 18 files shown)


john-h-kastner-aws:
Should we also add these assertions to the test existing tests fuzzing from bytes?

cedar-drt/fuzz/src/props.rs
```diff
@@ -0,0 +1,148 @@
...
+use crate::roundtrip_entities::pretty_assert_entities_deep_eq;
+use cedar_policy::{
+    Entities, Entity, Expression, Policy, PolicySet, Request, RestrictedExpression, Template,
+};
+
+/// An [`Entity`] should roundrtrip through serialization with json and then deserialization.
+/// The [`Entity`] gets converted to a singleton [`Entities`].
+pub fn entity_to_json_roundtrips(original: Entity) {
+    // Wrap in Entities for JSON serialization
+    let entities = Entities::from_entities([original.clone()], None)
```
whonore: Is the `clone` necessary?

whonore (approved):

john-h-kastner-aws (approved):

cedar-drt/fuzz/src/props.rs
victornicolet: That one is indeed unnecessary.

whonore (approved):

PR #961, opened 2026-05-29
Extension Parser for Decimal 

cruisesong7:
Because of the `module` keyword, all declared definitions are default to private. Therefore the `public` is needed for referencing it from the Thm file. The `abbrev` is a cosmetic change that makes the definition reducible by default.

### DIFF
--- a/cedar-lean/Cedar/Thm/Data/String.lean
+++ b/cedar-lean/Cedar/Thm/Data/String.lean
@@ -1,0 +1,95 @@
+import Std.Data.String
+import Batteries.Data.String
+
+/-- `splitOnPPrepend` on a list with no separator produces a single segment. -/
+theorem splitOnPPrepend_none (P : α → Bool) (l acc : List α)
+    (h : ∀ x ∈ l, P x = false) :
+    List.splitOnPPrepend P l acc = [(acc.reverse ++ l)] := by
+  induction l generalizing acc with
+  | nil => rw [List.splitOnPPrepend.eq_def]; simp
+  | cons a t ih =>
+    rw [List.splitOnPPrepend.eq_def]
+    have ha : P a = false := h a (List.mem_cons.mpr (.inl rfl))
+    simp only [ha]
+    rw [ih (a :: acc) (fun x hx => h x (List.mem_cons.mpr (.inr hx)))]
+    simp [List.reverse_cons, List.append_assoc]
+
+/-- `splitOnPPrepend` on a list with exactly one separator produces two segments. -/
+theorem splitOnPPrepend_single (P : α → Bool) (as bs acc : List α) (sep : α)
+    (hsep : P sep = true) (has : ∀ x ∈ as, P x = false) (hbs : ∀ x ∈ bs, P x = false) :
+    List.splitOnPPrepend P (as ++ sep :: bs) acc = (acc.reverse ++ as) :: [bs] := by
+  induction as generalizing acc with
+  | nil =>
+    rw [List.nil_append, List.splitOnPPrepend.eq_def]
+    simp only [hsep, ite_true]
+    rw [splitOnPPrepend_none P bs [] hbs]; simp
+  | cons a t ih =>
+    simp only [List.cons_append]
+    rw [List.splitOnPPrepend.eq_def]
+    have ha : P a = false := has a (List.mem_cons.mpr (.inl rfl))
+    simp only [ha]
+    rw [ih (a :: acc) (fun x hx => has x (List.mem_cons.mpr (.inr hx)))]
+    simp [List.reverse_cons, List.append_assoc]
+
+/-- Splitting `s₁ ++ sep ++ s₂` on `sep` yields `[s₁, s₂]` when neither part contains `sep`. -/
+theorem splitToList_eq (s₁ s₂ : String) (p : Char → Bool) (sep : Char)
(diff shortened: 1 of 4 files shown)


cedar-lean/Cedar/Thm/Data/String.lean
```diff
@@ -0,0 +1,95 @@
+import Std.Data.String
+import Batteries.Data.String
+
+/-- `splitOnPPrepend` on a list with no separator produces a single segment. -/
+theorem splitOnPPrepend_none (P : α → Bool) (l acc : List α)
```
victornicolet: Nit: Maybe the comment should say exactly what the theorem does; there's the additional "accumulator prepend" operation on the list. 
I'd imagine showing another theorem like `splitToList_nosep_eq` showing `l.splitToList p = [ l ]` when `l` has no `p` satisfying character possible.
The name of the theorem also doesn't match the statement.

cedar-lean/Cedar/Thm/Ext/Decimal.lean
```diff
@@ -0,0 +1,385 @@
...
+import Cedar.Thm.Data.String
+
+namespace Cedar.Thm.Decimal
+open Cedar.Spec.Ext
+
+def IsWfStr (s : String) : Prop :=
+ ∃ left right,
+    s.splitToList (· = '.') = [left, right] ∧
+    left ≠ "-" ∧
+    0 < right.length ∧
```
victornicolet, line 11: Is this condition necessary given the `(toNat?' right).isSome` later on?

cedar-lean/Cedar/Thm/Ext/Decimal.lean
```diff
@@ -0,0 +1,385 @@
...
+      if rightNat < 10 then "000" ++ toString rightNat
+      else if rightNat < 100 then "00" ++ toString rightNat
+      else if rightNat < 1000 then "0" ++ toString rightNat
+      else toString rightNat
+    (toString d).splitToList (· = '.') = [leftPart, rightPart] ∧
+    rightPart.length = 4 ∧
+    toInt?' leftPart = some (if d < 0 then -(↑(d.natAbs / Nat.pow 10 4) : Int)
+      else (↑(d.natAbs / Nat.pow 10 4) : Int)) ∧
+    toNat?' rightPart = some rightNat ∧
+    (!leftPart.startsWith "-") = !(d < 0) := by
```
victornicolet, line 50: Can the `!` on both sides of `=` be simplified? In general, the goal of this theorem is a little hard to read.

cedar-lean/Cedar/Thm/Ext/Decimal.lean
```diff
@@ -0,0 +1,385 @@
...
+def IsWfStr (s : String) : Prop :=
+ ∃ left right,
+    s.splitToList (· = '.') = [left, right] ∧
+    left ≠ "-" ∧
+    0 < right.length ∧
+    right.length ≤ DECIMAL_DIGITS ∧
+    (toInt?' left).isSome ∧
+    (toNat?' right).isSome
+
+def computeValue (s : String) : Int :=
```
victornicolet, line 18: Would the proofs have been possible with a function that returns none when the value is out of range?

cedar-lean/Cedar/Thm/Ext/Decimal.lean
```diff
@@ -0,0 +1,385 @@
...
+      have := Nat.div_add_mod d.toInt.natAbs 10000; omega
+    rw [h3, Int.natAbs_of_nonneg hge]
+
+/-- Roundtrip: `Decimal.parse (toString d) = some d`. -/
+theorem parse_toString_roundtrip (d : Decimal) :
+    Decimal.parse (toString d) = some d :=
+  parse_of_isWfStr (toString d) d (toString_isWfStr d) (computeValue_toString d)
+
+/-- `parse` fails iff the string is malformed or its value overflows `Int64`. -/
+theorem parse_eq_none_iff (s : String) :
```
victornicolet, line 321: This theorem and the one above are the main theorems, correct? Might be worth documenting it in the file, or refactoring the lemmas out to another file.

cedar-lean/Cedar/Thm/Ext/Decimal.lean
cruisesong7: No it is not necessary, as one can easily prove an utility lemma for `∀ s : String, (toNat?' s).isSome → s.length > 0`. This is more a style/ergonomics choice. 

Pros to remove: minimal spec
Pros to keep: 1. spec aligns with parser and more clear (no hidden implication). 2. Downstream proof connivence.

cedar-lean/Cedar/Thm/Ext/Decimal.lean
cruisesong7: This is the same style choice as above. But this particular theorem is more acting like a lemma for the main theorems later, and does not carry too much semantic meaning itself. So I would advocate for keeping it as it is right now for the downstream proof convenience.

cedar-lean/Cedar/Thm/Ext/Decimal.lean
cruisesong7: Yes. Better documentations will be introduced with the documentation page I built with [Verso](https://verso.lean-lang.org/), where all the meaningful theorems and defs will be listed.

PR #963, opened 2026-06-01
Prove level checking rules out entityDoesNotExist errors (#642)

### LINKED ISSUE #642: Extend level based slicing soundness proof to rule out entity-does-exist errors
### Category

Lean formalization

The current proof shows that entity slicing cannot change the result of evaluating an expression, but the current statement of the property suggests that a policy that dereferences a non-existent before slicing would get the same error after slicing.

The entity slicing function will return `None` if the traversal of attributes tries to access an entity that doesn't exist. This should ensure that, if entity slicing succeeds, then policy evaluation cannot ever try to dereference an entity that doesn't exist. 

.

### Additional context

_No response_

MavenRain:
*Issue #, if available:* https://github.com/cedar-policy/cedar-spec/issues/642

*Description of changes:*   Closes #642 by proving that level checking, together with a store-closure hypothesis, rules out `entityDoesNotExist` errors entirely (not just that full-store and sliced evaluation agree).

**Interpretation.** Following @john-h-kastner-aws's framing on the issue, this is stated as a standalone validation guarantee, independent of slicing:

`validatesAtLevel p n ∧ EntitiesClosedAtLevel entities request n → NoEntitiesError`

`EntitiesClosedAtLevel` (new) says every entity reachable from the request roots within `n` dereferences is present in the store.  This hypothesis is genuinely new, not a consequence of well-formedness: `InstanceOfSchema` only constrains entries that are present, so `principal.manager.name` can type- and level-check while `principal.manager`'s uid is absent, and the second `getAttr` then errors with `entityDoesNotExist`.  That gap is exactly what the closure condition closes.

**Main results** (in the new `Levels/NoEntitiesError.lean`):
  - `typecheck_policy_with_level_no_dne`: the policy-level guarantee above.
  - `typecheck_policy_with_level_no_dne_on_slice` (in `Levels.lean`): the sliced
    variant, which follows trivially from the existing semantic-equivalence
    soundness theorem.
  - `level_based_no_dne_expr`: the per-expression induction. Only two cases carry
    real content (`getAttr` on an entity and `.getTag`); every other operator
    follows from the inductive hypotheses.

**Upstream simplification.** The crux, `checked_eval_entity_exists`, needed `checked_eval_entity_reachable` *without* its `entities.contains euid` hypothesis (we must conclude existence, not assume it).  That hypothesis turned out to be vestigial: every `ReachableIn` base case is `in_start` from request-root membership, and only `step` needs `find? = some`, which it constructs locally. So I dropped `contains` from the six `Slice/Reachable/*` helpers and its one consumer in `Slice.lean`. This is a net simplification (the diff in those files is negative) and is why they appear in the PR.

**Addressed in review (Kastner)**: folded the key lemma into `checked_eval_entity_exists`; removed unused `closed_at_level_succ`;  moved `bind_ne_error` to Thm/Data/Control.lean; added the validator-level lift `validate_with_level_no_dne(_wf)`; and added the executable `Entities.closedAtLevel` with the `closedAtLevel_iff` equivalence, over which the top-level theorem is stated.

### DIFF
--- a/cedar-lean/Cedar/Thm/Validation/Levels/NoEntitiesError.lean
+++ b/cedar-lean/Cedar/Thm/Validation/Levels/NoEntitiesError.lean
@@ -17,6 +17,7 @@
 import Cedar.Spec
 import Cedar.Data
 import Cedar.Validation
+import Cedar.Thm.Data.Control
 import Cedar.Thm.Validation.Typechecker
 import Cedar.Thm.Validation.Typechecker.Types
 import Cedar.Thm.Validation.Slice
@@ -64,18 +65,14 @@ lookups (`hasAttr`/`hasTag`/`mem`), so for them the result follows from the
 inductive hypotheses alone.  Hence only `level_based_no_dne_get_attr` and the
 `.getTag` branch of `level_based_no_dne_binary_app` carry real content.
 
-## The crux (`reachable_of_checked_eval_entity`)
+## The crux (`checked_eval_entity_exists`)
 
-The hard part is `reachable_of_checked_eval_entity`: it is
-`checked_eval_entity_reachable` (Slice/Reachable.lean) with the
-`entities.contains euid` *hypothesis removed*.  The existing lemma assumes the
-reached entity exists; we must instead *conclude* existence, so we need
-reachability that does not presuppose it.  Reachability does not actually
-require the endpoint to exist (`ReachableIn.in_start` only needs membership in
-the work set; only intermediate `step`s require `find? = some`), so the
-generalisation should hold, but proving it means dropping the `contains`
-hypothesis from the `Slice/Reachable/*` helper lemmas it depends on.  That is
-the bulk of the work for this issue; the rest of this file is plumbing.
+`checked_eval_entity_reachable` (Slice/Reachable.lean) already places a
+level-checked entity at `ReachableIn … (n + 1)`, and with its `entities.contains`
+hypothesis dropped upstream it does so without presupposing existence.  Composing
+that with `EntitiesClosedAtLevel` gives `checked_eval_entity_exists`, the single
+lemma consumed by the two dereferencing cases (`getAttr` on an entity, `.getTag`).
+Everything else is plumbing.
 -/
 
 namespace Cedar.Thm
@@ -90,19 +87,31 @@ the request roots (`request.sliceEUIDs`) within `level` dereferences is present
 in the store.  This is the closure condition #642 adds.
 
 Phrased relative to the request because that is what `checked_eval_entity_reachable`
-and friends produce (`ReachableIn entities request.sliceEUIDs euid (n + 1)`),
-which makes `reachable_entity_exists` below definitional.
-
-Alternative considered: a request-independent, store-intrinsic closure ("every
-entity in `es` has all of its `sliceEUIDs` references in `es`, up to depth n").
-That is strictly stronger and could be offered as `EntitiesClosed es n` with a
-derivation `EntitiesClosed es n → EntitiesClosedAtLevel es r n` for any request
-whose roots lie in `es`; but the roots (principal/action/resource) need not be
-in the store, so the request-relative form is the one the proof consumes.  Open
-to switching if a store-only statement is preferred on review.
+and friends produce (`ReachableIn entities request.sliceEUIDs euid (n + 1)`).
+`Entities.closedAtLevel` (Spec/Slice.lean) is the executable algorithm that
+decides this predicate, and `closedAtLevel_iff` proves the two equivalent, so the
+guarantees below can be discharged by *running* the check on a concrete store.
 -/
 def EntitiesClosedAtLevel (entities : Entities) (request : Request) (level : Nat) : Prop :=
   ∀ euid, ReachableIn entities request.sliceEUIDs euid level → entities.contains euid
+
+/--
+`Entities.closedAtLevel` (the executable check) agrees with the
+`EntitiesClosedAtLevel` predicate.  `Entities.closedAtLevel` reuses the reachable
+set computed by the slicing algorithm (`Entities.sliceAtLevel.sliceAtLevel`), and
+membership in that set coincides with `ReachableIn` (`reachable_then_mem_slice` /
+`reachable_of_mem_slice`), so the algorithm checks `contains` on exactly the
+entities the predicate quantifies over.
+-/
+theorem closedAtLevel_iff {entities : Entities} {request : Request} {n : Nat} :
+  entities.closedAtLevel request n = true ↔ EntitiesClosedAtLevel entities request n
+:= by
+  simp only [Entities.closedAtLevel, Set.all_eq_true, EntitiesClosedAtLevel]
+  constructor
+  · intro h euid hr
+    exact h euid (reachable_then_mem_slice hr)
+  · intro h euid hmem
+    exact h euid (reachable_of_mem_slice hmem)
 
 /--
 Per-expression inductive hypothesis, parallel to `TypedAtLevelIsSound`: a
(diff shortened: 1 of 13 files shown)


cedar-lean/Cedar/Thm/Validation/Levels/NoEntitiesError.lean
```diff
@@ -0,0 +1,762 @@
...
+-/
+theorem reachable_of_checked_eval_entity {e : Expr} {n nmax : Nat} {c c' : Capabilities} {tx : TypedExpr} {env : TypeEnv} {request : Request} {entities : Entities} {v : Value} {path : List Attr} {euid : EntityUID}
+  (hc : CapabilitiesInvariant c request entities)
+  (hr : InstanceOfWellFormedEnvironment request entities env)
+  (ht : typeOf e c env = .ok (tx, c'))
+  (hl : tx.EntityAccessAtLevel env n nmax path)
+  (he : evaluate e request entities = .ok v)
+  (ha : Value.EuidViaPath v path euid) :
+  ReachableIn entities request.sliceEUIDs euid (n + 1)
+:= checked_eval_entity_reachable hc hr ht hl he ha
```
john-h-kastner-aws: Kind of odd to have the key lemma be an alias for something we already had. In the rest of the proof, it looks like this this lemma is always composed with `reachable_entity_exists`, so IMO the key lemma really ought to be something like

```lean
theorem checked_eval_entity_exists
  (hc : CapabilitiesInvariant c request entities)
  (hr : InstanceOfWellFormedEnvironment request entities env)
  (ht : typeOf e c env = .ok (tx, c'))
  (hl : tx.EntityAccessAtLevel env n nmax path)
  (he : evaluate e request entities = .ok v)
  (ha : Value.EuidViaPath v path euid)
  (hcl : EntitiesClosedAtLevel entities request n)
: entities.contains euid
```

cedar-lean/Cedar/Thm/Validation/Levels/NoEntitiesError.lean
```diff
@@ -0,0 +1,762 @@
...
+    evaluate e request entities ≠ .error .entityDoesNotExist
+
+/-! ## Closure lemmas -/
+
+/--
+Closure is downward closed in the level: fewer hops reach fewer entities
+(`reachable_succ`), so closure at `n + 1` implies closure at `n`.  Needed
+because dereferencing operators decrement the level for their sub-expressions.
+-/
+theorem closed_at_level_succ {entities : Entities} {request : Request} {n : Nat}
```
john-h-kastner-aws: Doesn't look like you actually use this lemma anywhere right now

cedar-lean/Cedar/Thm/Validation/Levels/NoEntitiesError.lean
```diff
@@ -0,0 +1,762 @@
...
+:= checked_eval_entity_reachable hc hr ht hl he ha
+
+/-! ## Generic monadic + per-operation "never DNE" helpers -/
+
+/--
+Monadic bind preserves "not this error": if neither the first computation nor
+the continuation (on success) yields error `e`, the bind does not either.  The
+workhorse for every error-propagation (non-dereferencing) operator case.
+-/
+theorem bind_ne_error {α β ε} {r : Except ε α} {f : α → Except ε β} {e : ε}
```
john-h-kastner-aws: Probably move this to `Thm/Data/Control.lean`

cedar-lean/Cedar/Thm/Validation/Levels/NoEntitiesError.lean
```diff
@@ -0,0 +1,762 @@
...
+* `.binaryApp .getTag`           (`getTag uid tag es` → `Entities.tags`)
+
+Every other operator either short-circuits or uses the infallible `*OrEmpty`
+lookups (`hasAttr`/`hasTag`/`mem`), so for them the result follows from the
+inductive hypotheses alone.  Hence only `level_based_no_dne_get_attr` and the
+`.getTag` branch of `level_based_no_dne_binary_app` carry real content.
+
+## The crux (`reachable_of_checked_eval_entity`)
+
+The hard part is `reachable_of_checked_eval_entity`: it is
```
john-h-kastner-aws: Most of this discussion isn't necessary given that you were able to trivially remove the hypothesis on the existing lemma

cedar-lean/Cedar/Thm/Validation/Levels/NoEntitiesError.lean
```diff
@@ -0,0 +1,762 @@
...
+
+Alternative considered: a request-independent, store-intrinsic closure ("every
+entity in `es` has all of its `sliceEUIDs` references in `es`, up to depth n").
+That is strictly stronger and could be offered as `EntitiesClosed es n` with a
+derivation `EntitiesClosed es n → EntitiesClosedAtLevel es r n` for any request
+whose roots lie in `es`; but the roots (principal/action/resource) need not be
+in the store, so the request-relative form is the one the proof consumes.  Open
+to switching if a store-only statement is preferred on review.
+-/
+def EntitiesClosedAtLevel (entities : Entities) (request : Request) (level : Nat) : Prop :=
```
john-h-kastner-aws, line 95: Slightly larger request here: I'd like to _also_ have a executable definition of the algorithm for checking if a set of entities is closed at a level, then prove it equivalnt to this predicate and state the main theorems in terms of the executable definition. That way cedar users have a concrete algorithm they could implement if they want this additional guarantee

Happy to leave that for a follow up PR for just open another issue for it if you don't want to do it now.

cedar-lean/Cedar/Thm/Validation/Levels/NoEntitiesError.lean
```diff
@@ -0,0 +1,762 @@
...
+
+/-! ## Policy- and validator-level wrappers
+
+Mirror `typecheck_policy_with_level_is_sound` /
+`validate_with_level_is_sound_wf`, threading `EntitiesClosedAtLevel` through to
+the top-level expression theorem.  Stubbed until the expression theorem is
+complete.
+-/
+
+theorem typecheck_policy_with_level_no_dne {p : Policy} {tx : TypedExpr} {n : Nat} {env : TypeEnv} {request : Request} {entities : Entities}
```
john-h-kastner-aws, line 710: The final step for the end-to-end proof would to be lift this to `isAuthorized` which should ofc be trivial since `isAuthorized` itself can't produce an `.entityDoesNotExist` error

E.g., https://github.com/MavenRain/cedar-spec/blob/72c93a782a5a9a6ec226253f952272b61822e245/cedar-lean/Cedar/Thm/Validation.lean#L86

john-h-kastner-aws (commented):
Looks pretty good. The main theorem captures exactly what I had in mind. A few comments and two places I'd like to extend the proof.

MavenRain:
Thanks for the review!  All six points are in the latest push.
  
   1. Key lemma.  `reachable_of_checked_eval_entity` and `reachable_entity_exists`
   are gone, replaced by a single `checked_eval_entity_exists` that composes
   `checked_eval_entity_reachable` with the closure hypothesis and concludes
   entities.contains euid directly;  both dereferencing call sites use it.  One
   correction to the sketch: since c`hecked_eval_entity_reachable` yields
   `ReachableIn … (n + 1)` from `EntityAccessAtLevel … n`, the closure hypothesis
   sits at level n + 1 (`hcl : EntitiesClosedAtLevel entities request (n + 1)`).
   2. `closed_at_level_succ`.  Removed; it was unused.
   3. `bind_ne_error`.  Moved to `Thm/Data/Control.lean`, beside the other `Except`
   bind lemmas.
   4. Crux discussion.  Trimmed, now that dropping the contains hypothesis
   upstream turned out trivial.
   5. Lift to `isAuthorized`.  Added `validate_with_level_no_dne_wf` (Levels.lean)
   and the user-facing `validate_with_level_no_dne` (Validation.lean, next to
   `validate_with_level_is_sound`): for a level-validated policy set against a
   closed store, ∀ p ∈ ps, evaluate p.toExpr request entities ≠ .error
   .entityDoesNotExist.  I stated it at the evaluate boundary rather than on the
   `isAuthorized` Response, because the authorizer surfaces evaluation results
   only through the kind-agnostic `errorPolicies / satisfiedPolicies` (a Set
   PolicyID), so there is no DNE kind to observe on the response itself; these
   are exactly the evaluations `isAuthorized` performs, and it raises no
   `entityDoesNotExist` of its own.  I'd be glad to add an explicit `isAuthorized`-named
   corollary if you would prefer one.
   6. Executable closure check.  Implemented here rather than deferred.
   `Entities.closedAtLevel : Entities → Request → Nat → Bool`(Spec/Slice.lean)
   reuses the slicing algorithm's reachable-entity set and checks each member is
   present.  `closedAtLevel_iff` proves closedAtLevel … = true ↔
   EntitiesClosedAtLevel … (the new converse `reachable_of_mem_slice` is the other
   half of the existing `slice_contains_reachable`), and
   `validate_with_level_no_dne` is stated over the executable check so a user can
   run it.

john-h-kastner-aws (approved):
looks great

john-h-kastner-aws:
>  I'd be glad to add an explicit isAuthorized-named corollary if you would prefer one.

I think it could be a nice addition to both this and the main validation soundness theorem. Something like "forall pid in errorPolicies, `(ps.find pid).toExpr.evaluate` is one of the permissible error kinds". IMO better in a separate PR after merging this one.

victornicolet (approved):

PR #968, opened 2026-06-11
fuzz targets for proto Schema::decode

victornicolet:
Fuzz targets for decoding protobufs of schemas:
- adding roundtrip properties on output of the target testing from bytes,
- new target testing from generated protobuf models.

### DIFF
--- a/cedar-drt/fuzz/src/props.rs
+++ b/cedar-drt/fuzz/src/props.rs
@@ -18,8 +18,10 @@
 
 use crate::roundtrip_entities::pretty_assert_entities_deep_eq;
 use cedar_policy::{
-    Entities, Entity, Expression, Policy, PolicySet, Request, RestrictedExpression, Template,
+    Entities, Entity, Expression, Policy, PolicySet, Request, RestrictedExpression, Schema,
+    Template,
 };
+use cedar_policy_core::validator::ValidatorSchema;
 
 /// An [`Entity`] should roundrtrip through serialization with json and then deserialization.
 /// The [`Entity`] gets converted to a singleton [`Entities`].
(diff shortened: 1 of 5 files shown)


cedar-drt/fuzz/src/props.rs
```diff
@@ -146,3 +148,32 @@ pub fn policyset_to_json_deserializes(original: PolicySet) -> PolicySet {
...
+    let _roundtripped = Schema::from_json_value(json.clone()).unwrap_or_else(|e| {
+        panic!("JSON from Schema failed to re-parse.\nJSON: {json}\nError: {e}")
+    });
+}
+
+/// A [`Schema`] should roundtrip through Cedar schema syntax. Convert to a JSON
+/// schema fragment, print to Cedar text, then parse back as a [`Schema`].
+pub fn schema_to_cedar_roundtrips(schema: &Schema) {
+    let validator_schema: &ValidatorSchema = schema.as_ref();
+    let fragment = validator_schema
```
john-h-kastner-aws: Can this be `validator_schema.to_cedarschema()`

cedar-drt/fuzz/src/props.rs
```diff
@@ -146,3 +148,32 @@ pub fn policyset_to_json_deserializes(original: PolicySet) -> PolicySet {
...
+pub fn schema_to_json_roundtrips(schema: &Schema) {
+    let validator_schema: &ValidatorSchema = schema.as_ref();
+    let fragment = validator_schema
+        .to_json_schema()
+        .expect("ValidatorSchema could not be converted to JSON schema fragment");
+    let json = serde_json::to_value(&fragment)
+        .unwrap_or_else(|e| panic!("Schema fragment could not be serialized to JSON.\nError: {e}"));
+    let _roundtripped = Schema::from_json_value(json.clone()).unwrap_or_else(|e| {
+        panic!("JSON from Schema failed to re-parse.\nJSON: {json}\nError: {e}")
+    });
```
john-h-kastner-aws: Did you try asserting equality here?

victornicolet: I guess I should rename the function, like some other functions here it's not trying to assert equality but just thread the thing through the different conversions.

john-h-kastner-aws: I think I do expect exact equality here, so we could test for it

victornicolet: In the specific case of the schemas that came out of the generators in the fuzz targets we have in this PR? 
You may be right, I'll think about it.

john-h-kastner-aws (approved):

cedar-drt/fuzz/src/props.rs
victornicolet: Adding the equality test discovered this: https://github.com/cedar-policy/cedar/pull/2404 using the bytes target

whonore (approved):

PR #970, opened 2026-06-16
Update model decoding to handle z3 models

john-h-kastner-aws:
Adds changes to model decoding from  https://github.com/cedar-policy/cedar/pull/2408 and https://github.com/cedar-policy/cedar/pull/2368

Current diff tests only check decoding models from cvc5, so these changes shouldn't be necessary to pass nightly tests, but good to make ofc. There are no proofs on this code, so this change is just in the implementation


whonore (approved):

cedar-lean/Cedar/SymCC/Decoder.lean
```diff
@@ -287,6 +299,15 @@ where
...
       | other                                => fail "Option (BitVec 7)" other
+    | [.symbol "_", .symbol bvStr, .numeral w] =>
+      if bvStr.startsWith "bv" then
+        match (bvStr.drop 2).toNat? with
+        | .some val =>
+          if w == 0 then fail "non-zero width" w
+          else if val >= 2^w then fail s!"value fitting in {w} bits" val
+          else Term.bitvec (BitVec.ofNat w val)
+        | .none => fail "numeric bv value" bvStr
+      else fail "indexed bitvec (_ bvN W)" bvStr
```
victornicolet, line 310: iiuc this means we expect that the only indexed terms we might get are bitvectors. This makes sense, but I think worth noting.

victornicolet (approved):

PR #973, opened 2026-06-22
Update TPE DRT for API changes

john-h-kastner-aws:
* Use public API to access true/false/error/residual permits/forbids
* Use `true_permits`/`true_forbids` instead of `satisfied_`
* Replace `residual_policies` call with `policies`

Fixes break caused by change in cedar-policy/cedar#2401. Also depends on cedar-policy/cedar#2418 

unrelated change: made the error case comparison slightly stricter, failing on some ffi error that were previously ignored.

### DIFF
--- a/cedar-lean-ffi/src/lean_ffi/tpe.rs
+++ b/cedar-lean-ffi/src/lean_ffi/tpe.rs
@@ -141,12 +141,12 @@ mod test {
         }
         check_set!(
             lean.satisfied_permits,
-            rust_inner.satisfied_permits(),
+            rust_inner.true_permits(),
             "satisfied_permits"
         );
         check_set!(
             lean.satisfied_forbids,
-            rust_inner.satisfied_forbids(),
+            rust_inner.true_forbids(),
             "satisfied_forbids"
         );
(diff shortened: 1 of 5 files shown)


katherine-hough (approved):

katherine-hough (approved):

cedar-lean-ffi/src/lean_ffi/tpe.rs
```diff
@@ -108,79 +108,79 @@ mod test {
...
-                let pst_expr = pst::Expr::try_from(ast_expr)
+                let id = rp.id().to_string();
+                let pst_policy = rp
+                    .to_pst()
                     .expect("ast->pst conversion should succeed for residuals");
-                (id, pst_expr)
+                let [pst::Clause::When(pst_expr)] = pst_policy.body().clauses().as_slice() else {
+                    panic!("expected exactly one `when` clause in residual")
+                };
+                (id, pst_expr.as_ref().clone())
```
victornicolet, line 183: That looks much nicer than before.

victornicolet (approved):

PR #979, opened 2026-06-27
Formalize and Verify the Decimal parser

cruisesong7:
## Summary
- Formalize the Decimal parser via an attribute grammar: `IsWfStr` defines
  well-formedness (the context-free structure) and `computeValue` defines
  the semantic attribute (the synthesized integer value)
- Prove `parse_toString_roundtrip`: `parse ∘ toString = some`
- Prove `parse_eq_none_iff`: parse fails iff not well-formed or value overflows Int64
- Prove `parse_sound`: if parse succeeds then input is well-formed, value is
  in Int64 range, and the result has exactly that computed value
- Prove `toString_injective`: distinct decimals produce distinct strings
- Extract reusable string utility lemmas to `Cedar.Thm.Data.String`

## Test plan
- [ ] `lake build` passes on `cruise-decimal`

### DIFF
--- a/cedar-lean/Cedar/Thm/Ext/Decimal.lean
+++ b/cedar-lean/Cedar/Thm/Ext/Decimal.lean
@@ -1,0 +1,184 @@
+module
+
+public import Cedar.Thm.Ext.Decimal.Lemmas
+
+import all Cedar.Data.Int64
+import all Cedar.Spec.Ext.Decimal
+import all Cedar.Spec.Ext.Util
+import all Cedar.Thm.Ext.Decimal.Grammar
+import all Cedar.Thm.Ext.Decimal.Lemmas
+
+namespace Cedar.Thm.Decimal
+open Cedar.Spec.Ext
+
+/-- Completeness of `Decimal.parse`: if a string is well-formed and its computed value
+    matches `d.toInt`, then parsing accepts the string as `d`. -/
+public theorem parse_complete (s : String) (d : Decimal)
+    (hwf : IsWfStr s) (hval : computeValue s = some d.toInt) :
+    Decimal.parse s = some d := by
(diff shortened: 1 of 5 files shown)


cedar-lean/Cedar/Thm/Ext/Decimal.lean
```diff
@@ -0,0 +1,466 @@
...
+        rw [String.toList_append]
+        intro h
+        cases List.mem_append.mp h with
+        | inl h => exact absurd (hz '_' h) (by decide)
+        | inr h =>
+          rw [Nat.toString_eq_repr, Nat.toList_repr] at h
+          exact Nat.underscore_not_in_toDigits h
+      simpa [String.contains] using h
+    split
+    · -- "000" ++ toString n, n < 10
```
MavenRain: A `have zeropad_ok (pad) (hp) : . . .` could cut some duplication between the different cases where the primary discriminant is the number of leading zeros.

cedar-lean/Cedar/Thm/Ext/Decimal.lean
```diff
@@ -0,0 +1,466 @@
...
+        rw [List.getLast?_eq_some_getLast hne] at h
+        have hmem := List.getLast_mem hne
+        injection h with h; rw [h] at hmem
+        simp [List.mem_cons] at hmem
+    · split
+      · -- "00" ++ toString n, 10 ≤ n < 100
+        rename_i h2 h1
+        rw [hno_us "00" _ (by simp)]
+        simp [String.toNat?, String.Slice.toNat?]
+        simp[toDigits_foldl_roundtrip _]
```
MavenRain: Nit: missing space between `simp` and `[`

MavenRain (approved):

PR #990, opened 2026-07-22
Prove termination for `SExpr.decodeLit`

john-h-kastner-aws:
This was a `partial def` for no particular reason. Filling in termination was not complicated.

Review with hide whitespace enabled.

### DIFF
--- a/cedar-lean/Cedar/SymCC/Decoder.lean
+++ b/cedar-lean/Cedar/SymCC/Decoder.lean
@@ -212,7 +212,7 @@ instance : Coe α (Result α) where
 def SExpr.fail {α β} [Repr α] (expected : String) (actual : α) : Result β :=
   .error s!"expected {expected}, but got {reprStr actual}"
 
-partial def SExpr.decodeType (types : IdMap TermType) : SExpr → Result TermType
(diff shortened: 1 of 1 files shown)


cedar-lean/Cedar/SymCC/Decoder.lean
```diff
@@ -243,92 +245,105 @@ partial def SExpr.decodeLit (ids : IdMaps) (expectedTy : Option TermType := .non
     match expectedTy with
     | .some (.option ty) => Term.none ty
     | _                  => fail "option type for bare 'none'" expectedTy
-  | .symbol e       => enumOrEmptyRecord e
-  | .sexpr xs       => construct xs
-  | other           => fail "literal expr" other
-where
-  enumOrEmptyRecord (s : String) : Result Term :=
```
john-h-kastner-aws, line 250: I couldn't get the termination proof to go through with this as a `def` inside the mutual, but it's small enough to inline, and then the proof goes through easily.

whonore (approved):

victornicolet (approved):

john-h-kastner-aws:
merging with failing checks. Lean checks look good. I need to make another fix to the package version update PR to fix other failures.

PR #997, opened 2026-07-29
Use schema in TPE to reduce has/hasTag/in/==/is

john-h-kastner-aws:
Extends the TPE model so that `has`, `hasTag`, `in`, `==`, and `is` reduce to literals when type information in the schema (plus error-freedom) is enough to reach a concrete value.

Large diff because `TPE.evaluate` now needs a schema.

Key changes addings this functions with calls at appropriate sites. 

```lean
/-- Does the schema prove that applying this operation to types `ty₁` and `ty₂` must produce `false`? -/
def mustBeFalse (env : TypeEnv) (op₂ : BinaryOp) (ty₁ ty₂ : CedarType) : Bool :=
  match op₂, ty₁, ty₂ with
  | .mem, .entity ety₁, .entity ety₂
  | .mem, .entity ety₁, .set (.entity ety₂) => !env.descendentOf ety₁ ety₂
  | .eq,  .entity ety₁, .entity ety₂        => ety₁ != ety₂
  | .hasTag, .entity ety, .string           => env.ets.tags? ety == some .none
  | _, _, _                                 => false
```

```lean
/-- Does the schema prove that a value of type `ty₁` cannot have attribute `a`? -/
def cannotHaveAttr (env : TypeEnv) (ty₁ : CedarType) (a : Attr) : Bool :=
  match ty₁ with
  | .record rty => (rty.find? a).isNone
  | .entity ety =>
    match env.ets.attrs? ety with
    | .some rty => (rty.find? a).isNone
    | .none     => false
  | _ => false
```

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cedar-lean/Cedar/TPE/Evaluator.lean
+++ b/cedar-lean/Cedar/TPE/Evaluator.lean
@@ -76,17 +76,24 @@ def or : Residual → Residual → CedarType → Residual
     if l.errorFree then true else .or l (.val true rty) ty
   | l, r, ty           => .or l r ty
 
-def apply₁ (req: PartialRequest) (op₁ : UnaryOp) (r : Residual) (ty : CedarType) : Residual :=
+/-  Given a unary operation applied to a non-value residual, reduce it to a value if possible -/
+def tryDecideResidual₁ (op₁ : UnaryOp) (r : Residual) : Option Value :=
+  if r.errorFree then
+    match op₁, r.typeOf with
+    | .is ety₁, .entity ety₂ => .some (ety₁ == ety₂)
+    | _, _ => .none
+  else .none
+
+def apply₁ (op₁ : UnaryOp) (r : Residual) (ty : CedarType) : Residual :=
   match r with
   | .error _ => .error ty
   | _ =>
-    match r.asValue with
-    | .some v => someOrError (Spec.apply₁ op₁ v).toOption ty
-    | .none   =>
-      match op₁, r with
-      | .is ety, .var .resource _ => .val (req.resource.ty == ety) ty
-      | .is ety, .var .principal _ => .val (req.principal.ty == ety) ty
-      | _, _ => .unaryApp op₁ r ty
+  match tryDecideResidual₁ op₁ r with
+  | .some v => .val v ty
+  | .none =>
+  match r.asValue with
+  | .some v => someOrError (Spec.apply₁ op₁ v).toOption ty
+  | .none => .unaryApp op₁ r ty
 
 def inₑ (uid₁ uid₂ : EntityUID) (es : PartialEntities) : Option Bool :=
   if uid₁ = uid₂ then .some true else (es.ancestors uid₁).map (Set.contains · uid₂)
@@ -102,7 +109,24 @@ def getTag (uid : EntityUID) (tag : String) (es : PartialEntities) (ty : CedarTy
   | .some tags => someOrError (tags.find? tag) ty
   | .none => .binaryApp .getTag uid tag ty
 
-def apply₂ (op₂ : BinaryOp) (r₁ r₂ : Residual) (es : PartialEntities) (ty : CedarType) : Residual :=
+/- Given a unary operation applied to a non-value residuals, reduce it to a value if possible -/
+def tryDecideResidual₂ (env : TypeEnv) (op₂ : BinaryOp) (r₁ r₂ : Residual) : Option Value :=
+  if r₁.errorFree && r₂.errorFree then
+    match op₂, r₁.typeOf, r₂.typeOf with
+    | .mem, .entity ety₁, .entity ety₂
+    | .mem, .entity ety₁, .set (.entity ety₂) => if !env.descendentOf ety₁ ety₂ then .some false else none
+    | .eq,  .entity ety₁, .entity ety₂        => if ety₁ != ety₂ then .some false else none
+    | .hasTag, .entity ety, .string           => if env.ets.tags? ety == some .none then .some false else none
+    | _, _, _                                 => .none
+  else .none
+
+def apply₂ (env : TypeEnv) (op₂ : BinaryOp) (r₁ r₂ : Residual) (es : PartialEntities) (ty : CedarType) : Residual :=
+  match r₁, r₂ with
+  | .error _, _ | _, .error _ => .error ty
+  | _, _ =>
+  match tryDecideResidual₂ env op₂ r₁ r₂ with
+  | .some v => .val v ty
+  | .none =>
   match r₁.asValue, r₂.asValue with
   | .some v₁, .some v₂ =>
     match op₂, v₁, v₂ with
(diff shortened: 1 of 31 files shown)


cedar-lean/Cedar/TPE/Evaluator.lean
```diff
@@ -102,7 +102,22 @@ def getTag (uid : EntityUID) (tag : String) (es : PartialEntities) (ty : CedarTy
   | .some tags => someOrError (tags.find? tag) ty
   | .none => .binaryApp .getTag uid tag ty
 
-def apply₂ (op₂ : BinaryOp) (r₁ r₂ : Residual) (es : PartialEntities) (ty : CedarType) : Residual :=
+/-- Does the schema prove that applying this operation to types `ty₁` and `ty₂` must produce `false`? -/
+def mustBeFalse (env : TypeEnv) (op₂ : BinaryOp) (ty₁ ty₂ : CedarType) : Bool :=
+  match op₂, ty₁, ty₂ with
+  | .mem, .entity ety₁, .entity ety₂
```
luxas: minor nit: aligning the order of `evaluator.rs` makes it slightly easier to read.

cedar-lean/Cedar/TPE/Evaluator.lean
```diff
@@ -102,7 +102,22 @@ def getTag (uid : EntityUID) (tag : String) (es : PartialEntities) (ty : CedarTy
...
+  | .mem, .entity ety₁, .set (.entity ety₂) => !env.descendentOf ety₁ ety₂
+  | .eq,  .entity ety₁, .entity ety₂        => ety₁ != ety₂
+  | .hasTag, .entity ety, .string           => env.ets.tags? ety == some .none
```
luxas: this seems ok, but it's a little bit harder to reason about the equivalence between this and `evaluator.rs` when the logic is negated on the rust side. Given that this is the only use case for `may_have_tags` right now, would it be better to just inline this version into the Rust function?

cedar-lean/Cedar/TPE/Evaluator.lean
```diff
@@ -102,7 +102,22 @@ def getTag (uid : EntityUID) (tag : String) (es : PartialEntities) (ty : CedarTy
   | .some tags => someOrError (tags.find? tag) ty
   | .none => .binaryApp .getTag uid tag ty
 
-def apply₂ (op₂ : BinaryOp) (r₁ r₂ : Residual) (es : PartialEntities) (ty : CedarType) : Residual :=
+/-- Does the schema prove that applying this operation to types `ty₁` and `ty₂` must produce `false`? -/
+def mustBeFalse (env : TypeEnv) (op₂ : BinaryOp) (ty₁ ty₂ : CedarType) : Bool :=
```
luxas: nit: maybe specify as `binaryOpMustBeFalse` or `mustBeFalse_2`?

cedar-lean/Cedar/TPE/Evaluator.lean
```diff
@@ -76,16 +76,16 @@ def or : Residual → Residual → CedarType → Residual
...
   match r with
   | .error _ => .error ty
   | _ =>
     match r.asValue with
     | .some v => someOrError (Spec.apply₁ op₁ v).toOption ty
     | .none   =>
-      match op₁, r with
-      | .is ety, .var .resource _ => .val (req.resource.ty == ety) ty
-      | .is ety, .var .principal _ => .val (req.principal.ty == ety) ty
+      match op₁, r.typeOf with
```
luxas: nit: could consider breaking this out both here and in the rust code to `mustBeFalse_1` if we want those invariants to be separately tested/proven/inspected/evolved over time without disturbing the main TPE flow

cedar-lean/Cedar/Thm/TPE/Soundness/Binary.lean
```diff
@@ -45,15 +118,40 @@ theorem partial_evaluate_is_sound_binary_app
 {env : TypeEnv}
 (h₂ : InstanceOfWellFormedEnvironment req es env)
 (h₄ : RequestAndEntitiesRefine req es preq pes)
+(hwt₁ : Residual.WellTyped env x₁)
 (hwt : Residual.WellTyped env x₂)
```
luxas: call this now `hwt_2`?

cedar-lean/Cedar/Thm/TPE/Soundness/Binary.lean
```diff
@@ -34,6 +35,78 @@ open Cedar.Validation
...
+If TPE.mustBeFalse claims an operations must be false, then it is false under concrete
+evaluation for error-free operands.
+-/
+theorem must_be_false_is_false
+{env : TypeEnv}
+{op₂ : BinaryOp}
+{r₁ r₂ : Residual}
+{req : Request}
+{es : Entities}
+(h₂ : InstanceOfWellFormedEnvironment req es env)
```
luxas: nit: any specific reason this is called `h_2` (as opposed to omitting the 2 or it being e.g. 1), is it a convention for `InstanceOfWellFormedEnvironment`?

luxas (approved):
Thanks, LGTM.

cedar-lean/Cedar/Thm/TPE/Soundness/Unary.lean
```diff
@@ -68,16 +92,26 @@ theorem partial_evaluate_is_sound_unary_app
...
+        case isTrue hef =>
+          have hwt' := partial_eval_preserves_well_typed h₂ h_ref hwt
+          have ⟨v, hev, hcmp⟩ := is_evaluates_type_comparison (ety₁ := ety₁) h₂ hwt' htyeq
+            ((Residual.error_free_spec _).mp hef)
+          have hev₁ : Except.toOption (x₁.evaluate req es) = some v := by
+            rw [hᵢ₁, hev]; rfl
+          rw [to_option_some] at hev₁
+          simp [Residual.evaluate, hev₁, hcmp, Except.toOption]
+        case isFalse =>
+          simp [Residual.evaluate]
```
victornicolet: nit: should we have a `simp only`?

cedar-lean/Cedar/Thm/TPE/SchemaInformedReduction.lean
```diff
@@ -0,0 +1,153 @@
...
+This file proves correctness the schema-informed reductions in TPE.
+-/
+namespace Cedar.Thm
+
+open Cedar.Data
+open Cedar.Spec
+open Cedar.Validation
+
+/-- A record value whose type does not declare an attribute `a` does not have `a`. -/
+theorem has_attr_false_of_record_type_absent
```
victornicolet: nit: `of_absent_attr_in_record_type` to make it clear it's the absence of the attribute in the record type? Same for below for entities. It just read a little like the record type is absent now.

victornicolet (approved):

cedar-lean/Cedar/TPE/Evaluator.lean
john-h-kastner-aws: Since this also needs to decide when `is` is `true`, I've aligned naming for the three functions on `tryDecideResidual₁`, `tryDecideResidual₂`, and `tryDecideResidualHas`.

victornicolet (approved):

PR #998, opened 2026-08-04
Apply some minor generator optimizations

john-h-kastner-aws:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cedar-policy-generators/src/abac.rs
+++ b/cedar-policy-generators/src/abac.rs
@@ -30,6 +30,7 @@ use std::cell::RefCell;
 use std::collections::BTreeMap;
 use std::net::{Ipv4Addr, Ipv6Addr};
 use std::ops::{Deref, DerefMut};
+use std::str::FromStr;
 use thiserror::Error;
 
 // Mutate a hypothetically valid string (randomly).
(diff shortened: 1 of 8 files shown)


cedar-policy-generators/src/abac.rs
```diff
@@ -184,30 +185,26 @@ impl ConstantPool {
     }
 
     /// Get an arbitrary string constant from the pool, with maximum size
-    /// indicated by `bound`. If there are no string constants in the pool,
-    /// get a purely arbitrary string constant with that maximum size
+    /// indicated by `bound`. May truncate a longer string in the pool to this
+    /// length.
     pub fn arbitrary_string_constant_bounded(
```
john-h-kastner-aws, line 190: This is a behavior change in addition to optimization. The function now takes an arbitrary string from the constant pool and truncates it the max length. IMO this is good because it lets this get more variety while still drawing from the constant pool

cedar-policy-generators/src/schema.rs
```diff
@@ -62,9 +62,9 @@ pub struct Schema {
     /// list of entity types that occur as a valid principal for at least one
     /// action in the `schema`
     pub principal_types: Vec<ast::EntityType>,
-    /// list of Eids that exist as a non-`None` actions name for an action in
+    /// list of EntityUIDs that exist as a non-`None` actions name for an action in
     /// the schema.
-    pub actions_eids: Vec<ast::Eid>,
+    pub actions_euids: Vec<ast::EntityUID>,
```
john-h-kastner-aws, line 67: We only ever used this for `EntityUID`s, so this change is to construct it as those instead of as `Eid`s

victornicolet (approved):

cedar-policy-generators/src/abac.rs
victornicolet: Isn't also generating fresh string constants from time to time also desirable?

whonore (approved):

PR #1000, opened 2026-08-12
Upgrade Lean to v4.33.0

lianah:
Bump the toolchain from leanprover/lean4:v4.31.0 to v4.33.0 (and doc-gen4/batteries deps to match), then repair the proofs that broke.

*Description of changes:*

Most breakage is due to lean4#13895: backward.isDefEq.respectTransparency.types is now on by default, so metavariable types are compared at implicit rather than default transparency. When simp rewrites a subterm inside a match, a leftover Decidable/ite instance can still reference the un-rewritten form, so the two sides are defeq only at default transparency. This makes simp/grind/ split stop firing ("simp made no progress", "could not split"). The fixes rewrite whole subterms and finish instance-sensitive steps in term mode.

SymCC/Data/LT.lean is the one exception: it keeps the pre-4.33 behavior via a local `set_option backward.isDefEq.respectTransparency.types false` (with a comment), because there the new default causes whnf timeouts instead of a proof break.


john-h-kastner-aws (approved):

victornicolet:
Looks like the FFI build is failing.

cedar-lean/Cedar/Spec/Ext/Datetime.lean
```diff
@@ -101,32 +101,24 @@ def tzOffsetMinsLt60 (str : String) : Bool :=
...
-  Workaround an issue in the datetime library by checking that the year, month,
-  day, hour, minute, and second components of the datetime string are not longer
-  than expected.  https://github.com/leanprover/lean4/issues/7478
+  Check that the timezone offset, when present, is exactly 4 digits: `(+|-)hhmm`.
+  The date is split off on `T` first so the date's `-` separators are not
+  mistaken for the offset sign. A date-only string, or a time without a sign
+  (UTC `Z`), has no offset and passes.
 -/
-def checkComponentLen (str : String) : Bool :=
+def checkOffsetLen (str : String) : Bool :=
```
lianah, line 109: I had to update this to check the offset length to be 4 since the datetime library now allows 1 or 2 digits for each offset component. I removed the general component len check since the issue it references has since been fixed in the Lean datetime library and the unit tests should catch regressions. An alternative would be to keep it to be extra defensive in case the library implementation changes. @john-h-kastner-aws do you have any preference?

john-h-kastner-aws: This looks reasonable. At this point I imagine @cruisesong7 might have the strongest opinions about extension parsing

john-h-kastner-aws:
and the current CI failure is my fault. See https://github.com/cedar-policy/cedar-spec/pull/1002

victornicolet (approved):

john-h-kastner-aws (approved):

PR #1005, opened 2026-08-18
Add Correctness Proofs for Extension Parsers

cruisesong7:
Adds formal proofs for the existing Datetime, Decimal, Duration, and IP address extension parsers, and updates their [documentation](https://cruisesong7.github.io/cedar-spec/). After approval, the docs workflow should deploy a similar page from cedar-spec.

## How to read the proof changes

Each extension proof is divided into three layers:

1. `Cedar/Thm/Ext/<Extension>/Grammar.lean` defines a parser-independent contract: the accepted syntax (`IsWf*` and component structures), its rendering, and its denotation (`computeValue`, `v4Value`, or `v6Value`). These definitions are the specification against which the executable parser is checked.
2. `Cedar/Thm/Ext/<Extension>/Lemmas.lean` connects the concrete parser implementation to that contract. These files follow the parser bottom-up, starting with token and component parsers and ending with whole-parser bridge lemmas.
3. `Cedar/Thm/Ext/<Extension>.lean` contains the short public theorem surface assembled from those bridges. This is the best place to start reviewing.

The main public claims are:

- `parse_sound`: every successful parse has a well-formed witness and the expected value.
- `parse_complete`: every well-formed input whose denotation fits the extension value type parses successfully.
- `parse_eq_none_iff`: characterizes parser rejection.
- `parse_toString_roundtrip` and the corresponding injectivity theorem: serialization recovers the original value. Datetime uses partial `toString?` because not every `Int64` timestamp has a representation in the supported year range.
- `normalize_eq_iff_parse_eq`: two accepted strings normalize equally exactly when they parse to the same value.

For Datetime, it is not necessary to read `Datetime/Lemmas.lean` linearly, as the large middle portions of `Datetime/Lemmas.lean` are supporting proofs for the individual `Std.Time` format parsers. The useful checkpoints are:
- Bridge to `Std.Time`, beginning with `parse_some_decompose`.
- Main theorem: 5-way case split, especially `stdTime_alternation_value`.
- Forward parse-guard lemmas.


cedar-lean/Cedar/Spec/Ext/Datetime.lean
```diff
@@ -100,6 +100,7 @@ def tzOffsetMinsLt60 (str : String) : Bool :=
   | .some minsOffset => minsOffset < 60
   | .none => false
 
+-- `Std.Time` accepts one or two digits per offset field; Cedar requires exactly `hhmm`.
```
victornicolet: nit: leftover comment?

cedar-lean/Cedar/Spec/Ext/IPAddr.lean
```diff
@@ -201,9 +203,21 @@ private def parseNumSegsV6 (str : String) : Option (List (BitVec 16)) :=
...
+      [String.ofList current]
+
+public def splitDoubleColon (str : String) : List String :=
+  splitDoubleColonAux [] str.toList
+
+-- ANCHOR: parseSegsV6
 private def parseSegsV6 (str : String) : Option IPv6Addr := do
   let segs ←
-    match str.splitOn "::" with
+    match splitDoubleColon str with
```
victornicolet: Why was that change necessary? Does it make proving easier, or was there a potential bug?

cedar-lean/Cedar/Spec/Ext/IPAddr.lean
```diff
@@ -272,14 +289,14 @@ public instance IPNet.decLt (d₁ d₂ : IPNet) : Decidable (d₁ < d₂) :=
...
-private def toHex (n : Nat) : String :=
-  let a0 := hexDigitRepr ((n % 0x10000) / 0x1000)
-  let a1 := hexDigitRepr ((n % 0x1000) / 0x100)
-  let a2 := hexDigitRepr ((n % 0x100) / 0x10)
-  let a3 := hexDigitRepr ((n % 0x10) / 0x1)
+public def toHex (n : Nat) : String :=
+  let a0 := String.singleton ((n % 0x10000) / 0x1000).digitChar
+  let a1 := String.singleton ((n % 0x1000) / 0x100).digitChar
+  let a2 := String.singleton ((n % 0x100) / 0x10).digitChar
+  let a3 := String.singleton ((n % 0x10) / 0x1).digitChar
```
victornicolet: I'm generally not expecting changes in the spec, is this change necessary?

victornicolet (commented):
I had a first look at the changes in `Spec`. 
Can you include a guide in your PR explaining what how we should read the changes in the proofs?

cedar-lean/Cedar/Spec/Ext/IPAddr.lean
cruisesong7: This was a proof-engineering change, not a bug fix. I realized that `String.splitOn` has no surrounding theories from [Batteries](https://github.com/leanprover-community/batteries/blob/4cac2177c37f5530c4da76aa8e4307f3fc9e4dcb/Batteries/Data/String/Lemmas.lean#L487-L488) (see the TODO). So to reason it we need to work with [String.splitOnAux](https://leanprover-community.github.io/mathlib4_docs/Init/Data/String/Legacy.html#String.splitOnAux), a raw UTF-8 position-based substring search. Other extension parsers that split strings use single-character separators with `String.splitToList`, whose [String.splitToList_of_valid](https://leanprover-community.github.io/mathlib4_docs/Batteries/Data/String/Lemmas.html#String.splitToList_of_valid) lemma reduces proofs to `List.splitOnP`.

We could of course try to brute-force a theories for splitOnAux if preserving the spec is the top priority.

cedar-lean/Cedar/Spec/Ext/IPAddr.lean
cruisesong7: Oh this one I must admit that it's an overlooked change from Claude. I suspect it's because the proof needed access to the unexposed body of `hexDigitRepr`, so it must had "cheated" and inlined that definition in the specification instead. I have restored the original spec and added `import all Init.Data.Repr` to the theorem module so the proof can unfold it locally. The proof does not really change.

cedar-lean/Cedar/Spec/Ext/IPAddr.lean
cruisesong7: In my newest commit I restore the spec's splitOn implementation, and prove the equivalence between s.splitOn "::" and my previous customized helper, see `splitOn_doubleColon_eq` in `Thm/Ext/IPAddr/Lemmas.Lean`. Turns out that if we focus on a very specific use case of splitOn the proof is still feasible despite some difficulties.

cedar-lean/doc/CedarDoc/IPAddr.lean
```diff
@@ -0,0 +1,282 @@
+import VersoManual
```
victornicolet, line 17: nit: for consistency with the rest of the files, we should have the copyright header (for this file and some others in this PR).

cedar-lean/Cedar/Thm/Ext/Duration/Grammar.lean
```diff
@@ -0,0 +1,170 @@
...
+
+/-- Render an optional duration component as its string representation.
+    `none` produces `""`, `some digits` produces `digits ++ suffix`. -/
+public def durationChunk (digits? : Option String) (suffix : String) : String :=
+  match digits? with
+  | none => ""
+  | some digits => digits ++ suffix
+
+/-- Render a required duration component as `toString n ++ suffix`. -/
+public def durationComponent (n : Nat) (suffix : String) : String :=
```
victornicolet, line 45: This looks like the same function that is in the spec? Also some overlap with durationChunk, maybe to avoid definition duplication they should call each other.

cedar-lean/Cedar/Spec/Ext/Datetime.lean
```diff
@@ -292,4 +298,35 @@ public def Duration.toHours (duration: Duration) : Int64 :=
 public def Duration.toDays (duration: Duration) : Int64 :=
   duration.toHours / 24
 
+/-- Render a duration component as `toString n ++ suffix`. -/
+private def durationComponent (n : Nat) (suffix : String) : String :=
```
victornicolet, line 302: Some overlap with a function in the theorems.

cedar-lean/Cedar/Thm/Ext/Duration.lean
```diff
@@ -0,0 +1,259 @@
...
+properties against `IsWfDuration` and `computeValue`.
+-/
+
+namespace Cedar.Thm.Duration
+open Cedar.Spec.Ext
+open Datetime
+
+/-- Failure characterization for `Duration.parse`: parsing rejects exactly strings that are
+    not well-formed or whose computed value overflows the `Int64` range. -/
+public theorem parse_eq_none_iff (str : String) :
```
victornicolet, line 40: Do you think this is a main theorem, or is it just a lemma of soundness & completeness? It does contain more visible information because it characterizes the fact that computeValue succeeds but not within bounds, but I don't know this is a top-level theorem.

cedar-lean/Cedar/Thm/Ext/Duration.lean
```diff
@@ -0,0 +1,259 @@
...
+    simp [h_add] at h
+    subst h
+    rw [Int64.sub?_add?_inverse dt.val dur.val i h_add]
+    rfl
+
+/-- `parse ∘ toString` roundtrip: parsing the string representation recovers the original. -/
+public theorem parse_toString_roundtrip (d : Duration) :
+    Duration.parse (Duration.toString d) = some d := by
+  let totalMs := d.val.toInt.natAbs
+  let days := totalMs / MILLISECONDS_PER_DAY.toNat
```
victornicolet, line 186: Ideally this is deduplicated with `Duration.toString` in the spec that returns the human-readable canonical form.

victornicolet (commented):
Looks good so far!
I've reviewed Decimal and Duration. A few comments on Duration and duplicated definitions I think.

cedar-lean/Cedar/Thm/Ext/Decimal/Grammar.lean
```diff
@@ -0,0 +1,94 @@
...
+
+/-- Compute the integer value that a decimal string represents, or `none` if its unsigned body
+    does not split into parsable `Natural` and `Fraction` fields. This directly implements the
+    grammar's value function:
+
+      value = sign × (nat(Natural) × 10⁴ + nat(Fraction) × 10^(4 − |Fraction|))
+      where sign = −1 if Sign is '-', else 1
+ -/
+-- ANCHOR: computeValue
+public def computeValue (s : String) : Option Int :=
```
lianah, line 77: `computeValue` is part of the trusted spec for Decimal but contains quite a bit of maybe not obviously correct string mangling. I wonder if we can restructure it in a more declarative form taking advantage of the existential in `IsWfDecimal`:

Define the `decimalValue` separately on each component (`fieldValue` is the same as in `Datetime`):
```
/-- The grammar's value formula, over the three witnessed productions:
      value = sign × (nat(Natural) × 10^4 + nat(Fraction) × 10^(4 − |Fraction|)) -/
public def decimalValue (sign natural fraction : String) : Int :=
  signValue sign *
    (fieldValue natural * Int.pow 10 DECIMAL_DIGITS
     + fieldValue fraction * Int.pow 10 (DECIMAL_DIGITS - fraction.length))
```

And a declarative version:
```
/-- `s` denotes the value `v`. -/
public def IsDecimalValue (s : String) (v : Int) : Prop :=
  ∃ sign natural fraction,
    s = sign ++ natural ++ "." ++ fraction ∧
    IsWfSign sign ∧
    IsNatural natural ∧
    IsWfFrac fraction ∧
    v = decimalValue sign natural fraction
```

Then `IsWfDecimal` can be just:
```
public def IsWfDecimal (s : String) : Prop := ∃ v, IsDecimalValue s v
```
I am not sure if this will make the proofs more challenging, but I think it does make the trusted spec more readable/declarative.

.github/workflows/deploy_docs.yml
```diff
@@ -3,7 +3,7 @@ name: Deploy Lean Documentation to Pages
 on:
   # Runs on pushes targeting the default branch
   push:
-    branches: ["main"]
+    branches: ["main", "cruise-docs"]
```
victornicolet: I don't think this should get merged in main.

.github/workflows/deploy_docs.yml
```diff
@@ -35,21 +35,23 @@ jobs:
...
+        working-directory: ./cedar-lean/doc
+        shell: bash
+        run: source ~/.profile && .lake/build/bin/generate-doc
+      - name: Copy static assets
+        run: cp -r cedar-lean/doc/static cedar-lean/doc/_out/html-multi/
       - name: Upload artifact
         uses: actions/upload-pages-artifact@v5
         with:
-            path: docs/
+            path: cedar-lean/doc/_out/html-multi/
```
victornicolet: What was being deployed here before? I think I'm ok with that change, but someone else should double check.

cedar-lean/Cedar/Spec/Ext/Decimal.lean
```diff
@@ -45,9 +45,10 @@ namespace Decimal
 public def decimal? (i : Int) : Option Decimal :=
   Int64.ofInt? i
 
+-- ANCHOR: parse
 public def parse (str : String) : Option Decimal :=
   match str.splitToList (· = '.') with
-  | ["-", _] => .none -- String.toInt? "-" == some 0
+  | ["-", _] => .none -- guard against bare "-"; redundant on current stdlib (`String.toInt? "-" = none`) but robust to stdlib changes
```
victornicolet: nit: I prefer the old comment style, this is a little too verbose.

cedar-lean/Cedar/Thm/Ext/Duration/Grammar.lean
victornicolet: This is still relevant, we shouldn't duplicate this. It already exists in the spec cedar-lean/Cedar/Spec/Ext/Datetime.lean (which specifies duration and datetime).

cedar-lean/Cedar/Thm/Ext/Duration.lean
```diff
@@ -0,0 +1,208 @@
...
+    parsing accepts the string as `d`. -/
+public theorem parse_complete (str : String) (d : Duration)
+    (hval : IsDurationValue str d.val.toInt) :
+    Duration.parse str = some d := by
+  rw [parse_eq_duration?_of_isDurationValue hval]
+  exact duration?_of_val_toInt d
+
+/-- Soundness of `Duration.parse`: if parsing succeeds, the declarative grammar relation assigns
+    the input exactly the returned duration's value. -/
+public theorem parse_sound (str : String) (d : Duration)
```
victornicolet, line 56: Should this also ensure `IsWfDuration str`? It's probably implied, but I think it's more readable if the theorem states it explicitly?

cedar-lean/Cedar/Thm/Ext/Decimal/Grammar.lean
```diff
@@ -0,0 +1,102 @@
...
+    The existential supplies the split, so no string surgery (`front`/`drop`/`splitToList`) and no
+    `Option` appear here. The relation is single-valued (`isDecimalValue_unique`), and a string is
+    well-formed exactly when it has some value (`isWfDecimal_iff_exists_value`). -/
+-- ANCHOR: IsDecimalValue
+public def IsDecimalValue (s : String) (v : Int) : Prop :=
+  ∃ sign natural fraction,
+    s = sign ++ natural ++ "." ++ fraction ∧
+    IsWfSign sign ∧
+    IsNatural natural ∧
+    IsWfFrac fraction ∧
```
victornicolet: I think by itself this is nice and readable. 

But ideally we wouldn't need to repeat the body of `IsWfDecimal` here. What if we had a shared def relating the components:
```
public def IsDecimalGrammar (s sign natural fraction : String) : Prop :=
  s = sign ++ natural ++ "." ++ fraction ∧
  IsWfSign sign ∧
  IsNatural natural ∧
  IsWfFrac fraction
```
that can be reused in `IsDecimalValue` and `IsWfDecimal`? 
Feel free to argue against it. 
But I think duplicating definition bodies makes it harder to read an identify "what's the single spec I should refer to".

cedar-lean/Cedar/Thm/Ext/Decimal.lean
```diff
@@ -0,0 +1,124 @@
...
+public theorem parse_complete (s : String) (d : Decimal)
+    (hval : IsDecimalValue s d.toInt) : Decimal.parse s = some d := by
+  rw [parse_eq_decimal?_of_isDecimalValue hval]
+  exact Int64.ofInt?_toInt d
+
+/-- Soundness of `Decimal.parse`: if parsing succeeds, then the input is a well-formed rendering
+    whose grammar value is exactly the returned decimal's value. (The value is automatically in
+    `Int64` range, since `d : Decimal = Int64`, so no range conjunct is stated.) -/
+public theorem parse_sound (s : String) (d : Decimal) (h : Decimal.parse s = some d) :
+    IsDecimalValue s d.toInt := by
```
victornicolet, line 52: Similar to my comment in `Duration.lean`, I would have like to see that this implies `s` is well-formed, without having to look at `IsDecimalValue`, which has the body of the well-formedness statement.

cedar-lean/Cedar/Thm/Data/String.lean
```diff
@@ -0,0 +1,344 @@
...
+    used wherever a grammar pins a field's width (the datetime grammar's `YYYY`, `MM`, `SSS`, …). -/
+-- ANCHOR: IsFixedDigits
+public def IsFixedDigits (n : Nat) (s : String) : Prop :=
+  IsDigits s ∧ s.length = n
+-- ANCHOR_END: IsFixedDigits
+
+/-- `Digit{1,n}`: between one and `n` decimal digits. `IsDigits` supplies the lower bound and the
+    length constraint the upper (the decimal grammar's `Fraction ::= Digit{1,4}`). -/
+-- ANCHOR: IsDigitsUpTo
+public def IsDigitsUpTo (n : Nat) (s : String) : Prop :=
```
lianah: Nit: I read this initially as the digits are smaller than n, not the length. Maybe rename `isAtMostNDigits` or similar?

cedar-lean/Cedar/Thm/Data/String.lean
```diff
@@ -0,0 +1,344 @@
...
+module
+
+public import Std.Data.String
+public import Batteries.Data.String
+public import Cedar.Spec.Ext.Util
+import all Cedar.Spec.Ext.Util
+import all Init.Data.String.Search
+import all Init.Data.String.Slice
+
+open Cedar.Spec.Ext
```
lianah, line 26: Should this be in a namespace like `Set`,`Map`& others are?

cedar-lean/Cedar/Thm/Ext/Duration.lean
cruisesong7: I think adding it as a hypothesis would weaken the theorem. Currently it assumes only successful parsing, then from there it derive the str is well formed and denotes the right value.

.github/workflows/deploy_docs.yml
cruisesong7: The old one is doc-gen4 API documentation, I will update the path here to publish both.

cedar-lean/Cedar/Thm/Ext/Duration/Grammar.lean
cruisesong7: Changing the according defs from `private` to `public` in spec to avoid this duplication.

cedar-lean/Cedar/Thm/Ext/Duration.lean
victornicolet: I meant adding it as a conclusion.

cruisesong7: nvm I think I get your point

.github/workflows/deploy_docs.yml
```diff
@@ -61,4 +63,4 @@ jobs:
     steps:
       - name: Deploy to GitHub Pages
         id: deployment
-        uses: actions/deploy-pages@v5
\ No newline at end of file
+        uses: actions/deploy-pages@v5
```
lianah, line 79: Looking at `ci.yaml` it only runs `lake build Cedar SymCC` (and some FFI stuff) so I worry we are not actually building the new proofs in CI. May want to follow the pattern for the other proofs and create a `Cedar/Thm/Ext.lean` that imports the extension parser proofs:
```
public import Cedar.Thm.Ext.Decimal
public import Cedar.Thm.Ext.Duration
public import Cedar.Thm.Ext.Datetime
public import Cedar.Thm.Ext.Datetime.Corollaries
public import Cedar.Thm.Ext.IPAddr
```
and then import that from `Cedar/Thm.lean`

Issue #1007, opened 2026-08-20
SymCC reuses selector names over different data-types


lianah:
The SymCC engine that compiles Cedar policies into SMT generates a data-type for each entity type with a `eid` selector. All entity types share the same selector type which, while supported by both CVC5 and Z3, is inconsistent with the [SMT-LIB standard](https://smt-lib.org/papers/smt-lib-reference-v2.7-r2026-03-27.pdf):

> Similarly, constructors and selectors are function symbols, so none of them can be a previous declared/defined function symbol.

This means that [yaspar](https://github.com/yaspar-org/yaspar) rejects such SMT-LIB files. This does not seem to be an issue with data-types created for record attributes as those are already encoded as `R_i_j` where `i` is the index for the record and `j` the attribute index in the sorted list of attributes. We could take a similar approach for `eid`. 

This issue tracks updating the SMT encoding to be consistent with SMT-LIB, and requires updates to both the Lean and Rust implementations.

PR #1008, opened 2026-08-20
Updated SymCC to emit unique selector names for eid.

lianah:
*Issue #, if available:*

https://github.com/cedar-policy/cedar-spec/issues/1007

*Description of changes:*

Updated SymCC to emit unique selector names for `eid`.


john-h-kastner-aws (approved):
Might be worth adding another fuzz target for this. Since Yaspar only parses, we could generate fully arbitrary (possibly ill-typed) terms and assert that their encoding parses

victornicolet (approved):

PR #1010, opened 2026-08-24
Add residual evaluation TPE differential test target

john-h-kastner-aws:
Test that taking a residual and evaluating it on concrete requests and entities is the same in Rust and Lean. Particularly interesting since the Rust implementation does not define evaluation on a residual and instead translate to an expression before evaluating, creating room for mistakes.

The target uses arbitrary residuals without (correct) type annotations. This is fine because types are irrelevant for this step of partial eval.

This is a large change because it needs to add a protobuf layer for communicating the residual.

### DIFF
--- a/cedar-drt/fuzz/src/tpe.rs
+++ b/cedar-drt/fuzz/src/tpe.rs
@@ -27,7 +27,7 @@ use cedar_policy_core::ast::{self, Value};
 use cedar_policy_core::tpe::residual::Residual;
 use cedar_policy_generators::abac::ABACRequest;
 use cedar_policy_generators::hierarchy::HierarchyGenerator;
-use cedar_policy_generators::schema as generator_schema;
+use cedar_policy_generators::schema;
 use cedar_policy_generators::schema_gen::SchemaGen;
 use libfuzzer_sys::arbitrary::{self, Arbitrary, Unstructured};
 use log::debug;
@@ -185,7 +185,6 @@ impl<'a> Arbitrary<'a> for TpeFuzzTargetInput {
 /// A schema, a hierarchy, and eight (request, residual) pairs.
 #[derive(Debug, Clone)]
 pub struct TpeResidualFuzzTargetInput {
-    pub schema: Schema,
     pub entities: Entities,
     pub residual: Residual,
     pub reqs: [ABACRequest; 8],
@@ -194,9 +193,9 @@ pub struct TpeResidualFuzzTargetInput {
 impl<'a> Arbitrary<'a> for TpeResidualFuzzTargetInput {
     fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
         let settings = abac::FuzzTargetInput::<true>::settings();
-        let gen_schema = generator_schema::Schema::arbitrary(settings.clone(), u)?;
+        let gen_schema = schema::Schema::arbitrary(settings.clone(), u)?;
         let hierarchy = gen_schema.arbitrary_hierarchy(u)?;
-        let requests = [
+        let reqs = [
             gen_schema.arbitrary_request(&hierarchy, u)?,
             gen_schema.arbitrary_request(&hierarchy, u)?,
             gen_schema.arbitrary_request(&hierarchy, u)?,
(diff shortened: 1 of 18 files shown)


cedar-drt/fuzz/src/tpe.rs
```diff
@@ -176,6 +182,99 @@ impl<'a> Arbitrary<'a> for TpeFuzzTargetInput {
     }
 }
 
+/// A schema, a hierarchy, and eight (request, residual) pairs.
+#[derive(Debug, Clone)]
+pub struct TpeResidualFuzzTargetInput {
+    pub schema: Schema,
+    pub entities: Entities,
+    pub residual: Residual,
+    pub reqs: [ABACRequest; 8],
```
whonore, line 191: Is there anything special about 8?

whonore (approved):

cedar-drt/fuzz/src/tpe.rs
```diff
@@ -176,6 +182,99 @@ impl<'a> Arbitrary<'a> for TpeFuzzTargetInput {
...
+}
+
+impl<'a> Arbitrary<'a> for TpeResidualFuzzTargetInput {
+    fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
+        let settings = abac::FuzzTargetInput::<true>::settings();
+        let gen_schema = generator_schema::Schema::arbitrary(settings.clone(), u)?;
+        let hierarchy = gen_schema.arbitrary_hierarchy(u)?;
+        let requests = [
```
victornicolet: Something like:
```
let reqs = (0..8)
            .map(|_| gen_schema.arbitrary_request(&hierarchy, u))
            .collect::<Result<Vec<_>, _>>()?;
```
maybe? 
Also I don't think the `let reqs: [_; 8] = rrequests....` below is doing anything (wrap in ok, collect, unwrap), the `requests` array is already of the right type.

cedar-policy-generators/src/residual.rs
```diff
@@ -0,0 +1,140 @@
...
+        ExprKind::ExtensionFunctionApp { fn_name, args } => ResidualKind::ExtensionFunctionApp {
+            fn_name: fn_name.clone(),
+            args: Arc::new(
+                Arc::unwrap_or_clone(args)
+                    .into_iter()
+                    .map(|a| residual_of_expr(a, u))
+                    .collect::<Result<Vec<_>>>()?,
+            ),
+        },
+        // We can't translate these, but on the principal that it's better not to error when generating
```
victornicolet: nit: principal -> principle

cedar-drt/fuzz/src/tpe.rs
```diff
@@ -176,6 +182,99 @@ impl<'a> Arbitrary<'a> for TpeFuzzTargetInput {
...
+        })
+    }
+
+    fn try_size_hint(
+        depth: usize,
+    ) -> arbitrary::Result<(usize, Option<usize>), arbitrary::MaxRecursionReached> {
+        Ok(arbitrary::size_hint::and_all(&[
+            generator_schema::Schema::arbitrary_size_hint(depth)?,
+            HierarchyGenerator::size_hint(depth),
+            generator_schema::Schema::arbitrary_request_size_hint(depth),
```
victornicolet: There's only one size hint for the 8 requests, maybe we can have 8?

victornicolet (approved):

cedar-drt/fuzz/src/tpe.rs
john-h-kastner-aws: copy-paste from another target

cedar-drt/fuzz/src/tpe.rs
john-h-kastner-aws: Yeah, that's leftover. I was generating a residual per request at first (since that's what you need to do if you're _evaluating_ an expression to get a residual), but now it's just one residual reused.

cedar-policy-generators/src/residual.rs
john-h-kastner-aws: classic

cedar-drt/fuzz/src/tpe.rs
john-h-kastner-aws: but IMO not worth the map/collect. Easier just to have the 8 gen calls to directly build an array

whonore: Too bad [`std::array::try_from_fn`](https://doc.rust-lang.org/stable/std/array/fn.try_from_fn.html) isn't stabilized

victornicolet (approved):

